Issue #300, opened 2014-09-16
FetchAsync in LocalProxy DeadLock


mbernard:
I'm having an issue when I'm using the Async dataPortal methods (FetchAsync(), etc...) at the same time as using a Local Server DataPortal

```
 public async Task Fetch(Type objectType, object criteria, DataPortalContext context, bool isSync)
    {
      if (isSync || Csla.ApplicationContext.LogicalExecutionLocation == ApplicationContext.LogicalExecutionLocations.Server)
      {
        return await _portal.Fetch(objectType, criteria, context, isSync);
      }
      else
      {
        var tcs = new TaskCompletionSource();
        var bw = new Csla.Threading.BackgroundWorker();
        bw.DoWork += (s, o) =>
        {
          o.Result = _portal.Fetch(objectType, criteria, context, isSync).Result;
        };
        bw.RunWorkerCompleted += (s, o) =>
        {
          if (o.Error == null)
            tcs.TrySetResult((DataPortalResult)o.Result);
          else
            tcs.TrySetException(o.Error);
        };
        bw.RunWorkerAsync();
        return await tcs.Task;
      }
    }
```
In my case I'm always falling the in "else" block which seems to dead lock on the ".Result" which for some reason never completes

I just don't get the "else" block and why this method cannot use the code inside the "if" block all the time.

When I change the code to always enter the "if" block no more dead lock and all my async code run just fine.

rockfordlhotka:
@mbernard , there's a compiler warning in the code in WinRT that needs to be addressed so I am reopening this issue. Please take a look at the warning and address it, thank you!

rockfordlhotka:
@mbernard - the warning also exists in the Android project, and I suspect in the iOS project (haven't got to that one yet)

mbernard:
Here a new pull request that fixes the compilation warning
https://github.com/MarimerLLC/csla/pull/317

JasonHardman:
DataPortal fetch no longer executes on a background thread for IOS, Android and .NET.  I have not tried the other DataPortal methods or platforms.

rockfordlhotka:
Reverted the changes for now - obviously need to take another run at this.

mbernard:
@rockfordlhotka I'm investigating the issue. I confirm that it doesn't seem to work properly at the moment

rockfordlhotka:
Thank you @mbernard

mbernard:
[...]

PR #317, opened 2015-01-07
#300 - Fixes a compilation warning

mbernard:
fixes a compilation warning introduced by a fix for issue #300


mbernard:
I'm not able to load the Android and Ios project at the moment, but let me know if there's another compilation error/warning

PR #327, opened 2015-01-26
Closes #306 Use Xamarin DataAnnotations implementation

### LINKED ISSUE #306: Use new Xamarin DataAnnotations implementation
The latest version of Xamarin (at least Xamarin.Forms) appears to include a System.ComponentModel.DataAnnotations implementation. We'll need to change CSLA so it uses this implementation rather than the one built into CSLA.

rockfordlhotka:
Closes #306 Use Xamarin DataAnnotations implementation for iOS and Android

### DIFF
--- a/Source/Csla.Android/Csla.Android.csproj
+++ b/Source/Csla.Android/Csla.Android.csproj
@@ -49,6 +49,7 @@
   <ItemGroup>
     <Reference Include="Mono.Android" />
     <Reference Include="System" />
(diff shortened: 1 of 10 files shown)


weitzhandler:
Any solution for UWP?

rockfordlhotka:
There's a problem with UWP and DataAnnotations?

weitzhandler:
Not directly, but it's impossible to add the `System.ComponentModel.Annotation` package to a portable app targeting iOS, Droid and UWP.

Issue #364, opened 2015-06-02
Analyzer: Capture Result of Save() to Variable


JasonBock:
Most of the time, when Save() is called on a business object, the intent is to do this:

obj = obj.Save();

However, it's easy to forget to do this, and it can lead to weird, unexpected states in the BO.

This analyzer would flag this as a warning, potentially with a code fix to add the assignment.

(Note: this may be a little more difficult if the result is set to a property or field, as that may be a valid intent by the developer, so need to ensure this case isn't flagged as an issue. I personally don't know of any time where I assigned the result of Save() this way, but it's possible.)

JasonBock:
Jonny, I talked to Rocky about this, and we think that maybe it should be considered an error rather than just a warning if the return of `Save()` is ignored. However, we may be missing case(s) where this is valid. Do you know of any?

jonnybee:
@JasonBock, I do not know or remember any case where I would not assign the result of save to a variable. 

Will you cover both Save and SaveAsync methods and also the overload with forceUpdate parameter?

JasonBock:
OK, I'll make this an error :).

Forgot about SaveAsync, I'll add that too.

PR #421, opened 2015-09-09
Addresses #364

JasonBock:


JasonBock:
I'm not entirely convinced the pull request went in as expected. Can you let me know when you've done it so I can take a look at what got in? (Or, more importantly, what may not have?)

Issue #399, opened 2015-07-26
Update NuGet packages for NuGet 3.0


rockfordlhotka:

rockfordlhotka:
http://blogs.msdn.com/b/lucian/archive/2015/07/27/writing-a-nuget-package-for-vs2015-rtm-and-uwp.aspx

dazinator:
It seems the information on that link above was retracted and we are waiting on the NuGet team for this now.

I remember blogging about issues with [NuGet packages and ASP.NET vNext projects](http://darrelltunnell.net/blog/2015/01/18/asp-net-5-vnext-projects-why-your-nuget-package-may-fail-to-install) (this was back in January) - and at that time, powershell scripts within NuGet packages were just plain not supported. I'm very much hoping that has now been rectified!

rockfordlhotka:
Putting this back onto the backlog. This isn't important until either

1. I add full support for dnx as per #337 
1. UWP differentiates from WinRT so there's a real need for a UWP-specific DLL in the core package

rockfordlhotka:
Closing this issue, as it turns out that old-style projects (pre-.NET Core) can and probably should stick with the existing NuGet model, and I already have the UWP package building using the new model, so all is well.

PR #483, opened 2015-11-03
Update to latest NuGet.exe 

rockfordlhotka:


rockfordlhotka:
Should be linked to #399

Issue #518, opened 2015-12-10
EF7 Support


dazinator:
EF 7 RC 1 is released, I can see CSLA has support for EF 4,5 & 6 - is now a good time to consider adding EF7 support?

dazinator:
I have just started work on a PR, will submit shortly.

rockfordlhotka:
I suspect this new project will need to be a net46+dnxcore50 PCL so it supports full .NET and .NET Core.

dazinator:
I'll check if EF7 is usable from a PCL project..

dazinator:
Doesn't look like EF7 is compatible with DNX Core yet. Adding the NuGet package to such a PCL profile yields failure and many messages about: "some packages are not compatible with DnxCore, Version=v5.0."

Also issues like [this](https://github.com/aspnet/EntityFramework/issues/2336) seem to indicate the guidance for new ASP.NET projects that use EF7 is to comment out the DNX Core target from project.json

jonnybee:
My biggest concern with EF7 RC1 is that it is very slow - even slower than EF6 - and has a lot of bugs. Looking forward to RC2 tho...

dazinator:
@rockfordlhotka - I have a PR based on the shared project approach and a .NET 4.6 project. However would you prefer I move this to a PCL library, that just targets .NET 4.6 for the time being, until EF7 later supports DNX Core?

rockfordlhotka:
No, we can make that change if/when it is necessary.

dazinator:
Waiting on resolution of issue #521 - as this might inform the way forward with this a bit more.

rockfordlhotka:
Something is causing this project to not flow through to the nuget output. I suspect the release mode output directory is not correct, so the assemblies don't end up in the same folder structure as all other projects - and so the nuspec doesn't find them to build the nuget package.

PR #519, opened 2015-12-10
#518 - Adding support for EF7

dazinator:
This PR is all about doing the legwork to enable support for EF7, similar to how other versions of EF are supported using a `DbContextManager`

I followed the existing established structure for doing this by looking at how the other EF versions are catered for in the solution. This resulted in:

1. A new Shared Project containing the `DbContextManager`
2. A new .NET 4.6 project
3. A new .NET 4.6 unit tests project (unit test passing)
4. A new Nuspec file.

Please forgive the deletion of the empty unit test in the Csla.Data.EF5.Test project that you will no doubt see below. I should have submitted that seperately as it just falls under the category of general tidying.

# MASSIVE DISCLAIMER

The implementation of `DbContextManager` for EF7 is enough to make the existing unit test pass and that is all. However it is far from perfect. The constructor on `DbContext` is a bit different in EF7 than previous versions and I have currently only added support for the most simplistic scenario - and that's where you don't pass any arguments! If someone wants to chime in here with some ideas on how to further improve the `DbContextManager` once they have taken a look - support specifiying a database name / connection string then happy to discuss further.


Source/Csla.Data.EF7.Shared/DbContextManager.cs
```diff
@@ -0,0 +1,170 @@
...
+    private DbContextManager(string database, string label)
+    {
+      _label = label;                
+      if (string.IsNullOrEmpty(database))
+      {
+        _context = (C) (Activator.CreateInstance(typeof (C)));
+      }
+      else
+      {
+        _context = (C)(Activator.CreateInstance(typeof(C), database));
```
dazinator: I am aware this won't currently work because `DbContext` has no such constructor in EF7 - but left on purpose to raise as a point of concern and potentially get some ideas as to how to overcome in nicest way.

Source/csla.net.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.23107.0
+VisualStudioVersion = 14.0.24720.0
```
dazinator: Looks like i'm ahead! Hopefully this won't cause problems..

Source/Csla.Data.EF7.Shared/DbContextManager.cs
rockfordlhotka: Are there unit tests for EF7 then? If so, how would they pass if an instance of the context can't be created?

rockfordlhotka:
You do need to add the new EF7 NuSpec to the `Build All.ps1` and `Clean All.ps1` scripts so they are included in the NuGet build process.

Source/Csla.Data.EF7.Shared/DbContextManager.cs
dazinator: I specifically meant this call will fail on line 98:

```
(C)(Activator.CreateInstance(typeof(C), database));
```

There is no such constructor overload on `DbContext` that takes a database / connection string name in EF7.

Line 94 above works because it uses the default constructor for `DbContext` - and the unit test I added in this PR exercises this and is passing.

So basically this PR works as long as you don't want to use: `DbContextManager<T>.GetManager("somedbname")

I can add a failing unit test for that if you like..

I have an idea on how to solve this, but it would be a bit of a change to the "usual" API that you'd expect with `DbContextManager`

Basically I'm thinking you'd have to use it more like this (forgive if this doesn't quite compile, written directly in this chat and GitHub sadly doesn't have intellisense..)

```
DbContextManager<MyDbContext>.GetManager( (options) = >  {
 options.UseSqlServer("some connection string here") }
);
```

So my current idea is that DbContextManager would accept a callback, and prior to creating a new instance of DbContext, it would invoke the callback and allow the consumer to set the options.

In the example the `UseSqlServer` method is an extension method that is available to those that use the SqlServer relational provider for EF - but equally people could use whatever EF provider they liked and in which case they would have different extension methods to call (Postgres etc) because thats the way EF7 works :)

I'll submit something on this basis, and add a unit test around it.

dazinator:
Ah right - I was wondering how the build side of things worked. I'll take care of that in next commit.

Source/Csla.Data.EF7.Shared/DbContextManager.cs
rockfordlhotka: I'm a bit concerned with that solution, in that it has the potential for people to end up with mismatched lambda expressions scattered throughout their codebase, with whichever one runs first winning in a (somewhat) nondeterministic manner.

rockfordlhotka: Perhaps the DbContextManager should maintain a static dictionary of options, keyed by database name (or some arbitrary string). On app initialization (or on an app server in the data portal initialization method) people could define the options for their database(s), and then all the code that actually uses the DbContextManager would refer to those pre-defined configurations via that string name?

dazinator: That's true I guess the potential is there for that.

Part of the issue is that the arguments passed to GetManager need to be enough to create a `DbContext` instance, but as an instance is not always created (if it's being reused) then sometimes these arguments will ultimately be redundant (beyond determining the cache key).

The constructor in EF 7 for `DbContext` takes an object which cannot really be used for determining the cache key, and so it is going to be purely redundant if it its passed in to `GetManager` but not used as a `DbContext` is resolved from the cache.

So really there is a kind of "instance management" issue here with the constructor args.

I thought about:

```
DbContextManager<MyDbContext>.GetManager(options);
```

Where `options` is some instance of `DbContextOptions` that the consumer is responsible for creating.

But again - there is potential for the consumer to create lot's of redundant instances of this dotted around their code base, because they create it on demand:

```
var optionsBuilder = new DbContextOptionsBuilder();
var options = optionsBuilder.UseSqlServer("some connection string here") 
DbContextManager<MyDbContext>.GetManager(options);
```

So I kind of ruled that out.

The thing about the lamda approach is that atleast it's only invoked when the constructor arguements are actually required, because it's when a new `DbContext` is being constructed - so it kind of solves the problem of creating redundant instances of arguements that wont be used.

How about this idea then..

We define a `Factory` interface, something like 

```
public interface IDbContextOptionsFactory
{
    DbContextOptions GetOptions();
}

```

The consumer would have to implement that - by creating something like:

```
public class MyDbContextOptionsFactory : IDbContextOptionsFactory
{
    DbContextOptions GetOptions() 
 {
    var optionsBuilder = new DbContextOptionsBuilder();
    var options = optionsBuilder.UseSqlServer("some connection string here") 
    return options;
  }
}
```

Then DbContextManager usage becomes:

 ```
var optionsFactory = new MyDbContextOptionsFactory();
`DbContextManager<MyContext>.GetManager(optionsFactory)` 
```

`DbContextManager` would only invoke the factory when it needs the options because its creating a new `DbContext` instance.

That's still potentially redundant instances of the factory class being created this time though..

Now I am fresh out of ideas.. will sleep on it..

Comment (dazinator) [inline Source/Csla.Data.EF7.Shared/DbContextManager.cs]: > Perhaps the DbContextManager should maintain a static dictionary of options, keyed by database name (or some arbitrary string). On app initialization (or on an app server in the data portal initialization method) people could define the options for their database(s), and then all the code that actually uses the DbContextManager would refer to those pre-defined configurations via that string name?

Sorry - hadn't read this when I wrote the above.

Ok - let me think that through in all it's glory.

Comment (dazinator) [inline Source/Csla.Data.EF7.Shared/DbContextManager.cs]: Have been thinking some more along the lines of what you suggested.
I'm wondering if we should take it further, and use dependency injection - via the new Microsoft.Extensions.DependencyInjection namespace (Previously Microsoft.Framework.DependencyInjection in beta releases)

This would work similar to what you are saying except on startup, you'd register your DbContext's (with their configured options) with an `IServiceCollection` and then `DbContextManager` would resolve the instance using `IServiceProvider`. Think this is going to complicate things somewhat, and perhaps venturing into the new DI land is not something I should attempt into this PR just yet..

```
DbContextManager<C, TOptionsFactory> GetManager(string label)

```

dazinator:
Changes to build all and clean all powershell scripts now made.

Source/Csla.Data.EF7.Shared/DbContextManager.cs
rockfordlhotka: I have avoided taking a dependency on any specific IoC container because if CSLA does that then everyone using CSLA becomes locked into whatever I choose. As a practical matter this probably means that CSLA becomes unusable to anyone _not_ using whatever container I choose.

rockfordlhotka:
Shifting this to the 4.6.400 release, because I want us to have time to consider the whole IoC question and that won't get resolved in the next couple weeks over the holidays.

PR #547, opened 2016-03-08
Adding static property DataPortalServer to DataPortalSelector

BrettJaner:
I've noticed that in my Root objects there's a lot of "On Save" logic in the DoInsertUpdate() that is outside the scope of persisting the object.  Ideally, I'd like this logic out of my business layer and configured via my DI container.  Then I would slip in the logic using an implementation of IDataPortalServer between DatatPortalSelector and SimpeDataPortal.

Any feedback on my problem/pr is appreciated.  Thanks!

### DIFF
--- a/Source/Csla.Shared/Csla.Shared.projitems
+++ b/Source/Csla.Shared/Csla.Shared.projitems
@@ -232,6 +232,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)Server\AuthorizeRequest.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Server\ChildDataPortal.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Server\DataPortal.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Server\DataPortalBroker.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Server\DataPortalContext.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Server\DataPortalException.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Server\DataPortalExceptionHandler.cs" />
--- a/Source/Csla.Shared/Server/DataPortal.cs
+++ b/Source/Csla.Shared/Server/DataPortal.cs
@@ -175,12 +175,12 @@ namespace Csla.Server
 
             break;
           default:
-            portal = new DataPortalSelector();
+            portal = new DataPortalBroker();
             result = await portal.Create(objectType, criteria, context, isSync).ConfigureAwait(false);
             break;
         }
 #else
-        portal = new DataPortalSelector();
+        portal = new DataPortalBroker();
         result = await portal.Create(objectType, criteria, context, isSync).ConfigureAwait(false);
 #endif
         Complete(new InterceptArgs { ObjectType = objectType, Parameter = criteria, Result = result, Operation = DataPortalOperations.Create, IsSync = isSync });
@@ -264,12 +264,12 @@ namespace Csla.Server
             result = await portal.Fetch(objectType, criteria, context, isSync).ConfigureAwait(false);
             break;
           default:
-            portal = new DataPortalSelector();
+            portal = new DataPortalBroker();
             result = await portal.Fetch(objectType, criteria, context, isSync).ConfigureAwait(false);
             break;
         }
 #else
-        portal = new DataPortalSelector();
+        portal = new DataPortalBroker();
         result = await portal.Fetch(objectType, criteria, context, isSync).ConfigureAwait(false);
 #endif
         Complete(new InterceptArgs { ObjectType = objectType, Parameter = criteria, Result = result, Operation = DataPortalOperations.Fetch, IsSync = isSync });
@@ -401,12 +401,12 @@ namespace Csla.Server
             result = await portal.Update(obj, context, isSync).ConfigureAwait(false);
             break;
           default:
-            portal = new DataPortalSelector();
+            portal = new DataPortalBroker();
             result = await portal.Update(obj, context, isSync).ConfigureAwait(false);
             break;
         }
 #else
-        portal = new DataPortalSelector();
+        portal = new DataPortalBroker();
         result = await portal.Update(obj, context, isSync).ConfigureAwait(false);
 #endif
         Complete(new InterceptArgs { ObjectType = objectType, Parameter = obj, Result = result, Operation = operation, IsSync = isSync });
@@ -500,12 +500,12 @@ namespace Csla.Server
             result = await portal.Delete(objectType, criteria, context, isSync).ConfigureAwait(false);
             break;
           default:
-            portal = new DataPortalSelector();
+            portal = new DataPortalBroker();
             result = await portal.Delete(objectType, criteria, context, isSync).ConfigureAwait(false);
             break;
         }
 #else
-        portal = new DataPortalSelector();
+        portal = new DataPortalBroker();
         result = await portal.Delete(objectType, criteria, context, isSync).ConfigureAwait(false);
 #endif
         Complete(new InterceptArgs { ObjectType = objectType, Parameter = criteria, Result = result, Operation = DataPortalOperations.Delete, IsSync = isSync });
--- a/Source/Csla.Shared/Server/DataPortalBroker.cs
+++ b/Source/Csla.Shared/Server/DataPortalBroker.cs
@@ -1,0 +1,112 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="DataPortalBroker.cs" company="Marimer LLC">
+//     Copyright (c) Marimer LLC. All rights reserved.
+//     Website: http://www.lhotka.net/cslanet/
+// </copyright>
+// <summary>Allows interception of DataPortal call</summary>
+//-----------------------------------------------------------------------
+using System;
+using System.Threading.Tasks;
+
+namespace Csla.Server
+{
+  /// <summary>
+  /// Allows the Data Portal call to be intercepted by
+  /// a custom IDataPortalServer implementation.
+  /// </summary>
+  public class DataPortalBroker : IDataPortalServer
+  {
+    /// <summary>
+    /// Gets or sets a reference to a implementation of
+    /// IDataPortalServer to be used.
+    /// </summary>
+    public static IDataPortalServer DataPortalServer { get; set; }
+
+    /// <summary>
+    /// Create a new business object.
+    /// </summary>
+    /// <param name="objectType">Type of business object to create.</param>
+    /// <param name="criteria">Criteria object describing business object.</param>
+    /// <param name="context">
+    /// <see cref="Server.DataPortalContext" /> object passed to the server.
+    /// </param>
+    /// <param name="isSync">True if the client-side proxy should synchronously invoke the server.</param>
+    public Task<DataPortalResult> Create(Type objectType, object criteria, DataPortalContext context, bool isSync)
+    {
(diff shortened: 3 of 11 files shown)


rockfordlhotka:
Can you give an example of this "On Save" logic? I don't understand how the changes to `DataPortalSelector` address that?

I am a little concerned about this change, in that you are introducing a semantic change in the way the data portal works; now it would hold a reference to the underlying data portal provider. For the simple data portal that shouldn't be an issue, but for other server implementations there may be an assumption of a new instance per call - that needs to be confirmed one way or another to avoid issues.

In any case you can see that the CI build failed due to some unit tests failing. It looks like you aren't using the new property you implemented, so nothing ever sets the underlying field.

BrettJaner:
Thanks for the quick response, Rocky!

An example of "On Save" logic would be publishing a message to our message broker notifying that this particular business object has saved.  This call is currently inside our TransactionScope to make sure the broker is notified.  This makes sense for our Production environment, but when I'm developing locally, I don't necessarily care about my integrations and whether or not the message broker is down.  I should be able to remove the call to the broker by just changing the configuration in my DI container.

Here an example of what my custom IDataPortalServer implementation would be:

    public class CustomDataPortalServer : IDataPortalServer
    {
        private readonly OnSaveDispatcher _dispatcher;

        public CustomDataPortalServer(OnSaveDispatcher dispatcher)
        {
            _dispatcher = dispatcher;
        }

        public Task<DataPortalResult> Create(Type objectType, object criteria, DataPortalContext context, bool isSync)
        {
            var dp = new SimpleDataPortal();
            return dp.Create(objectType, criteria, context, isSync);
        }

        public Task<DataPortalResult> Delete(Type objectType, object criteria, DataPortalContext context, bool isSync)
        {
            var dp = new SimpleDataPortal();
            return dp.Delete(objectType, criteria, context, isSync);
        }

        public Task<DataPortalResult> Fetch(Type objectType, object criteria, DataPortalContext context, bool isSync)
        {
            var dp = new SimpleDataPortal();
            return dp.Fetch(objectType, criteria, context, isSync);
        }

        public Task<DataPortalResult> Update(object obj, DataPortalContext context, bool isSync)
        {
            var dp = new SimpleDataPortal();

            Task<DataPortalResult> result = null;

            _dispatcher.Dispatch(obj, () => result = dp.Update(obj, context, isSync));

            if (result == null)
                throw new InvalidOperationException("OnSaveDispatcher.Dispatch never called doDataPortalCall.");

            return result;
        }
    }

    public class OnSaveDispatcher
    {
        public void Dispatch(object obj, Action doDataPortalCall)
        {
            //Dispatch OnSaveHandlers for obj.GetType()
        }
    }

The OnSaveDispatcher dispatchs based on my DI container configuration.

As far as your concern regarding the custom data portal server being an implicit singleton, I can agree with that.  If you think exposing a Factory property instead would make more sense, I'm for it.  It just seemed all the IDataPortalServers were stateless, so I didn't know if the added complexity of added a Factory Property instead of Instance Property was worth it.

I guess a broken build is what I deserve for not running the tests locally.  I apologize and will push a new commit to the PR.

rockfordlhotka:
Have you reviewed the various options that already exist for supporting DI/IoC in the server-side data portal? Specifically we already have hooks where you can initialize your IoC container, and then take over control of the creation of the server-side objecs using that container?

BrettJaner:
I believe you are referring to an implementation of the IDataPortalActivator, and yes I have looked into this hook.  However, I'm against injecting services into my business objects for multiple reasons, but mainly because it is tough to get  it right for serializable objects.  Do I serialize the dependency or not?  If I don't, I need more code to handle injecting the dependency OnDeserialization.  And then, how do I guarantee that an implementation exist on the destination machine?  Someone did a good write up on CSLA and DI a while back, and still didn't get it 100% right.  https://cslanetwithdi.codeplex.com/discussions/572557

rockfordlhotka:
That was @JasonBock - he and I thought this through pretty thoroughly, and I'd really prefer if we follow along that line and enhance it if necessary rather than pursuing a new/duplicate solution to the same problem.

JasonBock:
@BrettJaner This is the approach that I've used:

http://magenic.com/Blog/Post/43/Abstractions-in-CSLA

Now, there are aspects to that approach that I need to update. What you'll see in the article is fairly simplistic, but I don't cover cases where you have to deal with serializable dependencies or if you mark an operation with `[RunLocal]`. I've resolved this since then, and some day I'll create a new article that talks about these upgrades...when I can find the time :). Keep in mind that CSLA is pretty unique in that the objects are mobile, so any dependencies it receives may go along with it if you're not careful. In most scenarios I've been in when I inject dependencies to an object, that object will physically remain in that host and it's never serialized. So DI/IoC in CSLA will naturally be harder.

Generally, the dependencies I need are usually server side (e.g. database connections and service calls). Wiring them up through property dependency and getting rid of them through an implementation of `IDataPortalActivator` is what the article describes (my article, not the one you refer to in the Codeplex post). I also do not use managed backing fields for those properties. Furthermore, I tend to make those fields non-public so they are not visible to clients but can be accessible via Reflection. That, or you can use two "versions" of the BO, one that has all of the server-side dependency properties in place, and the "client" one that doesn't even have those properties defined. You can do that with shared projects and partial classes rather elegantly in 2015. Previous VS versions you'll have to use linked files to achieve the same effect.

In the cases where a dependency resolved from the server needs to come over to the client, just make sure the dependency can be serialized, and then use a managed backing field. Everything should work as expected.

BrettJaner:
Good read @JasonBock .  It's nice to see others struggling with similar problems.  However, of course I do have some concerns :).  Wouldn't it of just been easier to do service location?  I feel going with your implementation, we'd lose some of key benefits of why we should favor DI over Service Location in the first place.  If services fetched via the service locator are never stored as fields on the BO, then you don't even need to worry about serialization.  Also, it is beneficial to be able to .Verfiy() your container's configuration to make sure everything is resolvable.  A business object could have a dependency not registered, and this would not be identified until the business object is used at runtime.  I guess you could write a unit test to cover that, but some of the advanced containers can now detect lifestyle mismatches when verifying (eg. a singleton dependent on a scoped disposable).  For Scope Management, I'd look at bringing this out of the business objects and implementing it in a custom proxy or portal that would then defer to Server.DataPortal.

    public class ScopedDataPortalServer : IDataPortalServer
    {
        private readonly Container _container;

        public CustomDataPortalServer(Container container)
        {
            _container = container;
        }

        public Task<DataPortalResult> Create(Type objectType, object criteria, DataPortalContext context, bool isSync)
        {
            (using scope = _container.BeginLifetimeScope())
            {
                var dp = new Server.DataPortal();
                return dp.Create(objectType, criteria, context, isSync);
            }
        }
    }

So I guess my goal isn't to inject dependencies into my business objects, but it is more to intercept the call between Server.DataPortal and DataPortal_*.  I think interception provides much more flexibility.  It almost seems like CSLA was built with this in mind.  Every DataPortal server side implements IDataPortalServer.  Why not let a developer plug in his own IDataPortalServer implementation where he/she sees fit?  If there wasn't the SetContext logic in Server.DataPortal,  writing my own IWcfPortal would satisfy my needs.  Should we pivot on this PR?  Would a change to Server.DataPortal to make SetContext more "virtual" be a more entertain-able PR?

JasonBock:
I avoid the service locator pattern. I think it ends up adding far more problems than it tries to address. Refer to this article for details on why:

http://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/

Most of the cases of injecting dependencies via property injection happen on the server-side of the object, so I don't really worry too much about serialization. Those cases where I need it, it's pretty easy to do.

I agree about the scope handling in the article. I did not like putting a scope on each object in the graph. That was ugly :). A client of mine has figured out how to keep the scope out of the graph, which has a lot of benefits. They also looked at using a custom `IDataPortalServer` to initialize and dispose the scope for each request, which works. The real problem with this is trying to keep a call context around so you can "reach up" and get the scope. `CallContext` works but I believe that's going away with .NET Core as Remoting doesn't exist there.

But as you said, if you **main** goal is interception, then yes, create a custom `IDataPortalServer` :).

BrettJaner:
Ah yes, another Mark Seemann fan!  Have you read is book, Dependency Injection in .NET?  Chapter 9 is on Inception.  It's the chapter he considers the climax of the book.  If you understand this chapter, you will understand the problem I'm trying to solve.  Interception allows me to handle all my DI concerns server side without dirtying up my business objects with dependencies.

SergeyBarskiy:
For what its worth, I am going to write this exact code that Brett has above in ScopedDataPortal for the same reasons he has this month in my multi-million lines of code written in 3.8 code base.  I feel that in web apps his approach is a very good clean one, giving us flexibility to consume services as needed when needed as well as test those services in complete isolation and exclude them from normal testing.  My 2 cents.

JasonBock:
Yes, I've read Seeman's book. I'm still curious how interception helps when a business object needs to talk to a REST service or database connection. How will it get those dependencies?

BrettJaner:
I've accepted the database connection as a non-injected dependency.  I write my business objects as cookie cutter as possible.  I even have SQL stored procedures that create the C# file.  And after that they don't change or if they do the decision isn't taken very lightly.  Based on this approach I'm not worried about testing the business objects themselves.  In my code base today, the only business object that have dependencies outside of a DB connection is the Root.  This is because work needs to be done pre- and post- save.  These dependencies could be removed from the business object with proper inception.  If you'd like a code example I should be able to spin something up quickly.

BrettJaner:
Here's a sample project of this feature in action

https://github.com/BrettJaner/Csla.ScratchPad

The reference graph is as follows Csla.ScratchPad.Console -> Csla.ScratchPad.ServiceLayer -> Csla.ScratchPad.BusinessLayer.  What's great about this approach is that I can deploy my BusinessLayer without the ServiceLayer if I see fit.  Eg. Csla.ScratchPad.NewConsole -> Csla.ScratchPad.BusinessLayer.

BrettJaner:
It's been awhile here.  Are we putting this pull request to bed?

rockfordlhotka:
I've been preoccupied with many other things, but I'm not convinced we should close this PR.

I _am_ concerned about providing two solutions to the same problem. At the same time, this change does open the data portal in a powerful way and I'm somewhat curious to see where people might take it :smiling_imp: 

What it is missing is a formal tracking issue that explains the change. If you can create that issue I'll link it to this PR and accept the change.

BrettJaner:
Added issue #564.  Let me buff up the pull request with some tests.

BrettJaner:
After careful consideration, I've amended the pull request a bit, hopefully for the better.  I don't want to make any changes to DataPortalSelector.  There's no way for a custom implementation of IDataPortalServer to delegate to the FactoryDataPortal because DataPortalContext.FactoryInfo's setter is internal.  I believe a custom implementation of IDataPortalServer should just delegate to DataPortalSelector instead.  This would require us to take the logic out of there.  So I created a new class, DataPortalBroker.  I'm not married to this name and am open to suggestions, but I didn't want to use the word "Interceptor" because we already have separate functionality using that term.  Every class that was calling into DataPortalSelector is now calling into DataPortalBroker.  The DataPortalBroker houses the static property for setting a custom implementation of IDataPortalServer.

rockfordlhotka:
So the idea is that a developer would set that static value in the data portal activator or something?

BrettJaner:
Ya that's the idea.  I could add implementation to check the config as well if we'd like.

rockfordlhotka:
Let's stick with the current code, and if there's demand to add the config setting we can easily do that later.

Are you ready for me to accept this PR?

BrettJaner:
Yes I'm ready

PR #548, opened 2016-03-11
Addressingissue131

JasonBock:
This doesn't fix all analyzer issues, but it does address one with interfaces defining DP operations


rockfordlhotka:
Closes #131

Issue #576, opened 2016-05-21
Update Resources pt and pt-BR


tfreitasleal:
Add missing Resources pt and pt-BR.
Single cosmetic fix on language neutral resources.

tfreitasleal:
Ready to commit.

PR #577, opened 2016-05-21
Update Resources pt and pt-BR #576

tfreitasleal:


tfreitasleal:
Update Resources pt and pt-BR #576

Issue #457, opened 2015-10-08
Analyzer: Find Usages of Non-Serializable Types


JasonBock:
There are two cases where a dev needs to use serializable types in CSLA to prevent errors from occurring:

* The argument types In a DP operation (e.g. `DataPortal_Fetch()`)
* The type supplied for the generic to `PropertyInfo<T>`

In either case, an error will be created to inform the developer if a non-serialized type in used in either case.

JasonBock:
I better be careful here :)

`PropertyInfo<T>` can take interface types. I do this all the time for dependency injection purposes. So I can't look at the type and assume that if it's not serializable that's good enough. The `T` from `PropertyInfo` isn't constrained at all, so....the rule would have to be: if it's not an interface, then check to see if the type is serializable or not.

@rockfordlhotka, what do you think?

JasonBock:
Actually, I'm not sure I can do this at all.

If a type to a DP operation isn't serializable, but a subclass is passed in, that....will probably work (haven't tried it). I think the only thing I can do that's totally safe in both scenarios is to look to see if it's sealed and not serializable. Anything else, and I really don't know if it's going to fail or not.

rockfordlhotka:
Technically yes, I think a subclass can be `Serializable` even if the parent type isn't. Obviously in that case the subclass takes responsibility for serializing any state higher in the inheritance hierarchy, so this can't be a very common thing to do - that'd be a lot of work and it would be fragile!

But you are right, I suppose creating an analyzer rule around this would be pretty risky given that somebody might go through that work.

_Still_, really, is someone going to create/use a non-serializable type and then _hope_ that they only provide serializable subclass instances for the value? Perhaps a property of type `object` would be the common scenario? Otherwise this seems like a really poor thing to do in the first place doesn't it?

JasonBock:
I would agree that this is probably a really rare case. But you know that means it's going to break someone if I make it an error :)

So I think the algorithm is (in both cases): If the type is not an interface, and it's not serializable, mark it with a warning.

I hesitate to make it an error because of that %0.00001 edge case, but I can do that if you want me to @rockfordlhotka :)

rockfordlhotka:
A warning sounds right - I don't want the fallout from that edge case either :wink:

GillesBer:
My 2 cents on this. 
[...]

PR #581, opened 2016-05-26
#457 Adding new analyzer

JasonBock:
#457


rockfordlhotka:
This PR needs to be attached to one or more issues, otherwise it looks good to go from my perspective.

:accept:

PR #606, opened 2016-06-30
Hide BusinessBase.BypassPropertyChecks from the debugger

klk103:
This fixes an issue I've been having with VS 2015, which now shows nonpublic properties at the top of the list when examining an object. VS will immediately get this property's value when you inspect a business object, but never disposes it, causing the object's reference count to permanently remain above zero and preventing SetProperty from running rules.


rockfordlhotka:
:accept: Closes #607

PR #642, opened 2016-09-16
Exposebypasspropertychecks

bradtwurst:
We have the need to interrogate the current status of
'_bypassPropertyChecks' in classes that inherit from business base.

Attempting to interrogate the status using 'BypassPropertyChecks'
has the side effect of entering the bypass mode


rockfordlhotka:
Is this not already exposed via an interface?

bradtwurst:
I do not see anything that exposes the _bypassPropertyChecks value.  

There is a BypassPropertyChecks property, but it returns the BypassPropertyChecksObject.GetManager(this), which, upon construction, sets the _businessObject._bypassPropertyChecks value to true.

We are attempting to use the current value of _bypassPropertyChecks, but I don't see a way to observe it with affecting the value.

Please correct me if I am misreading the situation.

rockfordlhotka:
Created #646 to track this issue.

Source/Csla.Shared/Core/BusinessBase.cs
```diff
@@ -3598,6 +3598,17 @@ protected override void OnSetChildren(Csla.Serialization.Mobile.SerializationInf
     [NotUndoable]
     private bool _bypassPropertyChecks = false;
 
+    /// <summary>
+    /// Is the business object currently bypassing property checks?
```
rockfordlhotka: Should start with the phrase "Gets a value"

Source/Csla.Shared/Core/BusinessBase.cs
```diff
@@ -3598,6 +3598,17 @@ protected override void OnSetChildren(Csla.Serialization.Mobile.SerializationInf
...
+    /// </summary>
+    protected internal bool IsBypassingPropertyChecks
+    {
+      get
+      {
+        return _bypassPropertyChecks;
```
rockfordlhotka: The get line returns a simple value - just put the whole thing on one line:

```
get { return _bypassPropertyChecks; }
```

Source/Csla.Shared/Core/BusinessBase.cs
bradtwurst: updated

Source/Csla.Shared/Core/BusinessBase.cs
```diff
@@ -3599,15 +3599,8 @@ protected override void OnSetChildren(Csla.Serialization.Mobile.SerializationInf
     private bool _bypassPropertyChecks = false;
 
     /// <summary>
-    /// Is the business object currently bypassing property checks?
     /// </summary>
```
rockfordlhotka: Did we totally lose the comment?

bradtwurst: Not in the code that I intended to commit.

However, the commit did not record the change properly.

An additional commit has been pushed for this line

Issue #523, opened 2015-12-12
Throw Exception For LocalProxy With Sync/Asycn Scenario


JasonBock:
See this conversation for details: https://github.com/MarimerLLC/cslaforum/issues/112

Specifically:

I suppose the 'server-side' data portal could see that
1.The client made a sync call
2.`ExecutionLocation`  is on the client
3.The concrete method about to be invoked is `async Task`  

Then it could throw a `NotSupportedException`

PR #643, opened 2016-09-20
Throwing exception when sync dataportal tries to call async method on client

BrettJaner:
Addresses issue #523.  I've tested it against @keithdv's use cases from https://github.com/MarimerLLC/cslaforum/issues/112 and all seems well.  I'm looking to get some feedback before I spend more time polishing.  Thanks!

### DIFF
--- a/Source/Csla.Shared/Reflection/MethodCaller.cs
+++ b/Source/Csla.Shared/Reflection/MethodCaller.cs
@@ -1081,47 +1081,25 @@ namespace Csla.Reflection
     }
 
     /// <summary>
-    /// Throws an exception if a synchronous data portal call is trying to invoke an asynchronous method on the client.
+    /// Returns true if the method provided is an async method returning a Task object.
     /// </summary>
-    /// <param name="isSync">True if the client-side proxy should synchronously invoke the server.</param>
     /// <param name="obj">Object containing method.</param>
-    /// <param name="methodName">Name of the method.</param>
-    /// <returns></returns>
-    public static void ThrowIfAsyncMethodOnSyncClient(bool isSync, object obj, string methodName)
-    {
-      if (isSync
-        && ApplicationContext.ExecutionLocation != ApplicationContext.ExecutionLocations.Server
-        && MethodCaller.IsAsyncMethod(obj, methodName))
-      {
-        throw new NotSupportedException(string.Format("Unable to synchronously call {0} which is an asynchronous method.", methodName));
-      }
-    }
-
-    /// <summary>
-    /// Throws an exception if a synchronous data portal call is trying to invoke an asynchronous method on the client.
-    /// </summary>
-    /// <param name="isSync">True if the client-side proxy should synchronously invoke the server.</param>
-    /// <param name="obj">Object containing method.</param>
-    /// <param name="methodName">Name of the method.</param>
-    /// <param name="parameters">
-    /// Parameters to pass to method.
+    /// <param name="method">Name of the method.</param>
     /// </param>
-    public static void ThrowIfAsyncMethodOnSyncClient(bool isSync, object obj, string methodName, params object[] parameters)
-    {
-      if (isSync
-        && ApplicationContext.ExecutionLocation != ApplicationContext.ExecutionLocations.Server
-        && MethodCaller.IsAsyncMethod(obj, methodName, parameters))
-      {
-        throw new NotSupportedException(string.Format("Unable to synchronously call {0} which is an asynchronous method.", methodName));
-      }
-    }
-
-    private static bool IsAsyncMethod(object obj, string method)
+    public static bool IsAsyncMethod(object obj, string method)
     {
       return IsAsyncMethod(obj, method, false, null);
     }
 
-    private static bool IsAsyncMethod(object obj, string method, params object[] parameters)
+    /// <summary>
+    /// Returns true if the method provided is an async method returning a Task object.
+    /// </summary>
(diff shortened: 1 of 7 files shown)


BrettJaner:
I agree Rocky.  I was looking to use the logic from the private method, MethodCaller.GetCachedMethod(), which lets me know if the server-side data portal method can be invoked asynchronously.  Would it be okay to leave the MethodCaller.IsAsyncMethod() method in MethodCaller, and just move MethodCaller.ThrowIfAsyncMethodOnSyncClient() method to (blank)??

BrettJaner:
Just talking that out quick, gave me another idea.  Why throw the exception at all?  Just invoke the server-side Data Portal synchronously.  I think that should be okay?

rockfordlhotka:
There's at least one scenario where calling the server-side dataportal sync will lock up the app though (I think in WPF, calling `FetchAsync` on the client, which calls a sync 'server-side' method using a local data portal).

BrettJaner:
I was unable to repro the FetchAsync -> void DataPortal_Fetch deadlock.  However, if a developer "awaits" anywhere inside a Fetch ->Task DataPortal_Fetch method call, the deadlock will happen anyway.  So back to throwing the exception.  I'm not sure we can do this either.  Currently, the way the pull request is implemented it is going to be a breaking change.  Today, if you synchronously call the data portal and the data portal method is async, the deadlock only happens if something it "awaited".

    public static Root Fetch()
    {
        return DataPortal.Fetch<Root>();
    }

    //Deadlocks today, exception will be thrown
    private async Task DataPortal_Fetch()
    {
        await Task.Delay(500);

        Id = 1;
    }

    //Works today, exception will be thrown
    private Task DataPortal_Fetch()
    {
        Id = 1;
    }

BrettJaner:
Don't want to leave this stagnant too long or I'll forget what I was doing :).

Rocky- is this a breaking change we are willing to accept?  I can't think of another way to implement this functionality.

rockfordlhotka:
The current implementation in the PR throws an exception right? So it is only a "breaking change" in that someone now gets an exception rather than ending up in a deadlock?

BrettJaner:
Depending how the DataPortal_* method is implemented.  If it returns a Task we assume it will do async work, but only if there is an "await" inside will it actually do async work.  No await, no deadlock regardless of whether it returns a task or not.  See my above code snippet.

Source/Csla.Shared/Reflection/MethodCaller.cs
```diff
@@ -1080,6 +1080,71 @@ public static object CallMethod(object obj, System.Reflection.MethodInfo info)
...
+    /// <param name="parameters">
+    /// Parameters to pass to method.
+    /// </param>
+    public static void ThrowIfAsyncMethodOnSyncClient(bool isSync, object obj, string methodName, params object[] parameters)
+    {
+      if (isSync
+        && ApplicationContext.ExecutionLocation != ApplicationContext.ExecutionLocations.Server
+        && MethodCaller.IsAsyncMethod(obj, methodName, parameters))
+      {
+        throw new NotSupportedException(string.Format("Unable to synchronously call {0} which is an asynchronous method.", methodName));
```
rockfordlhotka: This needs to be in the resx file in `Csla.Shared.Resources\Resources.resx` so it can be localized.

Source/Csla.Shared/Reflection/MethodCaller.cs
```diff
@@ -1080,6 +1080,71 @@ public static object CallMethod(object obj, System.Reflection.MethodInfo info)
...
+    /// <summary>
+    /// Throws an exception if a synchronous data portal call is trying to invoke an asynchronous method on the client.
+    /// </summary>
+    /// <param name="isSync">True if the client-side proxy should synchronously invoke the server.</param>
+    /// <param name="obj">Object containing method.</param>
+    /// <param name="methodName">Name of the method.</param>
+    /// <param name="parameters">
+    /// Parameters to pass to method.
+    /// </param>
+    public static void ThrowIfAsyncMethodOnSyncClient(bool isSync, object obj, string methodName, params object[] parameters)
```
rockfordlhotka: This method should probably be part of the data portal, not part of `Csla.Reflection`.

Source/Csla.Shared/Reflection/MethodCaller.cs
```diff
@@ -1080,6 +1080,71 @@ public static object CallMethod(object obj, System.Reflection.MethodInfo info)
...
     }
 
+    /// <summary>
+    /// Throws an exception if a synchronous data portal call is trying to invoke an asynchronous method on the client.
+    /// </summary>
+    /// <param name="isSync">True if the client-side proxy should synchronously invoke the server.</param>
+    /// <param name="obj">Object containing method.</param>
+    /// <param name="methodName">Name of the method.</param>
+    /// <returns></returns>
+    public static void ThrowIfAsyncMethodOnSyncClient(bool isSync, object obj, string methodName)
```
rockfordlhotka: This method should probably be part of the data portal, not part of `Csla.Reflection`.

BrettJaner:
I've moved the ThrowIfAsyncMethodOnSyncClient method to the Utilities static class since it is referenced in the SimpleDataPortal and FactoryDataPortal.  Let me know if this is not the right spot.

I'm assuming we are okay with this change breaking some apps that have implemented the async DataPortal "incorrectly"?  If we'd like to continue that discuss, let me know.

PR #672, opened 2016-12-13
AllowSerializableAttributeOnEnums

ajj7060:
Allows SerializableAttribute to be applied to enums, to reduce friction porting to PCL/Xamarin.


ajj7060:
See https://github.com/MarimerLLC/cslaforum/issues/284

Issue #692, opened 2017-02-18
System.InvalidCastException thrown on nullable enum with non null value


tfreitasleal:
Property definition

```
public static readonly PropertyInfo<byte?> RoleProperty = RegisterProperty<byte?>(p => p.Role, "Role");

public Roles? Role
{
    get { return GetPropertyConvert<byte?, Roles?>(RoleProperty); }
    set { SetPropertyConvert<byte?, Roles?>(RoleProperty, value); }
}
```
When RoleProperty is null, get execution throws no exception.
When RoleProperty is NOT null, get execution throws InvalidCastException.

PR #694, opened 2017-02-19
Fix PropertyConvert of nullable type to nullable enum property. #692

tfreitasleal:


tfreitasleal:
Closing as a ")" is missing.

Issue #472, opened 2015-10-22
Provide More Information in UndoException


JasonBock:
Today at a client they showed me an issue with their code that was generating this exception, and the only piece of information we got was:

```
Message=Edit level mismatch in CopyState
```

We had no idea which object was the problem. It would be nice if this exception could contain (either in the `Message` or custom properties):
- The name of the type of the object that caused this issue
- A `ToString()` dump of that object (and maybe this could be optimized out as a configuration switch)
- The type name of the parent if it isn't null (with a `ToString()` dump similar to the previous bullet point)
- The `EditLevel` value CSLA was expecting, and what it was on the object that caused the issue.

tfreitasleal:
hear! hear!

I know this exception quite well. The [Caliburn.Micro port for Windows Forms and Visual WebGUI](http://mvvmfx.codeplex.com/) includes a CSLA .NET adapter (ViewModelBase, etc). This adapter doesn't use `System.Windows.Forms.BindingSource` but binds directly visual elements to business objects and collections. So the well known Windows Forms rules of binding/unbinding to a `BindingSource` don't apply. I had to figure out how to do it and received dozens of `Edit level mismatch` exceptions in the process.

I agree that is important to include
* The name of the type of the object that caused this issue

but in the light of my past experience, I think it's also important to include some information on the mismatched edit levels, something like the object Edit level and the parent Edit level and some hint on why there is a mismatch.

I remember quite well having to set breakpoints in CSLA source code in order to understand what was going wrong. I know that if one follows the rules, one doesn't need this kind of information. But sometimes it's not possible to follow the rules and when this happens, you need all the information you can get.

rob-v:
PR #718

PR #718, opened 2017-04-04
Provide More Information in UndoException #472

rob-v:
Added TypeName, ParentTypeName, EditLevel, ExpectedEditLevel


rockfordlhotka:
@rob-v I need to get a [contributor agreement](https://github.com/MarimerLLC/csla/blob/master/CONTRIBUTING.md) from you.

It isn't clear to me why the build is failing, but it is consistent - I tried re-running it with the same result. Clearly something in the PR is causing the failure.

rob-v:
Ok, I'll send you the agreement. 

Resolved. I found out the error didn't show when rebuilding csla.net, but occured when rebuilding csla.test due to conditional compilation.

rockfordlhotka:
That's why the build server is so important 😃

PR #740, opened 2017-07-05
Fix #739

### LINKED ISSUE #739: Changing WcfProxy.DefaultBinding or its properties has no effect
I'm working on a PowerShell script that uses some of our existing business objects, and connects to a dataportal on our application server the same way that our other applications do, instead of using a local in-process portal.

I can't use the standard app.config approach (because PowerShell), so I thought that configuring the dataportal in the script would look like this (we use transport security instead of message):

```
[Csla.ApplicationContext]::DataPortalProxy = [Csla.DataPortalClient.WcfProxy].AssemblyQualifiedName
[Csla.ApplicationContext]::DataPortalUrlString = "https://portal.internal/Production.svc"
[Csla.DataPortalClient.WcfProxy]::DefaultBinding.Security.Mode = [System.ServiceModel.SecurityMode]::Transport
```

But apparently the Binding property that's loaded from DefaultBinding isn't actually used when creating a channel, so the security mode isn't applied to the binding that's ultimately used.

klk103:
Additionally, default for WcfProxy.DefaultBinding is changed from a BasicHttpBinding to a WSHttpBinding to match Chapter 4.1 of the book ("Setting the CslaDataPortalUrl Value")


klk103:
Ugh, let me redo this.

Issue #760, opened 2017-08-14
Add support for ASP.NET Core 2.0


rockfordlhotka:

rockfordlhotka:
https://blogs.msdn.microsoft.com/webdev/2017/08/14/announcing-asp-net-core-2-0/

rockfordlhotka:
Still need to update the NuGet package for ASP.NET Core.

Issue #688, opened 2017-01-24
ASP.NET Core and ApplicationContext.User authentication


dazinator:
First let me describe the setup!

I have an asp.net core project, with an MVC controller method - the method isn't important aside from the fact it uses authentication via an [Authorize] attribute:

```csharp
        [Authorize]
        [HttpPost]
        public async Task<IActionResult> Foo([FromBody]SomeCriteria crit)
        {
          // 
        }

```

Cookie authentication is pretty common, and in asp.net core it looks like this in Startup.cs

```csharp

            app.UseCookieAuthentication(new CookieAuthenticationOptions() { Events = new CustomCookieAuthenticationEvents() });

```

Notice I have set a custom events class - this is an optional thing you can implement if you want to intercept the standard cookie authentication lifetime events.

Here is mine:

```
 public class CustomCookieAuthenticationEvents : CookieAuthenticationEvents
    {

        private readonly Task _completedTask;

        public CustomCookieAuthenticationEvents()
        {
            _completedTask = Task.FromResult(true);

            this.OnSignedIn = LoginCslaUser;
            this.OnSigningOut = LogoutCslaUser;
            this.OnValidatePrincipal = Validate;
          
        }

        public Task LoginCslaUser(CookieSignedInContext context)
        {
            // sets the user for CSLA purposes.
            Csla.ApplicationContext.User = context.Principal;
            return _completedTask;
        }


        public Task LogoutCslaUser(CookieSigningOutContext context)
        {
            Csla.ApplicationContext.User = new UnauthenticatedPrincipal();
            return _completedTask;
        }

        public Task Validate(CookieValidatePrincipalContext context)
        {
            Csla.ApplicationContext.User = context.Principal;           
            return _completedTask;
        }
    }


```

Pretty sure I am handling the Validate() method incorrectly at the present, however that shouldn't matter for the purposes of this issue.

Now to the crux.. 

When the controller method is called, the [Authorise] attribute does it's job, and cookie authentication runs. The CustomCookieAuthenticationEvents class is invoked and sets ApplicationContext.User to a claims principle.

However as soon as the mvc controller method is actually entered, ApplicationContext.User returns a GenericPrinciple again.

I am guessing this is because CSLA is trying to detect HttpContext.Current which may not be working for asp.net core applications which rely on `IHttpContextAccessor` to get hold of HttpContext instance. I have not looked at the CSLA code, but I assume CSLA is then treating the application like a desktop application, and storing the principle on the current thread.

[...]

PR #769, opened 2017-09-08
ApplicationContext support for ASP.NET Core 2.0

rockfordlhotka:
Tracked by #688

Also some minor tweaks to fix issues with #760


Source/Csla.Web.Mvc.Shared/ApplicationContextManager.cs
```diff
@@ -0,0 +1,137 @@
...
+      return HttpContext?.User;
+    }
+
+    /// <summary>
+    /// Sets the current principal.
+    /// </summary>
+    /// <param name="principal">Principal object.</param>
+    public void SetUser(System.Security.Principal.IPrincipal principal)
+    {
+      HttpContext.User = (ClaimsPrincipal)principal;
```
dazinator: What do you think about this upcasting? ASP.NET core kind of mandates that your principal class derives from `ClaimsPrincipal` but CSLA doesn't mandate this - and currently the only time you would be aware of the problem is when you hit this cast exception at runtime. Any ideas? My first thought was to use compilation directives to change IContextManager interface so that for asp.net core the method argument is ClaimsPrincipal (is there an IClaimsPrincipal?) rather than IPrincipal? Is that an option?

NuGet/Definition/AspNetCore MVC.NuSpec
```diff
@@ -25,10 +25,17 @@ Supports the creation of AspNetCore applications.
         <dependency id="Microsoft.AspNetCore" version="1.1.1" />
         <dependency id="Microsoft.AspNetCore.Mvc" version="1.1.2" />
       </group>
+      <group targetFramework="netstandard2.0">
+        <dependency id="CSLA-Core" version="[4.6.3-Beta10]" />
```
dazinator: Just curious about the exact version match on [4.6.3-Beta10] - do you have automation that updates this with each build? Is it not preferable to allow a range?

rockfordlhotka: Yes, there's automation that updates this when the NuGet packages are built.

Source/Csla.Web.Mvc.Shared/ApplicationContextManager.cs
rockfordlhotka: Yeah, I saw that and tried to get rid of the cast. As you note, it is required by ASP.NET Core.

I don't know what to think about that exactly. If we change IContextManager we'd really need to change all uses of IPrincipal through CSLA to use ClaimsPrincipal (or does it have an interface?), and that'd be kind of a Big Deal(tm)...

dazinator: Ok. Well this works for me. If we wanted too, perhaps in the future, we could write our own adaptor to adapt the IPrinciple to a ClaimsPrinciple (assuming it wont cast) on SetUser and then return the original IPrinciple on GetUser. Probably not necessary for the time being.

Issue #788, opened 2017-10-12
Update samples to CSLA .NET 4.7.100


tfreitasleal:
Update CSLA .NET dependencies.
Make each solution use its own packages folder.

rockfordlhotka:
@tfreitasleal do you want to run through the demos now that I've pushed the latest 4.7.100 pre-release to NuGet? I suspect some may need changing/updating because I split out the netstandard platforms from full .NET as a platform.

tfreitasleal:
I'm on it.

PR #789, opened 2017-10-13
Make each solution use its own packages folder (#788)

tfreitasleal:
Rocky,

Please keep the issue open waiting for CSLA .NET 4.7

### DIFF
--- a/Source/Csla.Shared/ApplicationContext.cs
+++ b/Source/Csla.Shared/ApplicationContext.cs
@@ -676,10 +676,13 @@ namespace Csla
       }
     }
 
-#if !ANDROID && !IOS && !NETFX_CORE && !NETSTANDARD2_0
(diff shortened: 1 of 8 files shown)


rockfordlhotka:
You want this PR to close _after_ 4.7, not part of 4.7?

Oh, I see, you want to get the references all updated to the NuGet containing 4.7.

Hmm. We can do that - but I'd like the PR to close right after so on the _Releases_ page here in the repo people would find the updated samples too. That should work.

rockfordlhotka:
@tfreitasleal one thing that occurred to me this morning, is that the changes I'm making to the CSLA-Core package probably have substantial impact on some samples.

#793 means that the CSLA-Core package now includes only
* .NET 4
* .NET 4.5
* .NET 4.6
* netstandard 2.0

PR #808, opened 2018-01-03
Delete ICloneable.cs

ratku:
this ICloneable  is Exists In  NETFX_CORE 2.0


rockfordlhotka:
@ratku It looks like you deleted the file rather than using a compiler directive to remove it from `NETSTANDARD2_0`, so several other target platforms no longer build.

Issue #793, opened 2017-11-06
Split platform-specific code into satellite assemblies


rockfordlhotka:
Support for netstandard 2.0 really means having _exactly_ one Csla.dll that targets netstandard 2.0. All per-platform functionality needs to be in per-platform satellite assemblies.

As of Feb 19, 2018, all platforms already meet this requirement except for full .NET. Some aspects of full .NET (COM+, WCF, etc.) need to be split out into a brand new satellite assembly that supports full .NET.

--

Original description:

Right now the NuGet package includes output for numerous platforms, including:

* netstandard 2.0
* iOS 
* Android
* UWP

The problem we appear to be facing is that the iOS, Android, and UWP `Csla.dll` outputs are _actually_ platform-specific, which is causing some issues when an iOS or UWP app references a business DLL built using netstandard 2.0.

This is because the `Csla.dll` used at runtime is set by the app, so that business class library ends up trying to use the platform-specific assembly, not the netstandard assembly. Because these assemblies are not identical, this is a problem.

ddobrev:
When is this bug going to be fixed? Xamarin.Forms is dropping PCL in its next release and it's vital that CSLA can work with .NET Standard 2.0 instead.

rockfordlhotka:
I'm experimenting with some options.

The challenge is that this is probably a _major_ breaking change for anyone not running the latest VS (15.5) and using NS2.0 for Xamarin and UWP.

The solution I've been testing (with success) is to change the CSLA-Core package so it only contains
* netstandard 2.0
* .NET 4.0
* .NET 4.5

All the other targets (.NET 4.6, Xamarin, UWP, ns 1.5, ns 1.6) are technically superseded by NS2.0 - but only for people running the latest and greatest.

As a result, if I go down this road (and I think it is the only option) the CSLA 4.7.100 release will not work for anyone running on older versions of Xamarin or UWP.

rockfordlhotka:
Note to self: NS2.0 is pretty awesome, but people _do still use full .NET_ and so we can't actually eliminate support for full .NET 4.6.1 - that's where a lot of interesting Windows-only things around security, COM+, and so forth all exist...

rockfordlhotka:
Now encountering this issue:

https://github.com/dotnet/standard/issues/300

It doesn't appear that there is currently a solution to this problem, at least until a bug gets fixed in full .NET (4.7.2?).

[...]

PR #817, opened 2018-03-15
Update NuGet packages for NS2

rockfordlhotka:
This PR splits CSLA into two forks within NuGet - one covers full .NET, the other netstandard.

In full .NET:
* CSLA-Core
* Windows Forms
* WPF
* ASP.NET (non-core)

In netstandard 2.0:
* CSLA-Core-NS
* UWP
* Xamarin
* ASP.NET Core

Basically all the "legacy" stuff that still runs on full .NET relies on a `Csla.dll` built for full .NET. All the newer runtimes that support netstandard now rely on a `Csla.dll` built for NS2.

A client written using one of the new platforms can communicate with an app server running full .NET via the data portal.

Tracked by #793


tfreitasleal:
Hi Rocky,

This PR needs to be merged.

Issue #844, opened 2018-04-20
Update samples to CSLA .NET 4.7.101


tfreitasleal:

rockfordlhotka:
@tfreitasleal I pushed a 4.7.101 prerelease to NuGet and that should allow you to do all the updates to the samples.

PR #850, opened 2018-04-24
Update samples to CSLA .NET 4.7.101 #844

tfreitasleal:


tfreitasleal:
@rockfordlhotka Gitter is unstable and it looks like messages I typed earlier are lost.
1) Don't forget to update the azure WCF portal for __ProjectTracker__
2) The WcfPortal configuration of __SimpleNTier__ is broken. As this is more than a plain WcfPortal, I didn't fix it, because I could break something else.

Issue #650, opened 2016-10-18
Better exception when incorrect CslaDataPortalProxyFactory is specified


rockfordlhotka:
From this forum post: https://github.com/MarimerLLC/cslaforum/issues/256

PR #918, opened 2018-07-25
Csla 650

ajj7060:
Resolving issue https://github.com/MarimerLLC/csla/issues/650

### DIFF
--- a/Source/Csla.Shared/DataPortal.cs
+++ b/Source/Csla.Shared/DataPortal.cs
@@ -735,7 +735,8 @@ namespace Csla
         {
           var proxyFactoryType = 
             Type.GetType(ApplicationContext.DataPortalProxyFactory) ?? 
-            throw new InvalidOperationException($"Unable to load DataPortalProxyFactory {ApplicationContext.DataPortalProxyFactory}");
+            throw new InvalidOperationException(
+              string.Format(Resources.UnableToLoadDataPortalProxyFactory, ApplicationContext.DataPortalProxyFactory));
 
           _dataProxyFactory = (DataPortalClient.IDataPortalProxyFactory)MethodCaller.CreateInstance(proxyFactoryType);
         }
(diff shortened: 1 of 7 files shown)


Source/Csla.Shared/Reflection/MethodCaller.cs
```diff
@@ -118,8 +118,10 @@ private static DynamicMethodHandle GetCachedMethod(object obj, string method, bo
 
     private static Dictionary<Type, DynamicCtorDelegate> _ctorCache = new Dictionary<Type, DynamicCtorDelegate>();
 
-    private static DynamicCtorDelegate GetCachedConstructor(Type objectType)
-    {
+    private static DynamicCtorDelegate GetCachedConstructor(Type objectType) {
```
rockfordlhotka: This might be picky, but putting the brace on the same line doesn't follow the conventions used in the CSLA codebase.

Source/Csla.Shared/DataPortal.cs
```diff
@@ -733,7 +733,10 @@ internal static void LoadDataPortalProxyFactory()
         }
         else
         {
-          var proxyFactoryType = Type.GetType(ApplicationContext.DataPortalProxyFactory);
+          var proxyFactoryType = 
+            Type.GetType(ApplicationContext.DataPortalProxyFactory) ?? 
+            throw new InvalidOperationException($"Unable to load DataPortalProxyFactory {ApplicationContext.DataPortalProxyFactory}");
```
rockfordlhotka: Exception text needs to come from a resource file, otherwise it can't be localized into other languages. 

I too love the new $"" capabilities, but the resource strings all use $0, $1, etc. type parameters and we should keep that consistent as well.

ajj7060: Ahh, sorry, I forgot about the localization requirements.  I'll make the changes needed.

Source/Csla.Shared/Reflection/MethodCaller.cs
ajj7060: I'll fix; following code style is important, I think R# just moved it back without me noticing.

ajj7060:
@rockfordlhotka I've committed changes which should resolve the issues above

ajj7060:
@rockfordlhotka is the anything else i need to correct for this to be accepted?

PR #936, opened 2018-08-24
Data Annotation: Use [ValidationContext]

Troncho:
With this change you always get [target] and [MemberName] upon initializing [ValidationContext].

.Net DataAnnotations: [GetValidationResult], [IsValid(object)] and [IsValid(object, ValidationContext)] are programmed so as to handle backward compatibility.


rockfordlhotka:
@Troncho you should be able to alter the indentation in the code on your machine (using the same branch you had for the original work), then push that branch to your fork - that'll update the PR in-place.

rockfordlhotka:
Did you mean to be making these changes to the V3-8-x branch rather than master?

I'm only doing 3.8.x releases for critical bug fixes.

rockfordlhotka:
This appears to be an abandoned PR, so I'm closing it.

Issue #961, opened 2018-10-30
Create data portal health/instrumentation endpoint


rockfordlhotka:
It might be interesting to explore the idea of a data portal endpoint that an org can (optionally) expose to return some basic health stats about the server-side data portal.

For example, the data portal could track:

* Last 100 calls, with performance data, exception counts, etc.
* Last n exceptions

PR #976, opened 2018-11-08
Data portal dashboard implementation

rockfordlhotka:
#961


rockfordlhotka:
I created this branch against the wrong base, so I'm just discarding it and the PR and starting over.

Issue #967, opened 2018-11-04
Read CSLA config values from .NET Core config subsystem


rockfordlhotka:
CSLA should have its own configuration section in `appsettings.json` file for this and further configurations.

Example (just a few of them):
```
{
   "csla": {
      "proxyType":"name",
      "authentication":"windows",
      "serializationFormatter":"MobileFormatter"
    }
}
```
Beside [fluent interface](https://github.com/MarimerLLC/csla/issues/966), existing configuration mechanism implemented in .NET Core can be used to read from `appsettings.json` and other sources supported by .NET Core.

rockfordlhotka:
My understanding is that the .NET Core config subsystem takes the names of your settings (like you have in the json @itgoran) and flattens them for use in ENV variables or other scenarios. I don't know the exact algorithm, but something like `csla-proxyType` I think.

They use a precedence scheme (I don't remember the specifics) so they'll read the value from the config file, ENV variables, and maybe other sources, and there are rules as to which values ultimately "win" - which overrides the others.

In the end, the code throughout CSLA itself uses `ConfigurationManager.AppSettings` to access all the values, using hard-coded config value names.

So it seems to me that what we need to do is load the values from the .NET Core config subsystem, then map those values into the pre-defined `AppSettings` dictionary for use by all the existing code.

That doesn't seem too difficult.

rockfordlhotka:
I assembled a list of the settings keys:

CslaAuthentication
CslaAuthorizationProvider
CslaAutoCloneOnUpdate
CslaDataPortalActivator
CslaDataPortalExceptionInspector
CslaDataPortalInterceptor
CslaDataPortalProxy
CslaDataPortalProxyFactory
CslaDataPortalReturnObjectOnException
CslaDataPortalUrl
CslaDefaultTransactionIsolationLevel
CslaDefaultTransactionTimeoutInSeconds
CslaIsInRoleProvider
CslaMobileFactoryLoader
CslaObjectFactoryLoader
CslaPrincipalCacheSize
CslaPropertyChangedMode
CslaPropertyInfoFactory
CslaReader
CslaSerializationFormatter
CslaWriter
dbProvider

itgoran:
Great @rockfordlhotka , I was doing exactly the same thing.
Do you agree with the idea to use json like syntax, without CSLA prefix, since it will be in csla section?
Example:
1. 
```
"csla": {
    "Authentication":"windows",
    "AuthorizationProvider":"...",
[...]

PR #994, opened 2018-11-14
#967 Load configuration from appsettings.json

itgoran:
#967 Load configuration from '**csla**' section of the appsettings.json file.
Add tests for loading configurations.

### DIFF
--- a/Source/Csla.Shared/Configuration/CslaConfigurationOptions.cs
+++ b/Source/Csla.Shared/Configuration/CslaConfigurationOptions.cs
@@ -25,7 +25,7 @@ namespace Csla.Configuration
     public string PropertyChangedMode
     {
       get { return ConfigurationManager.AppSettings["CslaPropertyChangedMode"]; }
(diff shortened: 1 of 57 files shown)


itgoran:
@rockfordlhotka 
I've updated tests with reset logic.

**Note**
At my machine, if I execute tests from `csla.netcore.test` twice in a row from Visual Studio 2017, they failed (without my changes).
Is this case at your side?

Source/Csla.Shared/Configuration/CslaConfigurationOptions.cs
```diff
@@ -23,7 +25,7 @@ public class CslaConfigurationOptions
     public string PropertyChangedMode
     {
       get { return ConfigurationManager.AppSettings["CslaPropertyChangedMode"]; }
-      set { ConfigurationManager.AppSettings["CslaPropertyChangedMode"] = value; }
+      set { ApplicationContext.PropertyChangedMode = (PropertyChangedModes)Enum.Parse(typeof(PropertyChangedModes), value); }
```
rockfordlhotka: I think that in this (and the other properties) we should also be setting the AppSettings[] value as you were before? Keep AppSettings up to date with the current configuration values?

Issue #1012, opened 2018-11-27
Consider extension method for IConfiguration


rockfordlhotka:
Instead of code like this to configure CSLA:

```c#
      var config = new ConfigurationBuilder()
             .AddJsonFile("appsettings.coresettings.test.json")
             .Build();
      CslaConfigurationOptions cslaOptions = new CslaConfigurationOptions();
      config.Bind("csla", cslaOptions);
```
perhaps we should have an extension method to allow something like this:
```c#
      var config = new ConfigurationBuilder()
             .AddJsonFile("appsettings.coresettings.test.json")
             .Build()
             .ConfigureCsla();
```
or
```c#
    config.ConfigureCsla();
```
This extension method would look something like this:
```c#
    public static IConfigurationRoot ConfigureCsla(this IConfigurationRoot config)
    {
      config.Bind("csla", new CslaConfigurationOptions());
      return config;
    }
```

rockfordlhotka:
@itgoran what are your thoughts on this idea? Does it make sense?

itgoran:
@rockfordlhotka 
This would be the natural step forward in improving CSLA configuration experience on the dotnet core.

Also, I would like to suggest that we add the extension method under the namespace
`Microsoft.Extensions.Configuration` since this would allow usage without additional namespaces.

rockfordlhotka:
I think it should go in `Csla.Configuration`. Over the years I've coopted Microsoft namespaces a couple times, and that generally doesn't turn out well.

rockfordlhotka:
@itgoran I'll do this, as I'm working on a similar change for aspnet core.

rockfordlhotka:
@itgoran another thought I had is that maybe there should be a `UseCsla()` method for use in the ASP.NET `Configure` method. In a console app you have an `IConfiguration` object, but in ASP.NET Core it seems like it is almost more natural to have an `IApplicationBuilder`?

Something like this:
```c#
    public void Configure(IApplicationBuilder app, IHostingEnvironment env)
    {
      if (env.IsDevelopment())
      {
        app.UseDeveloperExceptionPage();
      }
      else
      {
        app.UseExceptionHandler("/Error");
      }

      app.UseStaticFiles();
      app.UseCookiePolicy();

      app.UseMvc();

      app.UseCsla();
    }
```

itgoran:
@rockfordlhotka 

I had similar thoughts when I was making CSLA configuration for dotnet core.
[...]

Issue #982, opened 2018-11-13
Add extension method to hook CSLA into ASP.NET Core IServiceCollection


rockfordlhotka:
CSLA needs an extension method for the `IServiceCollection` type so it naturally fits into the Startup.cs flow in AspNetCore.

Something like this:

```c#
  public static class CslaExtensions
  {
    public static CslaConfiguration AddCsla(this IServiceCollection sc)
    {
      // add servicedescriptor to sc collection
      return sc;
    }
  }
```

rockfordlhotka:
Duplicate of #787?

PR #1013, opened 2018-11-27
Add extension methods for service and config integration in .NET Core

rockfordlhotka:
#982 #1012 

* Add extension methods for service and config integration in .NET Core
* Add DataPortalService and IDataPortalService to enable .NET Core DI integration

### DIFF
--- a/Source/Csla.Shared/Configuration/ConfigurationExtensions.cs
+++ b/Source/Csla.Shared/Configuration/ConfigurationExtensions.cs
@@ -1,0 +1,42 @@
+﻿#if NETSTANDARD2_0
+//-----------------------------------------------------------------------
+// <copyright file="ConfigurationExtensions.cs" company="Marimer LLC">
+//     Copyright (c) Marimer LLC. All rights reserved.
+//     Website: http://www.lhotka.net/cslanet/
+// </copyright>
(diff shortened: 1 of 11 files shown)


Source/Csla.Shared/Configuration/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+  {
+    /// <summary>
+    /// Add CSLA .NET services for use by the application
+    /// and enable CSLA .NET configuration.
+    /// </summary>
+    /// <param name="services">IServiceCollection object</param>
+    public static ICslaBuilder AddCsla(this IServiceCollection services)
+    {
+      services.AddSingleton<IDataPortalService, DataPortalService>();
+      return null;
```
itgoran: @rockfordlhotka 
What is an idea behind returning `null`. Maybe some default implementation of ICslaBuilder?

rockfordlhotka: That might be better. Right now I don't have anything to actually do in an ICslaBuilder - but that's the convention of what these methods should return. You are probably right though, it is probably better to return an empty class that implements the empty interface rather than null.

Issue #1053, opened 2019-01-03
Update nuspec files from licenseURL to license tag


rockfordlhotka:
NuGet is changing from the use of a licenseURL tag to a license tag. We need to update our nuspec files accordingly.

https://docs.microsoft.com/en-us/nuget/reference/nuspec#license

PR #1108, opened 2019-04-01
Update nuspec files to use license element

rockfordlhotka:
#1053 Update nuspec files to use license element


rockfordlhotka:
Closes #1053

Issue #1167, opened 2019-05-30
Add option to allow FieldManager.UpdateChildren to ignore IsDirty


rockfordlhotka:
As per https://github.com/MarimerLLC/cslaforum/issues/755 from @j055

1. Add public `UpdateAll` methods to `ChildDataPortal` to match the existing public `Update` methods
1. Add a parameter to the private `Update` method in `ChildDataPortal` to override the `IsDirty` check as you suggest
1. Add an `UpdateAllChildren` method to `FieldManager` that calls the new `UpdateAll` method in the data portal

rockfordlhotka:
@j055 do you want to add this change to the v4.x branch as well so it goes into 4.11? If not that's fine, but if your need is pressing that would get you the change faster.

j055:
Sorry, I need to fork from 4.x first. Please delete the PR. My head hurts!

rockfordlhotka:
Yeah, this git stuff can be confusing... I only figured out the whole PR-against-a-branch thing recently, but I did add it to the [contributor instructions doc](https://github.com/MarimerLLC/csla/blob/master/docs/csla-github-flow.md).

PR #1172, opened 2019-05-31
Added UpdateAllChildren to FieldManager

j055:
https://github.com/MarimerLLC/csla/issues/1167

Added 2 unit tests

Note:
```FieldDataManager.UpdateAllChildren``` uses ```ChildDataPortal``` directly, whereas ```UpdateChildren``` uses ```DataPortal``` class.

### DIFF
--- a/Source/Csla.Shared/Core/FieldManager/FieldDataManager.cs
+++ b/Source/Csla.Shared/Core/FieldManager/FieldDataManager.cs
@@ -650,6 +650,25 @@ namespace Csla.Core.FieldManager
       }
     }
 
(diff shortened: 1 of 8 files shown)


rockfordlhotka:
In this case I think I'm ok with not calling the `DataPortal` class. If we need to make that happen in the future it wouldn't be hard to do, but right now there's not a pressing need (that I can see) for expanding that part of the overall API.

rockfordlhotka:
@j055 do you want to add this change to the v4.x branch as well so it goes into 4.11?

PR #1195, opened 2019-06-29
Addresses #1194, adding an IAuthorizationContext.

### LINKED ISSUE #1194: Create IAuthorizationContext
Simplify unit testing of AuthorizationRule subclasses.  Currently the use of AuthorizationContext means that you need to create the auth rule instance, then create the AuthorizationContext to execute the rule; this makes is more challenging to alter the inputs provided by the context for various scenarios.

Create an IAuthorizationContext, like the IRuleContext created in #948 to allow for mocking of the instance used in the rule's Execute.

As described in the problem description, this can be worked around by writing more testing code and use of `Lazy<T>` to delay building of the AuthorizationContext, but this requires extra code to be written and maintained in the unit tests.

ajj7060:
Resolves #1194

### DIFF
--- a/Source/Csla.Shared/Rules/IAuthorizationContext.cs
+++ b/Source/Csla.Shared/Rules/IAuthorizationContext.cs
@@ -1,4 +1,11 @@
-﻿using System;
+﻿//-----------------------------------------------------------------------
+// <copyright file="IAuthorizationContext.cs" company="Marimer LLC">
+//     Copyright (c) Marimer LLC. All rights reserved.
+//     Website: https://cslanet.com
+// </copyright>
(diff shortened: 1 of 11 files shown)


Source/Csla.Shared/Rules/IAuthorizationContext.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
```
rockfordlhotka: This file needs the standard copyright header

ajj7060: Should be fixed now.

Issue #1196, opened 2019-06-29
Provide criteria to per-type authorization rules


ajj7060:
**Is your feature request related to a problem? Please describe.**
In complex authorization scenarios, there is sometimes not enough information available for a per-type rule to determine if permission can be granted.  For example, a user may have view access to orders at one site, but may create new orders at a different site.  The only option today is to always response the user can create orders, but throw an exception if attempted at the view only site.  

**Describe the solution you'd like**
Provide access to the criteria object, if any, via the AuthorizationContext, so that the rule may use that information to make more detailed determination.

**Describe alternatives you've considered**
To work around this currently, we are creating an interface that must be implemented by the business object's class.  Then, in the HasPermission call, we are new-ing up an instance of business object and setting special properties only there for the auth rule to use and passing it instead of the type.  The downside is that this does not happen when Csla invokes the per-type rule, only in our `static CanCreateObject(int customerId)` method.

**Additional context**
This is discussed in [this](https://github.com/MarimerLLC/cslaforum/issues/621) forum thread.

ajj7060:
@rockfordlhotka I'm going to take a stab at this, if you want to assign it to me.

ajj7060:
@rockfordlhotka I merged the changes from your branch into my branch, and made the appropriate changes.  I can't do a PR to your branch though, so I'm not sure where to go next with this.  you can find my changes at https://github.com/ajj7060/csla/tree/1196-AuthorizationContext_CriteriaProperty

rockfordlhotka:
Thanks @ajj7060 - I thought you could do a PR to my branch, but now I know better. I wonder how to collaborate like that then? Hmm...

ajj7060:
@rockfordlhotka I was able to merge into my branch from yours, so perhaps you could do the reverse?  Seems odd that they don't have a PR for this kind of thing though.

rockfordlhotka:
@ajj7060 I just tried to create a PR to pull your changes into my 1102 branch, but they are no longer compatible.

However, I'm about to merge my 1102 branch into master (whew!!!), and so you should be able to do a normal PR of your changes into master in a few minutes.

PR #1200, opened 2019-07-08
Addresses #1196 Add ability to pass criteria via authorization context

ajj7060:
Addresses #1196, including the criteria sent to the data portal through the authorization context for per-type rules so that auth rules may have more information to determine permissions.


ajj7060:
I think I'll wait until my other PR is incorporated, as I'll have to add the criteria to the interface.

PR #1203, opened 2019-07-10
#1196 Add Criteria property to IAuthorizationContext

### LINKED ISSUE #1196: Provide criteria to per-type authorization rules

ajj7060:
Resolves #1196, which adds a Criteria property to IAuthorizationContext, and updates the data portal to supply it for per-type rules.

### DIFF
--- a/Source/Csla.Shared/DataPortalT.cs
+++ b/Source/Csla.Shared/DataPortalT.cs
@@ -85,7 +85,7 @@ namespace Csla
       {
         DataPortal.OnDataPortalInitInvoke(null);
 
-        if (!Csla.Rules.BusinessRules.HasPermission(Rules.AuthorizationActions.CreateObject, objectType))
+        if (!Csla.Rules.BusinessRules.HasPermission(Rules.AuthorizationActions.CreateObject, objectType, criteria))
           throw new Csla.Security.SecurityException(string.Format(
             Resources.UserNotAuthorizedException,
             "create",
@@ -305,7 +305,7 @@ namespace Csla
       {
         DataPortal.OnDataPortalInitInvoke(null);
 
-        if (!Csla.Rules.BusinessRules.HasPermission(Rules.AuthorizationActions.GetObject, objectType))
+        if (!Csla.Rules.BusinessRules.HasPermission(Rules.AuthorizationActions.GetObject, objectType, criteria))
           throw new Csla.Security.SecurityException(string.Format(
             Resources.UserNotAuthorizedException,
             "get",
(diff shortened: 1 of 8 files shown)


rockfordlhotka:
@ajj7060 this looks good, but I have a fairly major concern. Specifically, a lot of the data portal code is changed by the enhancements in PR #1163 

One of the changes is that "criteria" is no longer necessarily a single value. It could be a list of values.

I really hate to ask this, but would you consider applying your changes to the https://github.com/rockfordlhotka/csla/tree/1102-dataportal branch instead of master?

If you look at the updated code (especially `DataPortalT`) you'll see that the flow is similar, but not identical. I'm _sure_ it is different enough that auto-merge will be impossible, and re-implementation would be required.

However, looking through your changes in this PR I really don't think it'll be too hard to apply your changes (in concept) to my working branch.

The one thing that might be a little tricky is that "criteria" really does flow through `DataPortalT` as type `object`, but before it is used it is always converted into a concrete value based on a set of rules such that it can be passed into methods as a parameter array (`params` keyword).

As an improvement over the old code, those rules are now codified into a couple helper methods (`GetCriteriaFromArray` and `GetCriteriaArray`):

https://github.com/rockfordlhotka/csla/blob/1102-dataportal/Source/Csla.Shared/DataPortalT.cs#L73

ajj7060:
@rockfordlhotka Ah, I didn't consider those changes when I did this.  I'll give it a shot.  I should be able to commit directly to that branch, correct?  Or is there some other git magic I need to learn to incorporate things? 😄 

Also, since there can be more than one criteria param, do you think the `IAuthoriztionContext.Criteria` property should be an `object[]`?  I assume that's the only way to go, unless there's something in that branch to use instead?

PR #1210, opened 2019-07-22
Fix Samples\SimpleNTier - WpfUI missing references

rob-v:
WpfUI project has 2 missing references: Microsoft.Expression.Interactions. dll and System.Windows.Interactivity.dll. I found 1st in Source\Dependencies\Expression\WPF (different path than in .csproj), 2nd isn't in csla repo.
This 2nd .dll could be added here and path to Dependencies folder fixed, but there is "Blend.Interctivity.WPF.v4.0" NuGet package which contains exactly these 2 .dlls so I used it.


rockfordlhotka:
Thank you @rob-v !

Issue #1212, opened 2019-07-22
CSLA 4.11.1 breaks deletions when root object is list


GreatBarrier86:
**Question**
The change to `UndoableBase` in [#1191](https://github.com/MarimerLLC/csla/commit/73f1e3dfdcfffb4e45b376fabef4ffab23d6adb0#diff-e0c293996c39ea1f85c5e8b80ac4b540) seems to have caused a new problem in my scenario. In order to delete an item from my list, i'm calling `DoRemove(myItem)` followed by `SaveAsync`. Since `ViewModelBase.SaveAsync` clones the list, and your change below removes `_stateStack` from being serialized in `UndoableBase.OnGetState` and `UndoableBase.OnSetState`, when `BusinessListBase.AcceptChanges` is called, the loop through `DeletedList` that calls `C.AcceptChanges` throws an exception because `C.EditLevel - 1` ends up being `-1` while `parentEditLevel` is `0`.

**In a similar style to my previous issue, see if this breaks for you as well**
```
DiariesVM lvm = DiariesVM.Fetch(); //Assume this creates a DiariesVM that has a BusinessListBase Model containing 2 Diary BusinessBase objects
lvm.DoRemove(lvm.Model[0]);
lvm.SaveAsync();
```

**Code removed from CSLA UndoableBase**
![image](https://user-images.githubusercontent.com/29482393/61657612-5b174d80-ac91-11e9-8030-436d0926b002.png)

**UndoableBase Failure Point**
![image](https://user-images.githubusercontent.com/29482393/61658498-7a16df00-ac93-11e9-898c-bab2015b785e.png)

**Version and Platform**
CSLA version: 4.11.1
OS: Windows
Platform: Xamarin

GreatBarrier86:
No worries! Glad the report made sense!

rockfordlhotka:
I believe I have this resolved for 4.11.2. That's the good news @GreatBarrier86 !

There was a change in 5.0 (#1080) that actually masks the underlying issue with `MobileFormatter` and its interaction with `UndoableBase`. As a result, I am having a hard time porting the fix forward to 5.0.0 - more research is required.

GreatBarrier86:
@rockfordlhotka, great to hear about 4.11.2. How does this cause problems for 5.0? I probably cant offer advice but i'm curious to know. the link you provided has a lot of stuff in it so i can't fully appreciate the context like you can.

rockfordlhotka:
[...]

PR #1217, opened 2019-07-30
Fix issues with undo and cloning

rockfordlhotka:
#1212 Fix issues with undo and cloning

### DIFF
--- a/Source/Csla.Shared/Core/FieldManager/FieldDataManager.cs
+++ b/Source/Csla.Shared/Core/FieldManager/FieldDataManager.cs
@@ -1,7 +1,7 @@
 ﻿//-----------------------------------------------------------------------
 // <copyright file="FieldDataManager.cs" company="Marimer LLC">
 //     Copyright (c) Marimer LLC. All rights reserved.
-//     Website: https://cslanet.com
+//     Website: http://cslanet.com
 // </copyright>
 // <summary>Manages properties and property data for</summary>
 //-----------------------------------------------------------------------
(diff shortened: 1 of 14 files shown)


Source/Csla.Shared/Core/FieldManager/FieldDataManager.cs
```diff
@@ -1,7 +1,7 @@
 ﻿//-----------------------------------------------------------------------
 // <copyright file="FieldDataManager.cs" company="Marimer LLC">
 //     Copyright (c) Marimer LLC. All rights reserved.
-//     Website: https://cslanet.com
+//     Website: http://cslanet.com
```
Hadronicus: The website link was reverted to http. This was probably a mistake.

Source/Csla.Shared/Core/FieldManager/FieldDataManager.cs
```diff
@@ -756,21 +688,15 @@ protected override void OnSetState(SerializationInfo info, StateMode mode)
...
         {
-          //string xml = info.GetValue<string>("_stateStack");
-          byte[] xml = info.GetValue<byte[]>("_stateStack");
-          MobileList<SerializationInfo> list = (MobileList<SerializationInfo>)MobileFormatter.Deserialize(xml);
-          SerializationInfo[] layers = list.ToArray();
-          Array.Reverse(layers);
-          foreach (SerializationInfo layer in layers)
-            _stateStack.Push(layer);
+          var stackArray = info.GetValue<byte[][]>("_stateStack");
+          _stateStack.Clear();
```
Hadronicus: _stateStack.Clear(); is being called twice in the same control flow. Once just before the **if** statement and once inside it. It looks like this is probably just a copy/past mistake from the implementation in UndoableBase.

I'm not sure if the call before the **if** statement or the one inside is more appropriate.

Source/Csla.Shared/Core/FieldManager/FieldDataManager.cs
rockfordlhotka: Thanks, will be fixed in another PR

Source/Csla.Shared/Core/FieldManager/FieldDataManager.cs
rockfordlhotka: Thanks, will be fixed in another PR

Issue #1165, opened 2019-05-29
Allow .NET 4.6+ to use new configuration model


tabaguley:
While upgrading our CSLA code base to 4.10.* (from 4.6.*), I ran into a breaking change due to addition of the CSLA ConfigurationManager.  However ,I don't want to report this as a 'Bug' because I may simply not understand the intent and proper usage of this object.

I am referencing the Csla.NetStandard2.0 project.

Here's my diagnosis and suggestion(s) for consideration:

* ConfigurationManager uses several '#if NETSTANDARD2_0' conditional declarations.  Not sure what the reason for this is, but it has resulted in my App.config settings not being loaded (from WinForms -- .NET 4.8 -- applications).  

* I removed (commented out) the '#if NETSTANDARD2_0' blocks, then added the System.Configuration.ConfigurationManager NuGet package (currently, v4.5.0; NS2.0 compliant) to the Csla.NetStandard2.0 project.

* Problem solved.  So given that System.Configuration.ConfigurationManager is available for .NET Standard 2.0, I would like to suggest removing the conditional declarations in ConfigurationManager and adding this NuGet package.

Thank you.

rockfordlhotka:
That is a good observation @tabaguley.

The implement as-is exists because most .NET Framework code pre-dates the new configuration model and so uses the old `System.ComponentModel` scheme with web.config and app.config.

The CSLA config subsystem is an abstraction that attempts to make it a transparent transition - your CSLA-based code works regardless of which underlying .NET config system you are using.

Clearly the implementation needs to accommodate your scenario however. So it is sort of a bug/enhancement depending on your viewpoint 😉 

I wonder how far back the new config model goes? Right now CSLA supports .NET 4, 4.5. and 4.6.1. So if the new config subsystem goes back to 4.6.1 it would be reasonably easy to make this work. If the new subsystem only goes back to 4.7.2 that'd require a new compile target for CSLA - very doable, but more work to be sure.

tabaguley:
I see that you've added this as a To-Do for Version 5.0.0... Thank you.  That would be reasonable, as I don't believe this package is NS2.0, which would only support back to .NET 4.6.1.

I also just made an edit to the Title & detail where I put the wrong package name.

PR #1222, opened 2019-07-31
Allow .NET 4.6.1 to use new configuration model

rockfordlhotka:
#1165


rockfordlhotka:
@tabaguley would you take a look at these changes and see if they fit your description?

Issue #1283, opened 2019-09-07
Blazor Validation Component


TheCakeMonster:
I think we need a validation component for Blazor that interacts with EditContext, very similar to the DataAnnotationsValidator that has been provided in the box. Whilst CSLA supports data annotations, to get the full power of dependent properties, multiple rulesets and the more advanced features that CSLA's validation offers, we need to go beyond what the validator built into .NET Core can support.

I have been working on this, and it's very straightforward, as CSLA already does the heavy lifting. I have working code, and I'm starting to look at how to integrate it into the Csla solution.

The code needs to reference packages that target .NET Standard 2.0, but which are related to .NET Core 3.0 by their version numbers. Csla.Blazor is not named in a way that would encourage me to do this - it's not named to be for a specific .NET Core release.

Should I add the code to a shared project, for example Csla.Web.Shared, and then reference the latter from a new .NET Core 3.0 project, so that we can have assemblies to target different versions of .NET Core in the future? Or is it your intention only to support a single version of .NET Core at a time?

The packages I would need to reference are Microsoft.AspNetCore.Components and Microsoft.AspNetCore.Components.Web, currently at version 3.0.0-preview9.19424.4

PR #1284, opened 2019-09-08
Blazor validation component, including new unit tests project

TheCakeMonster:
Blazor validation component that is compatible with Blazor's built in EditForm/EditContext editing mechanisms. This validator offers support for extended validation concepts available to CSLA objects, such as business rules that incorporate multiple properties.

This first implementation exposes all validation rules without considering the complexity of severity.

Submission associated with feature request #1283

### DIFF
--- a/Source/Csla.Blazor.Test/Fakes/FakePerson.cs
+++ b/Source/Csla.Blazor.Test/Fakes/FakePerson.cs
@@ -1,0 +1,79 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+using Csla.Blazor.Test.Rules;
+
+namespace Csla.Blazor.Test.Fakes
+{
+  [Serializable()]
+  public class FakePerson : BusinessBase<FakePerson>
+  {
+    public static PropertyInfo<string> FirstNameProperty = RegisterProperty<string>(c => c.FirstName);
+    public static PropertyInfo<string> LastNameProperty = RegisterProperty<string>(c => c.LastName);
+    public static PropertyInfo<string> HomeTelephoneProperty = RegisterProperty<string>(c => c.HomeTelephone);
+    public static PropertyInfo<string> MobileTelephoneProperty = RegisterProperty<string>(c => c.MobileTelephone);
+
+    #region Properties 
+
+    [MaxLength(25)]
+    public string FirstName
+    {
+      get { return GetProperty(FirstNameProperty); }
+      set { SetProperty(FirstNameProperty, value); }
+    }
+
+    [Required]
+    [MaxLength(25)]
+    public string LastName
+    {
+      get { return GetProperty(LastNameProperty); }
+      set { SetProperty(LastNameProperty, value); }
+    }
+
+    public string HomeTelephone
+    {
+      get { return GetProperty(HomeTelephoneProperty); }
+      set { SetProperty(HomeTelephoneProperty, value); }
+    }
+
+    public string MobileTelephone
+    {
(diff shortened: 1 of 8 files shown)


Source/Csla.Blazor/CslaValidator.cs
```diff
@@ -0,0 +1,35 @@
...
+
+  /// <summary>
+  /// Class offering integration of CSLA validation into Blazor edit forms
+  /// This is a component with no visible behaviour; it simply requests hook up of validation
+  /// </summary>
+  public class CslaValidator : ComponentBase
+  {
+    
+    [CascadingParameter]
+    public EditContext CurrentEditContext { get; set; }
```
rockfordlhotka: Is this the name the Blazor community has adopted? Not BindingContext or DataContext from the various XAML dialects?

Source/Csla.Blazor.Test/Fakes/FakePerson.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+using Csla.Blazor.Test.Rules;
+
+namespace Csla.Blazor.Test.Fakes
+{
+  [Serializable()]
+  public class FakePerson : BusinessBase<FakePerson>
+  {
+    public static PropertyInfo<string> FirstNameProperty = RegisterProperty<string>(c => c.FirstName);
```
rockfordlhotka: This is a small thing, but the new CSLA v5 snippets and recommendation is to exchange the `c=>c.Name` lambda for `nameof(Name)`

Source/Csla.Blazor.Test/Fakes/FakePerson.cs
```diff
@@ -0,0 +1,79 @@
...
+      base.AddBusinessRules();
+      BusinessRules.AddRule(new OneOfSeveralStringsRequiredRule(HomeTelephoneProperty, MobileTelephoneProperty));
+    }
+
+    #endregion
+
+    #region Data Access
+
+    [RunLocal]
+    protected override void DataPortal_Create()
```
rockfordlhotka: The new CSLA v5 approach is to use the `[Create]` attribute on this method. And I've stopped using the `DataPortal_` prefix on the method names too, as I think the code looks better without them.

Source/Csla.Blazor/CslaValidator.cs
TheCakeMonster: EditContext is the name of the built in class they have created for cascading down from their EditForm component, and CurrentEditContext is the name of the parameter in the built in DataAnnotationsValidator component on which mine is based.

It's consistent and interoperable with what the ASP.NET team have done; they seem to have chosen their own route here.

Source/Csla.Blazor.Test/Fakes/FakePerson.cs
TheCakeMonster: Ah, of course. I saw your issue about the potential change, but obviously failed to take it on board!

Ironically, I couldn't remember the exact syntax - I code generate at work, so I never write it by hand - so I copied this from somewhere else in the codebase.

I agree that the code should stay to stay up to date and be an example to others, so I'll create an issue for this and fix both this instance and the example of the old syntax I copied.

Source/Csla.Blazor.Test/Fakes/FakePerson.cs
TheCakeMonster: Ooh, more new functionality in the data portal! That one has passed me by - there was something like this in the code I copied, but it seems I didn't take that on board, either.

Maybe this change was in one of those raw HTML blog posts I have struggled with, hehehe! Yes, I'll change that.

Issue #1295, opened 2019-09-10
Enhanced Blazor ValidationMessage component to handle CSLA's severities


TheCakeMonster:
The Blazor ValidationMessage component included in .NET Core 3.0 displays all validation messages with the same style; at the moment there is no support in Blazor for the concept of rules with different severities, as we have in CSLA.

We could create a new Blazor component that allows messages of different severities to be visually distinct - probably by decorating the HTML generated with different CSS class names. This will allow developers to separate what messages are as a result of errors that prevent the form being submitted, from those that are warnings or information that do not prevent submission. Class names should be assigned to the component via parameters, so that they can be assigned by developers to suit their own naming conventions.

I would see this component being a replacement for the built in component in Blazor applications that use CSLA, where developers choose to make use of different rule severities. The existing component could still be used, but only where there is no requirement to segregate rules by severity.

Some options to consider; input is welcome

1) Use of CSS class names seems like a standard mechanism for offering this separation, and I think they can even be used to offer some accessibility guidance. Are there any Aria hints that might be an additional enhancement, to aid with the delivery of better accessibility?

2) Should messages of all severities be displayed at the same time, or should warnings and information only display when there are no errors?

3) Should we expose delegate parameters to which developers can bind boolean values to define whether (or when) they wish each of the lesser severities to display? Such parameters would be optional, so that the enhanced functionality they provide does not result in clutter for simpler use cases. Such parameters add complexity to the component, and are likely to increase the support and maintenance cost of it through its life, so it might be best to avoid this.

4) An alternative I considered is delivering a component that can be stacked or used in groups, so that people were able to use a separate component for each severity. To me, this seems like it might be overly complex and lead to very verbose Razor syntax in an edit form.

5) A more advanced option might be to offer a templated component, so that developers have complete control over the HTML used to output messages of each severity. However, this is likely to be too complex - and too verbose in Razor syntax - for most use cases. At best, this might be a more advanced component, provided in addition to a more basic one that we intend for the majority of circumstances. Perhaps this might be an appropriate method of delivering a ValidationSummary component at a later date?

[...]

PR #1306, opened 2019-09-16
Initial implementation of advanced ValidationMessage

TheCakeMonster:
An initial implementation of a CSLA-specific replacement for the built-in ValidationMessage component. This version outputs messages with Error, Warning and Information severities with different classes assigned, so that they can be styled differently.

Defaults for the class names for each severity are based on existing bootstrap styles, so that the component displays correctly when used in a new Blazor project with no custom styling.

Associated with issue #1295


Source/Csla.Blazor/BusinessRulesValidationMessages.razor.cs
```diff
@@ -0,0 +1,166 @@
...
+		private IEnumerable<string> GetBrokenRuleMessages(string propertyName, RuleSeverity severity)
+		{
+			IList<string> messages = new List<string>();
+			ICheckRules objectUnderTest;
+
+			// Attempt to gain access to the underlying CSLA object
+			objectUnderTest = CurrentEditContext.Model as ICheckRules;
+			if (objectUnderTest == null)
+			{
+				throw new ArgumentException("Model is not available, or does not implement ICheckRules");
```
rockfordlhotka: This string needs to come from a resource so it can be localized. I'll find an example to follow in this regard, as (iirc) we have all the resource strings flowing from `Csla.dll` to keep localization simple for everyone.

Source/Csla.Blazor/BusinessRulesValidationMessages.razor
```diff
@@ -0,0 +1,21 @@
+﻿@namespace Csla.Blazor
+@inherits BusinessRulesValidationMessageBase
+
+<div class="@WrapperClass" aria-live="polite">
```
rockfordlhotka: I wonder if we don't need 3 wrappers for more control over how the info/warn/error text is displayed?

Source/Csla.Blazor/BusinessRulesValidationMessages.razor.cs
TheCakeMonster: Yes, agreed.

CSLA is far better than .NET Core in this regard; all of the error messages in the built-in Blazor components seem to be hard-coded at the moment!

Source/Csla.Blazor/BusinessRulesValidationMessages.razor
TheCakeMonster: Yes, I did wonder that. I thought that the outer wrapper is important in getting the messages to all line up with one another, and putting another wrapper within it by type might be overkill.

Were you thinking instead of the current one, or in addition?

rockfordlhotka: I don't know the answer for sure - I'm not an html expert :)

But it seems like a UI might want to arrange info/warn/error in different ways (side-by-side instead of vertical, etc.) maybe?

So maybe there needs to be an outer-most wrapper, and then a wrapper for each type of message?

Source/Csla.Blazor/BusinessRulesValidationMessages.razor.cs
rockfordlhotka: Here's an example of pulling a resource string from `Csla.dll` in `Csla.Xaml` - so the technique should be the same for `Csla.Blazor`

https://github.com/MarimerLLC/csla/blob/master/Source/Csla.Xaml.Shared/TriggerAction.cs#L87

Source/Csla.Blazor/BusinessRulesValidationMessages.razor
TheCakeMonster: I'm intending to discuss this in more detail with our UI expert tomorrow. I think it has merit, but I'd value his input - he might be better at HTML than either of us, and is exactly the sort of person we are targeting with the change.

I started the conversations with him about the flexibility and accessibility of this component on Monday, but he's not in the office on even days, and needed some time to mull it over anyway. 

I'm hoping to complete the PR for this component before the drop on Monday, but it's sensible to leave it out if we have any doubts at all, as once it's there, it can't really change without breaking people.

Source/Csla.Blazor/BusinessRulesValidationMessages.razor.cs
TheCakeMonster: Hehehe. Well, I could make up a few translations, for a laugh ...

Hmm, despite knowing a bit of French and German, colloquialisms can always catch out the inexperienced, so US English it is.

TheCakeMonster: As an aside, for a conversation far in the future, it might be that _exception_ messages should not be in the user's language.

On the assumption that you shouldn't show exception details to an end user - because they will both confuse the non-technical, and inform the technical and mischievous, the destination for the exception is the logging system, for reading by the people who run the system, not use the system.

Developers are, in general, likely to be able to speak English. Class names are in English, and programming language features remain in English no matter where people are (I think?) As a result, the people who are looking in the logging system of a multilingual system, might be more likely to benefit from English than from the language of the _user_.

I'm not disagreeing that I should do this - I will do it, for consistency if nothing else - but it makes you think.

For better or worse, our sites are not multilingual - I asked that question, and was told not to make them so. But if they were, a Russian or Greek error message in the logging system - generated because the user's culture was one of those when they hit the website - would be of no help to me at all!

I think that's one to think about for CSLA 6.0. Or maybe 7.0. ;)

TheCakeMonster: I suppose, to complete this line of random thought at an inappropriate moment just before a major release, I should say that we are talking about there being two languages of importance, and that both of the languages we are talking about are represented in the system in which the application is executing.

From memory, I think the user's language is represented by CurrentCulture.UICulture. I think the developer's culture is fairly likely to be represented by the culture in which the base operating system was installed. However, in a cloud-based world, even _that_ isn't guaranteed!

Language is _hard_ - ridiculously hard. If only everyone spoke CSLALanguage. Ah, now _there's_ an idea ...

Source/Csla.Blazor/BusinessRulesValidationMessages.razor
TheCakeMonster: UI expert said he could imagine no need for this, but at the same time saw no reason _not_ to, so for the sake of flexibility, I'll make this change. Tick! Thanks.

TheCakeMonster:
I'm hoping I've just committed the requested changes correctly.

I haven't added to Resources.resx yet, for two reasons.
1) Mainly it's because I wasn't sure if changing a file that wasn't part of the original PR would break it. I'm still trying to get used to GitHub - even with your cheat sheet, I don't feel like it's anything like natural yet.
2) I'm also not sure what the best way of changing this file is, as there is a designer file in a separate part of the solution. I'm used to updating resources files via the UI, but I don't think this works in the Csla core because of the project type (but I might be wrong.) Probably manual edits of both files are in order, but I don't want to break anything, so I'm approaching it with caution.

What I will do instead is create a separate issue to go back and fix all of the error messages in this way, as there is at least one message elsewhere that also needs the same treatment.

Source/Csla.Blazor/BusinessRulesValidationMessages.razor.cs
TheCakeMonster: I've temporarily used the simpler solution, but raised another issue to add to the resources file for this and several other error messages.

Issue #1287, opened 2019-09-10
Update NuGet packages for .NET Core 3


rockfordlhotka:
Upon release of .NET Core 3 there will be a whole set of dependencies used by CSLA that need updating to those new versions. This issue is to track those updates.

PR #1330, opened 2019-09-24
Update dependencies for .NET Core 3.0 release

rockfordlhotka:
#1287


rockfordlhotka:
Looks like the build might be stalled for the moment by appveyor - we need to wait for it to catch up to the .NET Core 3.0 release.

Issue #1370, opened 2019-10-10
Blazor - Extend ApplicationContextManager to leverage AuthenticationStateProvider


dazinator:
**Is your feature request related to a problem? Please describe.**
As discussed in https://github.com/MarimerLLC/csla/issues/1142 several options were proposed. The issue was closed however before any were implemented.

I propose this feature will show how CSLA's blazor ContextManager.User and blazor's AuthenticationStateProvider can work together. Propose we extend the blazor sample to have an authenticated view, with a simple login / logout mechanism. This would show the pure mechanics of what needs to happen when you login and logout. Then from that we can work out how we refactor that into Csla's blazor support packages (or not) - i.e by extending ContextManager, or some other means.

**Describe alternatives you've considered**
As discussed on linked issue.

**Additional context**

rockfordlhotka:
I like this idea a lot.

I am working through a bunch of issues with aspnetcore, HttpContext, and user principal things right now for v5.1. Tricky stuff given how they've changed things - like HttpContext isn't automatically available or even initialized at certain points.

Working through these details in Blazor would be very beneficial.

@dazinator is this something you are interested in figuring out?

dazinator:
@rockfordlhotka  PR submitted! To be honest I am not sure there would be that much to pull into CSLA's blazor support library. See what you think. One question that does crop up (as per my notes on the PR) - is the issue of ClaimsPrincipal and MobileFormatter not playing together and whether that could possibly be improved.

PR #1372, opened 2019-10-11
#1370 - Blazor Sample show authentication and authorisation

dazinator:
This adds Authentication and Authorisation the blazor sample as suggested in #1370
There is a remaining issue I'll mention at the end.
First the changes:

- Added a item to the menu that displays "Login" or "Logout" based on your authentication status. When you click it, it navigates you to a new /login page.

image

- The login page, displays a (functional) Login or Logout button based on your authentication state:

image

- The existing "Add Person" menu item - now only displays after you login:

image

The corresponding "EditPerson" page is now protetced with an 'Authorize` attribute - meaning if you try to navigate directly to it, and you aren't logged in, it now will display this:

image

## Remaining Issue

Blazor's `AuthenticationStateProvider` works with `ClaimsPrincipal` and `ClaimsIdentity`. Now that the sample uses a ClaimsPrincipal - and sets Csla's ApplicationContext.User to this ClaimsPrincipal - the CSLA pages now error stating that the Principal doesn't work with MobileFormatter as it doesn't derive from a CSLA base class.

@rockfordlhotka given that `ClaimsPrincipal` is so staple in the landscape now, do you think it would be worth extending MobileFormatter to support ClaimsPrincipal directly?

### DIFF
--- a/Samples/BlazorExample/BlazorExample/Client/App.razor
+++ b/Samples/BlazorExample/BlazorExample/Client/App.razor
@@ -1,6 +1,10 @@
 ﻿<Router AppAssembly="@typeof(Program).Assembly">
   <Found Context="routeData">
-    <RouteView RouteData="@routeData" DefaultLayout="@typeof(MainLayout)" />
+    <AuthorizeRouteView RouteData="@routeData" DefaultLayout="@typeof(MainLayout)">
+      <Authorizing>
+       <p>Authorising..</p>
+      </Authorizing>
+    </AuthorizeRouteView>
   </Found>
   <NotFound>
     <LayoutView Layout="@typeof(MainLayout)">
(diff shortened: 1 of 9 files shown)


dazinator:
@rockfordlhotka
I just tried to rebase this branch on master to resolve the conflicts. I found some problems.
In `master` the csproj references this package:

```
    <PackageReference Include="Csla.Blazor" Version="5.1.0-R19100901" />
```

I can't restore that package version. Nuget is only able to restore the nearest match which is `5.1.0-R19101002`.

After doing so, the build fails locally with:

image

I though this might be due to it restoring the incorrect package version as describe above, so I aborted the rebase at that point. I'll come back and attempt to rebase again when things are a bit more stable.

dazinator:
I rebased on master. Master seems to have some issues so I fixed.
- The csproj in master had a package reference to Csla.Blazor that was outdated and not available on Nuget. I've corrected it.
- The `_imports.razor` file had `using` statement added for a Csla aspnetcore taghelpers package. This was causing a compilation error and as far as I could tell wasn't needed, so I have removed it.

rockfordlhotka:
Thank you, those were my bad - I must have accidentally checked in some work-in-progress code 😞

rockfordlhotka:
Regarding `ClaimsPrincipal`, CSLA does support it via a `CslaClaimsPrincipal`. The base type isn't serializable, so I created a way to wrap it and basically clone its contents. It should be a drop-in replacement for the base type, but can serialize to/from the client.

Issue #1386, opened 2019-10-22
Auto-create scoped DI service provider on data portal server


rockfordlhotka:
It might be possible for `Csla.Server.DataPortal` to automatically create/set a scoped DI service in `ApplicationContext` to help ensure things like DAL or database connection instances are scoped to the specific data portal call.

Today it is necessary to do this manually in the data portal controller (for example) before `Csla.Server.DataPortal` is invoked.

PR #1391, opened 2019-10-29
Create scoped service provider if one not already created

rockfordlhotka:
#1386 Create scoped service provider if one not already created

### DIFF
--- a/Source/Csla.Shared/Server/DataPortal.cs
+++ b/Source/Csla.Shared/Server/DataPortal.cs
@@ -74,8 +74,8 @@ namespace Csla.Server
       {
         if (ReferenceEquals(ApplicationContext.DefaultServiceProvider, ApplicationContext.ScopedServiceProvider))
         {
-          ApplicationContext.ScopedServiceProvider = 
-            ApplicationContext.DefaultServiceProvider.CreateScope().ServiceProvider;
+          ApplicationContext.ServiceProviderScope = 
+            ApplicationContext.DefaultServiceProvider.CreateScope();
         }
       }
 #endif
@@ -602,10 +602,11 @@ namespace Csla.Server
         _interceptor.Complete(e);
 
 #if !NET40 && !NET45
-      if (ApplicationContext.ServiceProviderScope != null)
+      var scope = ApplicationContext.ServiceProviderScope;
+      if (scope != null)
       {
-        ApplicationContext.ServiceProviderScope.Dispose();
         ApplicationContext.ServiceProviderScope = null;
+        scope.Dispose();
       }
 #endif
     }
--- a/Source/Csla.Shared/Server/DataPortal.cs
+++ b/Source/Csla.Shared/Server/DataPortal.cs
@@ -15,6 +15,9 @@ using System.Security.Principal;
 using System.Threading.Tasks;
 using Csla.Properties;
 using System.Collections.Generic;
+#if !NET40 && !NET45
+using Microsoft.Extensions.DependencyInjection;
+#endif
 
 namespace Csla.Server
 {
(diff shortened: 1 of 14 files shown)


Source/Csla.Shared/Server/DataPortal.cs
```diff
@@ -62,9 +65,20 @@ protected DataPortal(string cslaAuthorizationProviderAppSettingName)
     protected DataPortal(Type authProviderType)
     {
       if (null == authProviderType)
-        throw new ArgumentNullException("authProviderType", Resources.CslaAuthenticationProviderNotSet);
+        throw new ArgumentNullException(nameof(authProviderType), Resources.CslaAuthenticationProviderNotSet);
       if (!typeof(IAuthorizeDataPortal).IsAssignableFrom(authProviderType))
```
JasonBock: Curious....why does this constructor take a `Type`? Couldn't you use a constrained generic to ensure that the type is of a type `IAuthorizeDataPortal`?

Source/Csla.Shared/Server/DataPortal.cs
```diff
@@ -62,9 +65,20 @@ protected DataPortal(string cslaAuthorizationProviderAppSettingName)
...
-        throw new ArgumentException(Resources.AuthenticationProviderDoesNotImplementIAuthorizeDataPortal, "authProviderType");
+        throw new ArgumentException(Resources.AuthenticationProviderDoesNotImplementIAuthorizeDataPortal, nameof(authProviderType));
+
+#if !NET40 && !NET45
+      if (ApplicationContext.DefaultServiceProvider != null)
+      {
+        if (ReferenceEquals(ApplicationContext.DefaultServiceProvider, ApplicationContext.ScopedServiceProvider))
+        {
+          ApplicationContext.ScopedServiceProvider = 
+            ApplicationContext.DefaultServiceProvider.CreateScope().ServiceProvider;
```
JasonBock: I'm not sure what `ScopedServiceProvider` does....I'm assuming there's somewhere where you say the scope is "finished"? Do you dispose it somewhere else? And how long is this "scope" around for - the entire time a root operation executes (so child objects would get this same scope)?

rockfordlhotka: The data portal uses `ScopedServiceProvider` to do all its DI resolutions. If `ScopedServiceProvider` isn't explicitly set it'll default to the value of `DefaultServiceProvider`, which is the value captured (typically) in `Startup.cs` in the `UsingCsla` method call.

Your question about disposing the scope is good, because the answer is NO, right now there's no code to dispose the scope. I guess I assumed (incorrectly?) that it'd get GCed when the server call is complete and the reference goes out of scope.

So really then, after the server-side data portal is done with its work, and before returning to the caller, it should dispose the scoped context (if it is different from the default context).

JasonBock: It absolutely should dispose the scope :).

If this works the same as how scopes work in Autofac, then what it does is it'll hold on to references that are `IDisposable`-based. On disposal, the scope will then call `Dispose()` on all of those dependencies, because the user of the dependencies should not dispose them. If the scope isn't disposed, then if there is a finalizer in the scope (which there should be) then they'll get cleaned up there, but that's far from ideal.

So yes, you absolutely should dispose the scope.

rockfordlhotka: Well that was harder than expected, as I had to add a way to keep track of the `IContextScope` as well as the scoped service provider. Maybe I should track the scope _instead_ of the service provider? Hmm...

Source/Csla.Shared/Server/DataPortal.cs
rockfordlhotka: This is old code that works and I see no value in changing. The only reason for this at all is to enable some unit testing scenarios.

PR #1403, opened 2019-11-05
debug config for Csla.Web.Net4.5 & Csla.Web.Net4 missing constants

bradtwurst:


rockfordlhotka:
Thank you @bradtwurst - oversight on my part and I'm glad you caught/fixed it!

Issue #1318, opened 2019-09-20
Add Analyzers for Business Rules


JasonBock:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I think there's a couple of analyzers that could be added to CSLA related to business rules:

* A rule that doesn't call `AddOutValue()` or one of the `AddXyzResult()` methods on the given `RuleContext` within `Execute()` seems weird. This arguably should be flagged. No code fix for this one. If so, what should be the severity: informational, warning, or error?
* If a rule is asynchronous and derives from `BusinessRule`, it should be changed to `BusinessRuleAsync`, a new type that was recently added and should be used for asynchronous rules. This should have a code fix that will automatically change the type. What should be the severity: informational, warning, or error?
* All asynchronous rules must call `Complete()` on the given `RuleContext` within `Execute()`. It might be possible to create a code fix that adds a call to `Complete()` as the "last thing" the rule does. Flag as an error.

**Describe alternatives you've considered**
None

**Additional context**
None

JasonBock:
I was talking to @rockfordlhotka about these analyzers, and here's where I'm at.

The first one that changes an async `BusinessRule` to `BusinessRuleAsync` is now in the branch. However, `BusinessRuleAsync` still has `IsAsync` and `Complete()` on them. My question is, do these members even matter with `BusinessRuleAsync`? 

I think Rocky said he was going to check, but I'm thinking that those members are irrelevant now. That is, there should be no reason to set `IsAsync` to `true` in the constructor and call `Complete()` in `ExecuteAsync()`. By its nature, it **is** asynchronous, so there's no need for them. In fact, if this is true, what I could do is check for usages of these members and flag them for the developer, as they aren't necessary. I may also be able to provide a code fix to remove them.

As a side note, over time these members may be marked as obsolete over a couple of versions of CSLA, but again, it depends on how "true" asynchronous rules are handled.

rockfordlhotka:
### IsAsync

I think the trick is whether the rule is implemented through `IBusinessRuleAsync` or as a subclass of `BusinessRuleAsync`.

The `BusinessRuleAsync` type always returns `true` for `IsAsync`, because ... well, it is obvious right? There's no guarantee that someone manually implementing `IBusinessRuleAsync` would do the same thing - but I rather think that this is an edge case not worth worrying about with an analyzer.

### Complete method

[...]

Issue #1351, opened 2019-09-28
Analyzer: Update Code Fix Tests


JasonBock:
**Is your feature request related to a problem? Please describe.**
None

**Describe the solution you'd like**
I've noticed that the way I check for changed text in the code fix tests seems brittle. It doesn't always catch the changes I'm expecting. I need to re-evaluate how I'm doing the changed text evaluations so it's consistent.

**Describe alternatives you've considered**
None.

**Additional context**
None.

PR #1418, opened 2019-11-30
#1318 add business rule analyzers

JasonBock:
#1318


JasonBock:
Sigh.....it's because the changed text seems to change slightly when it's run in AppVeyor. Since #1351 is going to address code fix tests and expected text changes, I'm going to comment this out for now (I know, not pretty) and address the discrepancy there.

Issue #1424, opened 2019-12-02
Add ApplicationContextManagerTls type


rockfordlhotka:
There are some challenges with adapting existing apps to the new `AsyncLocal` approach for managing CSLA context (changed in #729). For example: https://github.com/MarimerLLC/cslaforum/issues/849

To help provide an easy workaround for this, perhaps CSLA should have an `ApplicationContextManagerTls` type - the original implementation using TLS - as part of the framework. It wouldn't be the _default_, but it would be in the framework, making its use as easy as a config change during app startup.

Issue #1417, opened 2019-11-19
Sync context disappears after async call causing DataPortal.Complete method to fail.


kellyethridge:
When using the `LocalProxy` synchronization fix and we flow the context everything works fine until the `Csla.Server.DataPortal.Complete` method call. This is because the `ApplicationContext.ClientContext` has been cleared from the `ConfigureWait(false)` methods used in the calling code in `Csla.Server.DataPortal`.

I'm using version 5.0.1 in VS2019, though this issue existed in the 4.x version as well. I'm using ASP.NET and I haven't heard about this issue from others, so I might be using CSLA completely wrong. In which case, I would like to correct my usage. I've been able to use CSLA because I use a custom built version.

I've created a repo that shows the error. If you were to compile a version of CSLA that does not use `ConfigureWait(false)` or does not call the `ApplicationContext.ClientContext` in the `DataPortal.Complete` method, the repo works correctly. When the project is run, the failure is shown in the browser developer tools network tab. I use Chrome for this.

Repo: https://github.com/kellyethridge/CslaApplicationContextTest

Thank you.

kellyethridge:
Hi Rocky,

I saw your commit and wanted to compile and try it out. Unfortunately, it seems this issue was not resolved. I saw how you changed to `Task.Factory.StartNew`, but that didn't seem to address this issue. It still all stems back to the `.ConfigureAwait(false)` calls in `Server.DataPortal`. I am wondering if a possible fix is to use the `CslaFlowSynchronizationContext` `appsetting` as a parameter to the `ConfigureAwait` calls? That way by default it would be configured with `false` as it is now and when set to `true` it will align with the `LocalProxy` usage of `TaskScheduler.FromCurrentSynchronizationContext`.

What are  your thoughts?

Thank You!

kellyethridge:
Hi Rocky,

I tried your 5.1.0 pre-release and as I feared, the issue still persists. I was wondering if you're uninterested in my possible fix or if I could create the fix and pull-request? Or if you have other ideas in mind.

Thanks, Rocky!

rockfordlhotka:
You tried it in the one I released today?

I did a bunch of work around that, and solve the issue in my test cases. I wonder what is different in your test cases?

rockfordlhotka:
I suspect the issue might be more complex than just changing the `ConfigureAwait` parameter.

[...]

Issue #1409, opened 2019-11-12
CslaClaimsPrincipal should honor roles from CslaIdentity


rockfordlhotka:
The base `ClaimsPrincipal` _appears to have_ a constructor that accepts an `IIdentity` instance, but in reality it converts that instance into a `ClaimsIdentity` and loses any custom property data in the process.

As a result, when a `CslaIdentity` is used to initialize a `CslaClaimsPrincipal` we should make an effort to automatically copy the `Roles` list from the identity into the newly created `ClaimsIdentity` to preserve those role values.

Issue #1396, opened 2019-11-01
Update use of iconUrl in nuspec files (it is deprecated)


rockfordlhotka:
https://aka.ms/deprecateIconUrl

PR #1425, opened 2019-12-02
Blazor enhancements and other fixes

### LINKED ISSUE #1312: Add Blazor UI to ProjectTracker solution

rockfordlhotka:
#1312 Add Blazor UI to ProjectTracker
#1409 CslaClaimsPrincipal enhancement
#1410 Add ModelChanged and ModelChanging events to Blazor ViewModel
#1396 IconUrl changes in NuSpec files
#1417 Fix issue with sync context in local data portal
#1424 Add ApplicationContextManagerTls type
#1435 Add authorization helpers for Blazor


Samples/BlazorExample/BlazorExample/Client/Pages/EditPerson.razor
```diff
@@ -33,17 +32,18 @@ else
         <td>Id</td>
         <td>@vm.Model.Id</td>
       </tr>
+      @if (vm.GetPropertyInfo<string>(nameof(vm.Model.Name)).CanRead)
```
TheCakeMonster: Have you considered using another component here in addition, a templated component?

We could combine a number of Blazor features to make the remaining markup very simple; templated components, a generic type parameter, and a cascading parameter.

The steps would be:

1) Create another component, let's say it's called EditRow. 
2) Into EditRow, copy this whole @if section
3) Replace <TextInput> with @ChildComponent, a parameter you define of type RenderFragment within EditRow. 
4) Add a generic @typeparam declaration for T
5) Add another parameter called For of type PropertyInfo<T>
6) Add a <CascadingValue> component to EditRow to cascade the value of @For into child components
7) Change TextInput to receive the PropertyInfo as a CascadingParameter rather than Parameter

If you did this, your markup for each property to be edited would be reduced to 3 lines:

<EditRow For="vm.GetPropertyInfo<string>(nameof(vm.Model.Name))">
    <TextInput>
</EditRow>

As well as simplifying the markup, I think this probably also reduces memory allocations, as you only call GetPropertyInfo() once, rather than 3 times.

Samples/BlazorExample/BlazorExample/Client/Pages/EditPerson.razor
```diff
@@ -62,10 +62,20 @@ else
   [Parameter]
   public string id { get; set; }
 
-  protected override async Task OnInitializedAsync()
+  protected override void OnInitialized()
   {
     vm.Saved += () => NavigationManager.NavigateTo("listpersons");
+     vm.ModelChanging += (o, n) =>
```
TheCakeMonster: As this exists for teaching other, less experienced users, might it be an aid to understanding to use the names old and new for the parameters here?

Samples/BlazorExample/BlazorExample/Client/Shared/TextInput.razor
```diff
@@ -0,0 +1,12 @@
+﻿<div>
+  <input @bind-value="Property.Value" @bind-value:event="oninput"
```
TheCakeMonster: It'll be interesting to see how oninput performs when using server-side Blazor and the connection is across the internet. I haven't had a server to deploy on and test this yet.

Samples/BlazorExample/BlazorExample/Client/Shared/TextInput.razor
```diff
@@ -0,0 +1,12 @@
...
+         disabled="@(!Property.CanWrite)" /><br />
```
TheCakeMonster: If PropertyInfo exposed an IsValid property, you could add the following here, as a nice accessibility enhancement:

aria-invalid="Property.IsValid"

aria-invalid tells screen-readers to inform the user that the input is not valid

Samples/ProjectTracker/ProjectTracker.Ui.Blazor.Server/App.razor
```diff
@@ -0,0 +1,12 @@
+<Router AppAssembly="@typeof(Program).Assembly">
+  <Found Context="routeData">
+    <AuthorizeRouteView RouteData="@routeData" DefaultLayout="@typeof(MainLayout)" />
+  </Found>
+  <NotFound>
+    <CascadingAuthenticationState>
```
TheCakeMonster: The location of the CascadingAuthenticationState component surprises me here. I'd expect that to be useful in the event that a route is found, and not useful in the event that it is not found. So isn't it in the wrong template/section?

Samples/ProjectTracker/ProjectTracker.Ui.Blazor.Server/Pages/ProjectEdit.razor
```diff
@@ -0,0 +1,229 @@
...
+    }
+    else if (viewMode == SubViewModes.Select)
+    {
+      @if (_resourceList == null)
+      {
+        <p>Loading resource list...</p>
+      }
+      else
+      {
+        <a href="javascript: void(0);" @onclick="ResetResourcesView">Cancel assignment</a>
```
TheCakeMonster: I'm not sure, but might the new preventDefault feature added in .Net Core 3.1 avoid the need for the javascript in anchor tags?

Samples/ProjectTracker/ProjectTracker.Ui.Blazor.Server/App.razor
rockfordlhotka: I thought so too, but I copied this from the latest 3.1 preview templates installed this week. I agree that it seems odd, but it also appears to work...

Samples/BlazorExample/BlazorExample/Client/Shared/TextInput.razor
rockfordlhotka: All this work is using server-side Blazor, but on my local network, so I don't really have a sense for performance. Michael Washington has been doing a lot of SSB over the Internet though, and says it is amazingly fast and responsive.

Samples/BlazorExample/BlazorExample/Client/Pages/EditPerson.razor
rockfordlhotka: You'd mentioned this in gitter. And I do think it is an excellent idea - I just haven't figured out how to create such a component. Do you have an example or reference?

Samples/ProjectTracker/ProjectTracker.Ui.Blazor.Server/Pages/ProjectEdit.razor
rockfordlhotka: I hope so! I did try that feature last week and it didn't work, which is why I'm still using the javascript. I should probably try again now that they've released a new preview.

Samples/ProjectTracker/ProjectTracker.Ui.Blazor.Server/Pages/ProjectEdit.razor
```diff
@@ -0,0 +1,229 @@
...
+    else
+      await vm.RefreshAsync(int.Parse(id));
+  }
+
+  private async void SelectResource()
+  {
+    viewMode = SubViewModes.Select;
+    _resourceList = (await ProjectTracker.Library.ResourceList.GetResourceListAsync())
+                      .Where(r => !vm.Model.Resources.Contains(r.Id)).ToList();
+    StateHasChanged();
```
TheCakeMonster: I wouldn't have expected that StateHasChanged was needed in this sort of scenario. This method is being triggered by an event handler in the same component. I thought it was only needed when a child component raises an event that requires a parent component to re-render itself?

It shouldn't do any harm, but it might be a line of code that isn't needed, that's all.

Samples/ProjectTracker/ProjectTracker.Ui.Blazor.Server/Pages/_Host.cshtml
```diff
@@ -0,0 +1,23 @@
...
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>ProjectTracker.Ui.Blazor</title>
+    <base href="~/" />
+    <link rel="stylesheet" href="css/bootstrap/bootstrap.min.css" />
+    <link href="css/site.css" rel="stylesheet" />
+</head>
+<body>
+  <app>
+    @(await Html.RenderComponentAsync<App>(RenderMode.ServerPrerendered))
```
TheCakeMonster: In .NET Core 3.1, I think this can be replaced with use of the <component> tag helper, but I think both work.

Samples/BlazorExample/BlazorExample/Client/Pages/EditPerson.razor
TheCakeMonster: Oh, I'm just an ideas man. Examples are implementation details! ;-)

Yep, I'm just updating Visual Studio, and then I'll see if I can knock something up and upload it to GitHub.

TheCakeMonster: Here you go. I think this does the job.

https://github.com/TheCakeMonster/Examples/tree/master/Blazor/TemplatedComponent/TemplatedCSLAEditRow

Samples/BlazorExample/BlazorExample/Client/Pages/EditPerson.razor
```diff
@@ -62,10 +62,20 @@ else
...
+    {
+      if (o != null)
+        o.PropertyChanged -= async (s, e) => await InvokeAsync(() => StateHasChanged());
+      if (n != null)
+        n.PropertyChanged += async (s, e) => await InvokeAsync(() => StateHasChanged());
```
TheCakeMonster: I've been wondering about the efficiency of calling StateHasChanged at this level. This is marking the top-level component as needing re-rendering on every change to any property. I think for most small to medium edit forms that would be fine, but on a very large edit form with lots of properties, this would make the re-render and diffing work harder than necessary.

I wonder if an alternative is to push the choice to re-render to a lower-level component, and only calling StateHasChanged within that child if the property with the which the component is associated has changed, as I believe we are passed the property name in the EventArgs of PropertyChanged.

The performance difference might be very small, and this might therefore be unnecessary. However, I seem to remember hearing of people having problems with React and/or Angular being too slow when an update is made on a very complex component graph. That was a JS performance issue, but at the moment Mono isn't any faster than JS; indeed, it appears it's about half the speed of JS on average because Mono is currently operating in interpreted mode when running on wasm.

Issue #1436, opened 2019-12-13
CslaValidator for Blazor treats warning and info severities as validation failures


TheCakeMonster:
There's a bug in the extension methods associated with/called by the CslaValidator component in Csla.Blazor that causes submission of an edit form to be blocked despite there being no broken rules with a severity level of Error, but only those with Warning or Information.

I've created a component that uses Blazor's built-in EditForm component to offer editing of a Csla object. The object includes custom business rules that add to the BrokenRules collection with failures at severity levels of warning and information. When these rules fail, the submission of the EditForm is blocked. CslaValidator ignores severity level when determining the messages that must be written to the ValidationMessageStore.

I think the expected behaviour would be to have warning and information messages displayed when they are present, but for it still to be possible to submit the form (i.e. for the EditForm component to be able to raise the OnValidSubmission event.)

**Version and Platform**
CSLA.Blazor Nuget package version: 5.1 latest preview - 5.1.0-R19110701
OS: Tested on .NET Core on Windows, but likely to affect all runtimes
Platform: Blazor server-side (and probably Blazor wasm as well)

**Code that Fails**
Unable to share the code with which I found the bug due to contractual obligations, but I will add unit tests which reproduce the issue to the PR which I intend to submit to resolve it.

**Stack Trace or Exception Detail**
No exception is being raised; this is simply causing incorrect behaviour when using the Blazor EditForm component.

**Additional context**
Code in EditContextCslaExtensions (in both the ValidateModel and ValidateField methods) adds all broken rules to the ValidationMessageStore without first considering their severity levels. Broken rules at severities of warning and information should be filtered out, so that only failures of severity Error cause the object to be considered invalid by EditForm, and submission of the form to be blocked.

This will have the knock-on effect of Blazor's built-in ValidationMessage component no longer showing message text for failures other than with severity levels of Error. However, we have already built the CslaValidationMessages component correctly displaying all severities with styling to make them look different, so this change to the behaviour of the old, built-in component seems acceptable.

Please feel free to assign this issue to me so that I can work on it - especially since I caused it!

PR #1437, opened 2019-12-14
Correction to CslaValidator to resolve issue #1436

TheCakeMonster:
Small change to CslaValidator to make it respect business rule severities. The original version of my code considered all broken rules to be validation failures. After this update, only broken rules with severities of Error are considered validation failures which should prevent form submission.

Also includes some extra unit tests and fixes to existing tests which were not exercising the correct part of the codebase. There is also localisation of one final exception message that was missed in a previous PR.

Resolves defect identified in issue #1436


rockfordlhotka:
@TheCakeMonster it looks like there were some dependency versioning issues. I have accepted a different PR that should fix them, so hopefully if you re-pull and merge master into your PR it'll fix the build issue.

TheCakeMonster:
@rockfordlhotka I think I've done what you suggested, but it appears to have made no difference. It appears that the unit tests are failing on the build server, despite all of them passing on my local machine.

I tried upgrading the tests project to .NET Core 3.1 in case that helped, but it continues to fail.

Just at the moment, I'm a bit stumped as to what the problem might be.

TheCakeMonster:
@rockfordlhotka I've been researching AppVeyor unit test failures and it's sparked a thought: do CSLA business rules run asynchronously now? I'm wondering whether the failures could be triggered by a timing issue, or something - but I am clutching at straws.

I can't get the tests to fail locally, and a lack of an ability to reproduce the problem in conjunction with limited knowledge of AppVeyor makes it hard to resolve.

TheCakeMonster:
Aha! Sorted it. 

It took a while to come up with a relatively simple idea, but I eventually twigged that I could use the console output from AppVeyor to output additional debug information to help me identify the problem.

It looks like the problem was caused by a missing resources entry for the MinLength rule. I'm not sure what culture AppVeyor builds would occur under, but there's something missing for whatever culture it is, and that causes the rule to fail, even though the rule conditions are met.

Never seen that before.

Sorry for the delay. I hope you don't pay per build!

rockfordlhotka:
Wow @TheCakeMonster - thank you for your dedication to solving the build issue!

(fortunately appveyor is free for open source projects)

Issue #1459, opened 2020-01-16
Poor performance when fetching large data set in CSLA 5.0.1


ka-fim:
**Describe the bug**
Fetching a large (~10k) data set using CSLA 5.0.1 takes several seconds (5-6). Doing the same with version 4.11.2 takes less than one second (~200ms). For a minimal example see https://github.com/ka-fim/CSLA-Performance

**Version and Platform**
CSLA version: 5.0.1
OS: Windows
Platform: .Net Framework 4.7.2

rockfordlhotka:
This was always a possibility due to the major rewrite of the method invocation code to implement method-based DI and some of the other new (and exciting) data portal features in CSLA 5.

I didn't want to optimize until the design and functionality was firm, which it now is. Optimization is clearly necessary, and is the next step.

Most likely, the initial optimization should be to build a method caching scheme similar to (but hopefully cleaner than) the one in CSLA 4. 

I think it'll be somewhat easier to make it clean this time because of the new way the data portal finds target methods. Right now it uses a relatively complex algorithm that (iirc) makes 3 passes to find potential matches, to score the potential matches, and then to pick the one with the highest score.

Creating some sort of cache with a resolution map should allow bypassing that entire process after the first invocation of a given method with a specific set of parameter types.

tabaguley:
I'm curious why it wasn't possible to retain the CSLA 4 methodology (which worked perfectly well), and add the new method-based DI functionality as an _option_??  It would be nice if the DataPortal went directly to hard-coded methods (ala CSLA 4-style) unless a configuration flag was set, which would then trigger some other way of doing things (for those users who may need such a feature).  

We've actually been considering down-grading back to CSLA 4's way of doing things simply because it worked so much better (and we've been delaying our role-out of numerous application updates until we could find and fix the performance issues of CSLA 5).  

Rocky, any idea how long it will take to get CSLA 5 optimized?  OR should we just merge the CSLA 4 DataPortal bits back into our CSLA 5 code-base?  Thank you.

rockfordlhotka:
@tabaguley if you are willing to do all that work, you might consider helping to implement the caching for the CSLA 5 implementation instead?

tabaguley:
I'm definitely willing to assist.  However, I have a few concerns I'd want to discuss with you first.  Is there a way to send you a private message via GitHub?  Or may I e-mail you?  I don't want to muck up this thread with a bunch of off-topic business.

rockfordlhotka:
[...]

PR #1466, opened 2020-01-20
Improve CSLA 5 data portal performance

rockfordlhotka:
#1459 

* Add caching to [FindDataPortalMethod](https://github.com/MarimerLLC/csla/blob/master/Source/Csla.Shared/Reflection/ServiceProviderMethodCaller.cs#L52)
* Add data portal perf test project

### DIFF
--- a/Samples/DataPortalPerf/DataPortalPerf/RetrieveEntities.cs
+++ b/Samples/DataPortalPerf/DataPortalPerf/RetrieveEntities.cs
@@ -1,0 +1,12 @@
+﻿using BenchmarkDotNet.Attributes;
+using Csla;
+
+namespace DataPortalPerf
+{
+  [MemoryDiagnoser]
+  public class RetrieveEntities
+  {
+    [Benchmark(OperationsPerInvoke = 100)]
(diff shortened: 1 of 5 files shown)


rockfordlhotka:
@JasonBock if you have time to take a look at the changes I think it is all stabilized now - I'm done, and have merged in master (which had a change that could affect outcomes). Thank you!!

Samples/DataPortalPerf/DataPortalPerf/RetrieveEntities.cs
```diff
@@ -0,0 +1,12 @@
+﻿using BenchmarkDotNet.Attributes;
+using Csla;
+
+namespace DataPortalPerf
+{
+  [MemoryDiagnoser]
+  public class RetrieveEntities
+  {
+    [Benchmark(OperationsPerInvoke = 100)]
```
JasonBock: I've never used `OperationsPerInvoke`, and I'm trying to figure out why you'd use it. I'm not saying it's wrong, just curious why this was done.

PR #1475, opened 2020-01-31
Blazor Example updated for both Mock and EF core implementation

SachinPNikam:
hello,
I udated the Blazor Example for EF core as well as mock .so we can use Database or mock based on startup code.
Thank  you


SachinPNikam:
attched person table script, let me know if the helpful , Thanks 
[personTable.zip](https://github.com/MarimerLLC/csla/files/4138876/personTable.zip)

rockfordlhotka:
Thank you for the PR, I'm on holiday this week and plan to review PRs sometime next week.

Samples/BlazorExample/BlazorExample/DataAccess/PersonDal.cs
```diff
@@ -1,74 +1,74 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
+﻿//using System;
```
rockfordlhotka: Shouldn't this file just be deleted/removed - no sense leaving it here commented out.

Samples/BlazorExample/BlazorExample/Server/Startup.cs
```diff
@@ -23,7 +24,14 @@ public void ConfigureServices(IServiceCollection services)
                   new[] { "application/octet-stream" });
       });
 
-      services.AddTransient(typeof(DataAccess.IPersonDal), typeof(DataAccess.PersonDal));
+      //services.AddTransient(typeof(DataAccess.IPersonDal), typeof(DataAccess.PersonDal));
```
rockfordlhotka: Is there a generic `AddTransient`? Like `AddTransient<DataAccess.IPersonDal, DataAccess.Mock.PersonDal>()`?

rockfordlhotka:
@SachinPNikam Nice changes, I like this a lot!

Just a couple comments - one about deleting a file instead of commenting out the contents, the other about `AddTransient` and whether we can make that code slightly nicer.

Can you respond to those? Thank you!

rockfordlhotka:
Also, you should probably put that table script into the project folder and add it to the repo so it is with the code.

Samples/BlazorExample/BlazorExample/DataAccess/PersonDal.cs
SachinPNikam: Sure I will correct this .

Samples/BlazorExample/BlazorExample/Server/Startup.cs
SachinPNikam: I will try

SachinPNikam:
as per your suggestions I did  PR for  Person table scripts in sample project, deleted old PersonDal .also now blazor sample default conneting to the mock .only  generic AddTransient related remains .
Thank you !

Issue #1481, opened 2020-02-04
IServiceScope (in ApplicationContext.ServiceProviderScope) is disposed if a DataPortal method is called inside another DataPortal.


Eduardo-Micromust:
Csla.Server.DataPortal.Complete disposes scope without checking if it is being used.

```
  [TestClass]
  public class DisposeScopeTest
  {
    [TestMethod]
    public void Test_Scope_Dispose()
    {
      var serviceCollection = new ServiceCollection();
      serviceCollection.AddScoped<DisposableClass>();

      ApplicationContext.DefaultServiceProvider = serviceCollection.BuildServiceProvider();

      _ = ClassA.GetClassA();
    }

  }

  public class DisposableClass
  : IDisposable
  {
    public bool IsDisposed { get; private set; } = false;
    public void Dispose()
    {
      IsDisposed = true;
    }
  }

  public class ClassA : BusinessBase<ClassA>
  {
    public static ClassA GetClassA()
    {
      return Csla.DataPortal.Fetch<ClassA>();
    }

    [Fetch]
    private void Fetch([Inject]DisposableClass disposable)
    {
      if (disposable.IsDisposed)
      {
        throw new ObjectDisposedException(nameof(disposable)); //No Exception here
      }

      _ = ClassB.GetClassB();
      if (disposable.IsDisposed)
      {
        throw new ObjectDisposedException(nameof(disposable)); //ObjectDisposedException is thrown
      }
    }
  }

  public class ClassB : BusinessBase<ClassB>
  {
    public static ClassB GetClassB()
    {
      return Csla.DataPortal.Fetch<ClassB>();
    }

    [Fetch]
    private void Fetch([Inject]DisposableClass disposable)
    {
    }
  }
```

Eduardo-Micromust:
I couldn't find where the IServiceScope is disposed in Csla.DataPortalClient.DataPortalProxyFactory.Create and Csla.Server.ObjectFactoryLoader.GetFactory. Is it ok to keep the IServiceScope in memory?

rockfordlhotka:
I don't know.

I suspect the reason you'd need to dispose the scope is so it can dispose any singleton or other instances it is referencing? Not like the scope itself holds any expensive resources (like a file handle, etc).

However, I do think that the data portal needs to dispose the scope before returning the call to the local client (so at the end of the logical server-side call).

Eduardo-Micromust:
> I don't know.
> 
[...]

PR #1484, opened 2020-02-05
#1481 Count if scope reference is zero before disposing

Eduardo-Micromust:
#1481 Created a Scope Reference Count to check if the IServiceScope can be disposed.


Source/Csla.Shared/ApplicationContext.cs
```diff
@@ -878,8 +878,20 @@ internal static IServiceProvider ScopedServiceProvider
           return null;
       }
     }
+
+    private static int _scopeRef;
```
rockfordlhotka: I don't think this can be `static`, because this is logical (and maybe physical) server-side code. That means many data portal requests may be running at the same time, each with their own scope.

Eduardo-Micromust:
I think there is a better fix to the code: Csla.Server.DataPortal.ClearContext is always executed at the end of the DataPortal flow. When the value of "_oldValue" is equals to ApplicationContext.LogicalExecutionLocations.Client, means that we are about to leave the server logical execution and must dispose the scope.

Source/Csla.Shared/ApplicationContext.cs
Eduardo-Micromust: I published a fix that use Csla.Server.DataPortal._oldLocation and do not create any variables.

rockfordlhotka:
@Eduardo-Micromust I am not sure this will work on a physical app server, because I don't think the old value will be "client" in that case. Can you confirm?

Source/Csla.Shared/Server/DataPortal.cs
```diff
@@ -752,6 +740,18 @@ private void SetContext(DataPortalContext context)
 
     private void ClearContext(DataPortalContext context)
     {
+#if !NET40 && !NET45
+      if (_oldLocation == ApplicationContext.LogicalExecutionLocations.Client)
```
rockfordlhotka: I don't think this is a valid assumption when the logical server-side code is running on a physical app server.

Eduardo-Micromust: Csla.ApplicationContext.LogicalExecutionLocation returns LogicalExecutionLocations.Client when the value is not present in the ContextDictionary:
```
    public static LogicalExecutionLocations LogicalExecutionLocation
    {
      get
      {
        object location = LocalContext.GetValueOrNull("__logicalExecutionLocation");
        if (location != null)
          return (LogicalExecutionLocations)location;
        else
          return LogicalExecutionLocations.Client;
      }
    }
```
 It happens when the first DataPortal call reach Csla.Server.DataPortal.SetContext and this value is stored in "_oldLocation":

    ```
private void SetContext(DataPortalContext context)
    {
      _oldLocation = Csla.ApplicationContext.LogicalExecutionLocation;
      ApplicationContext.SetLogicalExecutionLocation(ApplicationContext.LogicalExecutionLocations.Server);

      if (!context.IsRemotePortal && ApplicationContext.WebContextManager != null && !ApplicationContext.WebContextManager.IsValid)
        ApplicationContext.SetContext(context.ClientContext, context.GlobalContext);

```

This way we can use _oldLocation to know if the call is about to leave the logical server. Am I correct?

rockfordlhotka: Hmm. That is an unintentional behavior. Not that I plan to change it, but the behavior is a coincidence, not part of any specific design.

Now it becomes a pre-requisite for some functionality, and so stops being unintentional - does that make sense?

That's ok, but requires a unit test to confirm that this is the behavior (that the value starts out as "client"), so if it ever does change in the future we know that is bad.

Eduardo-Micromust: Totally makes sense. It is unintentional but excellent behavior: it created an easy way to check if the code is about to leave the server. 
I created tests to check the initial value of ApplicationContext.LogicalExecutionLocation.

rockfordlhotka: Thank you @Eduardo-Micromust !

Issue #1509, opened 2020-02-24
Add a For clause to CslaValidationMessages in Blazor


rockfordlhotka:
The current `CslaValidationMessages` implementation in `Csla.Blazor` has a `PropertyName` property, which works in the basic cases.

It would be ideal it it had a `For` property as well.

1. Provide symmetry with the standard `ValidationMessages` component
1. Support more complex object graph scenarios

PR #1512, opened 2020-03-15
Addition of 'For' parameter to CslaValidationMessages component in Csla.Blazor

TheCakeMonster:
#1509 Added 'For' parameter to CslaValidationMessages component in Csla.Blazor.

This is a breaking change, because the component had to be made generic for the use of an expression to be possible. Unfortunately, that has forced the addition of an extra parameter to the component for use when using PropertyName. If anyone is already using the component with the PropertyName parameter, they will have to change their code to also include the PropertyType parameter as well.

PropertyType need not be specified when using the new For clause, as .NET Core can infer the type from the expression provided.

### DIFF
--- a/Source/Csla.Blazor.Test/EditContextCslaExtensionsTests.cs
+++ b/Source/Csla.Blazor.Test/EditContextCslaExtensionsTests.cs
@@ -113,6 +113,26 @@ namespace Csla.Blazor.Test
     }
 
     [TestMethod]
+    public void ValidateModel_EmptyChildEmailAddress_OneValidationMessage()
+    {
+      // Arrange
+      FakePerson person = GetValidFakePerson();
+      EditContext editContext = new EditContext(person);
+      editContext.AddCslaValidation();
+
+      // Create a new child object (which is immediately invalid)
+      person.EmailAddresses.NewEmailAddress();
+
+      // Act
+      editContext.Validate();
+      IEnumerable<string> messages = editContext.GetValidationMessages();
+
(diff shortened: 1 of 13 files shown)


Brannos1970:
Rocky, I am trying this out but I am curious what "PropertyType" is supposed to be supplied. For example, I have a property int AppointmentStatusId, do I enter "int" or "typeof(@vm.Model.AppointmentStatusId)".

TheCakeMonster:
@Brannos1970 int would suffice. typeof() would probably work too, and has the slight advantage that if you changed the type of the property then the update would happen automatically.

Adding PropertyType was an unfortunate, but necessary, side effect of adding the more capable parameter - For. If you specify an expression in the For parameter then you need not specify PropertyType, as it is inferred from the parameter. This happens because the component was changed to be a generic component. However, the downside to this change is that the type cannot be inferred when specifying a property name as a string, hence the need to have a PropertyType to accompany it.

Either specify PropertyName AND PropertyType, OR specify For.

The syntax for the For parameter is a little convoluted, which is why I didn't do it this way to begin with. However, it is much more capable, as it allows display of validation rules from any property on any object in an entire, complex object graph. PropertyName plus PropertyType only works on a property of the root object.

rockfordlhotka:
I'll be changing the book and samples to all use `For` because that matches the behavior of other standard Blazor controls.

Brannos1970:
I found the For parameter to be better. You may want to consider phasing out the PropertyType parameter.

Issue #1522, opened 2020-03-20
Csla.Data.EntityFrameworkCore=> GetManager returning disposed DbContext object


SachinPNikam:
**Describe the bug**
Csla.Data.EntityFrameworkCore 
if we create the DbContextManager object by calling GetManager(database,label ) first time it will work but second time if we call the same it is returning disposed DbContext object 

**Version and Platform**
CSLA version:5.1.0
OS: Windows

JasonBock:
I really, really, **really** think all the context classes and managers should just go away. Use `[Inject]` on your operation methods and go the DI/IoC route.

Chicagoan2016:
@JasonBock , we have used `ConnectionManager` everywhere in our Projects -- projects that use native ADO.NET. We haven't encountered any problems but should we be on the lookout?
Thanks

SachinPNikam:
Hi @JasonBock thanks for suggestions i will look it that, for now the above issue is due to key is used to assign application.LocalContext but same key is not used for removing the db context. that is fixed in PR.

rockfordlhotka:
Don't worry, Jason is musing. There's no current plan to remove those types.

JasonBock:
I am musing, and I'm musing very **hard** on this :). Static members have the smell of service locator, which is an anti-pattern in my book. (Yes, I know extension methods are static, but they act on a given instance of a class). Not having anything globally accessible is hard, but should be minimized at all costs. I'd much rather be in a place where I'm getting a DB connection injected to my object, rather than use some kind of static manager whenever possible.

SachinPNikam:
Thanks Rocky !!!

PR #1523, opened 2020-03-20
#1522 Fixed DeRef() method updated to remove disposed dbcontext from LocalContext

SachinPNikam:
#1522 Fixed DeRef() method updated to remove disposed dbcontext from LocalContext


rockfordlhotka:
Thank you @SachinPNikam - nice enhancement (or bug fix, depending on how you look at it)

Issue #1542, opened 2020-04-03
WPFUi should use appsettings.json for DataPortal URL


BlagoCuljak:
To get better jumpstart on CSLA using WPF Ui, I have pushed ProjectTracker where WPFUi is using appsettings.json.

Hope to do the same for Blazor Client, but maybe it's better to wait for final release.

rockfordlhotka:
@BlagoCuljak this new set of changes (https://github.com/MarimerLLC/csla/issues/1548) will impact your PR I think.

BlagoCuljak:
We can undo it, if we're about to have more issues.

Didn't look too much into the issue, I'm on mobile app.

rockfordlhotka:
@BlagoCuljak Do you want to revisit this now that the dust has settled on the other changes around DI and app initialization?

I do think having config values read correctly would be a great step forward!

BlagoCuljak:
Sure, what do you propose? That we scrap these changes in favor of #1548?

rockfordlhotka:
Merge your reading of the config file/settings into the new (current) app config code.

https://github.com/MarimerLLC/csla/blob/master/Samples/ProjectTracker/ProjectTracker.Ui.WPF/App.xaml.cs

BlagoCuljak:
Dude, you always get me on some holiday. 😀 First Easter, now Labour day. If you don't want to wait for next week, could you please do it yourself? Thanks

rockfordlhotka:
There's no rush, I was just wondering if you remained interested in working on this.

BlagoCuljak:
Rocky, could you do the merge, please? I'm having issues that prevent me from doing so, sorry.

rockfordlhotka:
You mean for the original PR?

https://github.com/MarimerLLC/csla/pull/1543

I think the problem is that PR changed files (mostly `Program.cs`) that have been radically altered since the PR was created. It may not be possible to merge at this point.

It is almost certainly easier to take the _current_ `Program.cs`, which uses the new `UsingCsla` method coming in 5.2.0, and add the code to read the config settings using the standard .NET config subsystem.

PR #1543, opened 2020-04-03
Enhance WPF ProjectTracker app configuration

BlagoCuljak:
My first CSLA Pull request
#1542

### DIFF
--- a/Samples/ProjectTracker/ProjectTracker.Ui.WPF/ConfigurationExtensions.cs
+++ b/Samples/ProjectTracker/ProjectTracker.Ui.WPF/ConfigurationExtensions.cs
@@ -1,0 +1,38 @@
+﻿
+//-----------------------------------------------------------------------
+// <copyright file="ConfigurationExtensions.cs" company="Marimer LLC">
+//     Copyright (c) Marimer LLC. All rights reserved.
+//     Website: https://cslanet.com
+// </copyright>
+// <summary>Implement extension methods for .NET Core configuration</summary>
+//-----------------------------------------------------------------------
+using System;
+using Csla.Configuration;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace WpfUI
+{
+  /// <summary>
+  /// Implement extension methods for .NET Core configuration
+  /// </summary>
+  public static class ConfigurationExtensions
+  {
+    /// <summary>
+    /// Add CSLA .NET services for use by the application.
+    /// </summary>
+    /// <param name="services">ServiceCollection object</param>
+
+
+    /// <summary>
+    /// Configure CSLA .NET settings from .NET Core configuration
+    /// subsystem.
+    /// </summary>
+    /// <param name="config">Configuration object</param>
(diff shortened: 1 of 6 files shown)


rockfordlhotka:
Thank you @BlagoCuljak 

It seems like the configuration extension method applies to all platforms right? Perhaps it should go into the `Csla.Shared` project in the `Csla.Configuration` namespace.

What do you think?

BlagoCuljak:
Sure, that can be in Shared. I will make another pull request, sorry.

BlagoCuljak:
@rockfordlhotka I trying to do the same for Blazor projects, but Client and Server are not referenced to ProjectTracker.BusinessLibrary.Shared. Can I reference them to use configuration extension? Is that ok?

rockfordlhotka:
Blazor projects work differently, in that the `AddCsla` and `UseCsla` methods are used in `Startup.cs` or `Program.cs` (in the new client-side approach).

fwiw, I've started using the newer approach for my WPF and WinForms apps too. It is a little extra work to set up a builder and service collection, but worth it to get DI and all the other new goodness.

I approve and appreciate your new method because not everyone will want to use the new goodness in .NET. But for my part, I plan to use the new approach for ProjectTracker as I update everything for the books (which takes months - writing books isn't fast!).

BlagoCuljak:
Sure man, glad I could help!

rockfordlhotka:
btw, you don't need to do _another_ PR - you can just update this one.

rockfordlhotka:
@BlagoCuljak Are you still planning to update this PR to add that extension method to `Csla.Configuration`?

BlagoCuljak:
I have a question, ProjectTracker uses the [ProjectTracker.BusinessLibrary.Netstandard](https://github.com/MarimerLLC/csla/tree/master/Samples/ProjectTracker/ProjectTracker.BusinessLibrary.Netstandard) for objects, least in my Visual Studio. The Shared folder is unloaded, and not referenced in WpfUI.

rockfordlhotka:
Shared projects are just containers for code. So the `ProjectTracker.BusinessLibrary.Netstandard` project references the shared project to get the code. The shared project itself doesn't get compiled, doesn't generate a DLL, nothing. That code is compiled by a concrete project like the `Netstandard` one.

So other projects reference the concrete project that actually does compile the code into an assembly.

rockfordlhotka:
@BlagoCuljak I would like to get a contributor agreement from you - can you sign and send me that document?

BlagoCuljak:
Sure, no problem. Do you need it for this task or in general?

I'm Central European Time, and on Easter Holiday by middle of next week.

BlagoCuljak:
I know how shared projects work, but my issue is when I add file to Shared Project, it doesnt show up in BusinessLibrary, nor I can use it from WPF. 

If I add file to BusinessLibrary, then I can use it in other projects.

rockfordlhotka:
I'm reviewing the ProjectTracker solution, and I see that I did move all the code from the shared project to the netstandard project, so the shared project is no longer needed. I'll remove it from the repo.

Issue #1657, opened 2020-06-03
DataPortalExceptionInspector=> InspectException does not track actual objectType,and businessObject in case exception is thrown by any CommandBase.


SachinPNikam:
**Describe the bug**
DataPortalExceptionInspector=> InspectException does not track actual objectType,and businessObject in case exception is thrown by any CommandBase BO.
only track objectType as DataPortalTarget, businessObject  DataPortalTarget 

**Version and Platform**
CSLA version: 5.2.0

PR #1659, opened 2020-06-03
#1657 InspectException 

SachinPNikam:
#1657 Fixed DataPortalExceptionInspector=> InspectException does not track actual objectType,and businessObject in case exception is thrown by any CommandBase.


SachinPNikam:
Hello I want to merge only SimpleDataPortal.cs 
Thanks !

PR #1660, opened 2020-06-03
#1657 InspectException

### LINKED ISSUE #1657: DataPortalExceptionInspector=> InspectException does not track actual objectType,and businessObject in case exception is thrown by any CommandBase.

SachinPNikam:
Fixed #1657


Source/Csla.Shared/Server/SimpleDataPortal.cs
```diff
@@ -203,8 +203,8 @@ private async Task<DataPortalResult> Execute(DataPortalTarget obj, DataPortalCon
         }
         throw DataPortal.NewDataPortalException(
               "DataPortal.Execute " + Resources.FailedOnServer,
-              new DataPortalExceptionHandler().InspectException(obj.GetType(), obj, null, "DataPortal.Execute", ex),
-              obj);
+              new DataPortalExceptionHandler().InspectException(obj.Instance.GetType(), obj.Instance, null, "DataPortal.Execute", ex),
```
rockfordlhotka, line 206: @SachinPNikam do we need to confirm that `obj.Instance` isn't null?

SachinPNikam: I tested locally it's not null ,it is actual instance of Command base BO which the exception.

PR #1662, opened 2020-06-04
#1657 fixed  InspectException

SachinPNikam:
fixed  InspectException with null check obj.Instance


SachinPNikam:
hello rocky null check for obj.Instance done.

Issue #1820, opened 2020-09-23
Blazor authentication requires authenticated user; should default to allow unauthenticated


rockfordlhotka:
As per this [thread and solution](https://github.com/MarimerLLC/csla/discussions/1819#discussioncomment-83788).

rockfordlhotka:
Fix contributed by @JacoJordaan

PR #1821, opened 2020-09-25
Blazor authentication requires authenticated user; should default to allow unauthenticated

JacoJordaan:
Blazor authentication requires authenticated user; should default to allow unauthenticated #1820


rockfordlhotka:
Thank you @JacoJordaan !!

PR #1882, opened 2020-11-05
Support compression in Http data portal channel

### LINKED ISSUE #1881: Add code to support compression in the Http data portal channel
The data portal Http channel (HttpProxy and HttpPortalController) should be enhanced to support compression.

These changes shouldn't have an impact unless compression is configured in the web server startup code.

As per https://github.com/MarimerLLC/csla/discussions/1694

BlagoCuljak:
Closes #1881

### DIFF
--- a/Samples/ProjectTracker/ProjectTracker.AppServerCore/Startup.cs
+++ b/Samples/ProjectTracker/ProjectTracker.AppServerCore/Startup.cs
@@ -6,6 +6,9 @@ using Csla.Configuration;
 using Microsoft.Extensions.Hosting;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
 using ProjectTracker.Configuration;
+//using Microsoft.AspNetCore.ResponseCompression;
+//using System.IO.Compression;
+//using System.Linq;
 
 namespace ProjectTracker.AppServerCore
 {
@@ -22,6 +25,20 @@ namespace ProjectTracker.AppServerCore
     // This method gets called by the runtime. Use this method to add services to the container.
     public void ConfigureServices(IServiceCollection services)
     {
+      //services.AddResponseCompression(options =>
+
+      //{
+
+      //  options.Providers.Add<GzipCompressionProvider>();
+      //  options.Providers.Add<BrotliCompressionProvider>();
+      //  options.MimeTypes =
+      //      ResponseCompressionDefaults.MimeTypes.Concat(
+      //          new[] { "application/*", "text/*" });
+      //});
+      //services.Configure<GzipCompressionProviderOptions>(options =>
+      //{
+      //  options.Level = CompressionLevel.Fastest;
+      //});
       services.AddCors(options =>
       {
         options.AddPolicy(BlazorClientPolicy,
@@ -57,6 +74,7 @@ namespace ProjectTracker.AppServerCore
     // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
     public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
     {
+      //app.UseResponseCompression();
       if (env.IsDevelopment())
       {
         app.UseDeveloperExceptionPage();
--- a/Samples/ProjectTracker/ProjectTracker.AppServerCore/Startup.cs
+++ b/Samples/ProjectTracker/ProjectTracker.AppServerCore/Startup.cs
@@ -6,9 +6,9 @@ using Csla.Configuration;
 using Microsoft.Extensions.Hosting;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
 using ProjectTracker.Configuration;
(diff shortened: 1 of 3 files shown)


Samples/ProjectTracker/ProjectTracker.AppServerCore/Startup.cs
```diff
@@ -6,6 +6,9 @@
 using Microsoft.Extensions.Hosting;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
 using ProjectTracker.Configuration;
+//using Microsoft.AspNetCore.ResponseCompression;
```
rockfordlhotka, line 9: I think I'd rather just turn on compression in ProjectTracker rather than leaving these as comments. Might as well directly show folks the "right way" to set this up?

Source/Csla.Web.Mvc.Shared/Server/Hosts/HttpPortalController.cs
```diff
@@ -173,6 +173,8 @@ private async Task InvokePortal(string operation, Stream requestStream, Stream r
       var serializer = new MobileFormatter();
       var result = new HttpResponse();
       HttpErrorInfo errorData = null;
+      Response.Headers.Add("Content-type", "application/octet-stream,text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9");
```
rockfordlhotka, line 176: Does this apply if the use-text-serialization flag is set for Blazor? In that case the byte stream MUST BE just text, because of limitations of webassembly.

rockfordlhotka (changes requested):

rockfordlhotka:
Thank you for the PR @BlagoCuljak! I have done:

1. Updated the PR title
1. Updated the PR description to indicate it is linked to the issue and should close it
1. Started a review with a couple questions/comments - please provide your thoughts or updates

It is great to have you on the contributor team!

Samples/ProjectTracker/ProjectTracker.AppServerCore/Startup.cs
BlagoCuljak: If you're about to do that, then installing Microsoft.AspNetCore.ResponseCompression is a must. I didn't include it in PR.

Source/Csla.Web.Mvc.Shared/Server/Hosts/HttpPortalController.cs
BlagoCuljak: As far I have seen Blazor on requests adds all of these types.

And if our header response is not matched, that data will not be displayed.

I remember when I only had headers with Content type application-octet-stream, with no compression, Blazor did not work at all.

I will double check webasembly in the morning, to match all the headers.

Samples/ProjectTracker/ProjectTracker.AppServerCore/Startup.cs
rockfordlhotka: That needs to be a dependency added to the app server? Also to the various client projects?

Source/Csla.Web.Mvc.Shared/Server/Hosts/HttpPortalController.cs
rockfordlhotka: Thank you. It is Blazor WebAssembly that I'm concerned about.

Samples/ProjectTracker/ProjectTracker.AppServerCore/Startup.cs
BlagoCuljak: No, just add NuGet Microsoft.AspNetCore.ResponseCompression to AppServerCore.

I will do another PR today.

Source/Csla.Web.Mvc.Shared/Server/Hosts/HttpPortalController.cs
BlagoCuljak: Hi Rocky, I have tested WebAssembly and it should work, it accepts all headers, and don't see any reason why it should not work.
client


However, making prerelease 5.4 would make my testing whole lot simpler.  I don't think we will break anyone's installation, and we can make the proper ProjectTracker/CSLA adjustments for production version.

Samples/ProjectTracker/ProjectTracker.AppServerCore/Startup.cs
rockfordlhotka: I see where you uncommented the code, but not where the nuget package was added to the appservercore csproj file.

BlagoCuljak: Well, I guess you are right, the package was not needed. I remember adding it to my project (not the project tracker).

This seems fine as is.

PR #1926, opened 2020-11-30
Use binary/compression only if UseTextSerialization is false

### LINKED ISSUE #1925: Http compression throws an exception in Blazor client code (dotnet 5)
When attempting to invoke the data portal via the `HttpProxy` a `PlatformNotSupportedException` is thrown when interacting with the compression API calls added as part of #1881 

It is necessary to figure out which calls are causing this exception and ensure they aren't invoked when the `HttpProxy` is running in the wasm client context.

rockfordlhotka:
Closes #1925 Use binary/compression only if UseTextSerialization is false

### DIFF
--- a/NuGet/Definition/WPF.NuSpec
+++ b/NuGet/Definition/WPF.NuSpec
@@ -37,7 +37,7 @@ Supports the creation of WPF applications.
         <dependency id="Microsoft.Extensions.Hosting" version="3.1.9" />
         <dependency id="Microsoft.Extensions.Hosting.Abstractions" version="3.1.9" />
       </group>
-      <group targetFramework="net5.0-windows10.0.19041">
+      <group targetFramework="net5.0-windows10.0.19041.0">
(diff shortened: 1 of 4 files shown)


rockfordlhotka:
@BlagoCuljak I also think the list of items for the binary headers is too extensive. Surely they don't all apply to a proprietary binary steam?

```c#
        Response.Headers.Add("Content-type", "application/octet-stream,text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9");
```

The `application/octet-stream` makes sense, and that's about it. The data isn't signed by exchange, isn't any sort of image, and isn't XML, etc.

BlagoCuljak:
Hi Rocky, the headers that are in my code are all from Blazor Server Side, and I remember that I was able to see them when requests were made.

Certainly, you can try the same with your code, and if dataportal request/response is being implemented, then it's ok.

PR #1951, opened 2020-12-01
Consolidate Core and Data.EF projects into a single multi-targeted one

dotMorten:
This consolidates the core and Data.EF projects into a single project, and avoiding all the extra shared projects and platform projects.
You can package the project directly into a nuget with right-click pack (no nuspec needed now!).

You might want to consider consolidating the different nuget package versions for each target, just to simplify it a little, but I tried keeping it to the versions that were already there.

There are more projects you can do this to, but let's start with this one.

Before:
image

After:
image


rockfordlhotka:
Thank you @dotMorten - this is really nice work! (though I've just skimmed it, and need to review further to really understand)

dotMorten:
You're welcome. Let me know if you have any questions. Main difference is really just the new `csla.csproj`. Other than that, it's deleting the old now unused projects, moving the files in the `csla.shared` folder, and updating project references.
You might have to update your nuget build script and get rid of that old nuspec as you simply get the nuget package but using the `/t:Pack` target

PR #2005, opened 2020-12-24
Update CslaPermissionsPolicyProvider.cs

### LINKED ISSUE #2000: CSLA breaks other authorization policies
**Describe the bug**
If I add a policy in `Startup.ConfigureServices` with `services.AddCsla().WithBlazorServerSupport()` I will get an error `The AuthorizationPolicy named: 'TokenAuthentication' was not found.` when calling an MVC action which references my policy.

**Version and Platform**
CSLA version: 5.4.0

**Code that Fails**

Add a policy:
```            
services.AddAuthorization(options => options.AddPolicy(Constants.TokenAuthentication,
                          policy => policy.AddAuthenticationSchemes(Constants.TokenAuthentication).RequireAuthenticatedUser()));
```

and an attribute to an MVC controller:

`  [Authorize(Policy = Constants.TokenAuthentication)]`

`GetPolicyAsync` seems to ignore my custom policy and returns the fallback policy instead

```
    public Task<AuthorizationPolicy> GetPolicyAsync(string policyName)
    {
      CslaPermissionRequirement requirement;
      if (!CslaPolicy.TryGetPermissionRequirement(policyName, out requirement))
        return this.GetFallbackPolicyAsync(); 
      AuthorizationPolicyBuilder authorizationPolicyBuilder = new AuthorizationPolicyBuilder(Array.Empty<string>());
      authorizationPolicyBuilder.AddRequirements((IAuthorizationRequirement) requirement);
      return Task.FromResult<AuthorizationPolicy>(authorizationPolicyBuilder.Build());
    }
```

Remove this line and my policy works:

```
//services.AddCsla().WithBlazorServerSupport();
```

System.InvalidOperationException: The AuthorizationPolicy named: 'TokenAuthentication' was not found.

j055:
Use `AuthorizationOptions` from existing implementation and cache policies

Fixes #2000

### DIFF
--- a/Source/Csla.Blazor/CslaPermissionsPolicyProvider.cs
+++ b/Source/Csla.Blazor/CslaPermissionsPolicyProvider.cs
@@ -1,10 +1,12 @@
-﻿//-----------------------------------------------------------------------
+//-----------------------------------------------------------------------
 // <copyright file="CslaPermissionsPolicyProvider.cs" company="Marimer LLC">
 //     Copyright (c) Marimer LLC. All rights reserved.
 //     Website: https://cslanet.com
 // </copyright>
 // <summary>CSLA permissions policy provider</summary>
 //-----------------------------------------------------------------------
+
+using System;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.Extensions.Options;
@@ -16,10 +18,7 @@ namespace Csla.Blazor
   /// </summary>
   public class CslaPermissionsPolicyProvider : IAuthorizationPolicyProvider
   {
-    /// <summary>
-    /// Gets the fallback policy provider
-    /// </summary>
-    public DefaultAuthorizationPolicyProvider FallbackPolicyProvider { get; }
+    private readonly AuthorizationOptions _options;
 
     /// <summary>
     /// Creates an instance of the type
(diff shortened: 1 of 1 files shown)


Source/Csla.Blazor/CslaPermissionsPolicyProvider.cs
```diff
@@ -16,30 +18,32 @@ namespace Csla.Blazor
...
     /// <summary>
     /// Creates an instance of the type
     /// </summary>
     /// <param name="options">Authorization options</param>
     public CslaPermissionsPolicyProvider(IOptions<AuthorizationOptions> options)
     {
-      FallbackPolicyProvider = new DefaultAuthorizationPolicyProvider(options);
+      if (options == null)
+      {
+        throw new ArgumentNullException(nameof(options));
```
rockfordlhotka, line 31: Do we want an exception? Or to get the fallback from a `DefaultAuthorizationPolicyProvider`?

j055: `DefaultAuthorizationPolicyProvider` is already added as part of the ASPNET Host Defaults. I'm not sure that it's valid for options to be null. 

Incidentally when do CslaPolicy's get used? I haven't see any policy names starting with 'Csla:' even though I have `AuthorizationRules`

rockfordlhotka: That makes sense about the options.

Can you ask about the CslaPolicy stuff in a new Discussion? This isn't a great location for that conversation.

PR #2019, opened 2021-01-05
Find methods in base classes

### LINKED ISSUE #1994: ServiceProviderMethodCaller cannot resolve factory methods in base class.
Hello Rocky,

OK, I have to bring up the `ObjectFactory` failing with base classes in V5 again. You're right in the method for finding object factory methods hasn't changed, however now you're calling `ServiceProviderMethodCaller.FindDataPortalMethod` in `DataPortalT` before the factory method. That method is failing to find factory methods in an object factory base class. After some searching and reading issue #1789, I've confirmed the call to `factoryType.GetMethods(_factoryBindingAttr)` only returns the declared level methods because of the `BindingAttribute`. The original `MethodCaller.FindMethod` will iterate through the base classes trying to find the method.

`ServiceProviderMethodCaller.FindDataPortalMethod` iterates through business class bases instead. So I'm thinking the `_factoryBindingAttr` field could be changed so all methods for all levels of a factory class can be found?

rockfordlhotka:
Closes #1994

### DIFF
--- a/Source/Csla.Shared/Reflection/ServiceProviderMethodCaller.cs
+++ b/Source/Csla.Shared/Reflection/ServiceProviderMethodCaller.cs
@@ -22,8 +22,8 @@ namespace Csla.Reflection
   /// </summary>
   public static class ServiceProviderMethodCaller
   {
-    private static readonly BindingFlags _bindingAttr = BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.DeclaredOnly;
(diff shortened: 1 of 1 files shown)


rockfordlhotka:
@kellyethridge can you review this PR?

I think removing the `DeclaredOnly` limitation fixes the problem - at least it does in my test.

```c#
    private static readonly BindingFlags _bindingAttr = BindingFlags.Instance | BindingFlags.NonPublic;
    private static readonly BindingFlags _factoryBindingAttr = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic;
```

kellyethridge:
The call to `ServiceProviderMethodInfo FindDataPortalMethod` is indeed finding methods in an object factory's base class. So I'll call that a success :)

Issue #1738, opened 2020-07-27
Require initialization of Dependency Injection for use of CSLA .NET


rockfordlhotka:
The .NET Framework 4.8 and .NET Core (and .NET 5) support the same standard .NET dependency injection subsystem. 

CSLA has used this since CSLA version 3.0, but some of the code in CSLA is rather ugly because we support the DI _and pre-DI_ implementations of various concepts. Starting with version 6 we'll be requiring that a DI service provider be initialized to use CSLA .NET, allowing us to assume it exists within the CSLA implementation.

This is no different from how modern ASP.NET requires initialization of DI, and increasingly people are initializing DI in other hosts such as WPF, WinUI, Xamarin, Blazor, etc.

Using DI to create types means that private constructors for domain business classes is no longer supported. This hasn't been supported when using `MobileFormatter` for many years, and now it is not supported in any scenario.

JasonBock:
I can't say I agree or disagree because I don't exactly get what you mean by this. Does this mean a developer will always have to call `AddCsla()` now when an application "starts up", even if there are no dependencies involved?

rockfordlhotka:
Yes, exactly. That allows the code inside CSLA to stop using `Activator.CreateInstance` in favor of the new DI-based activator that does injection.

It also means we can do things like use DI within CSLA itself to inject things like config or other values throughout the framework.

JasonBock:
Part of me wonders if there's an analyzer in play, but I don't know if it's possible to write an analyzer that says, "hey, you never called `AddCsla()` in your code."

ShaiCohen26:
I _think_ @rockfordlhotka means that if you do not initialize Dependency Injection, CSLA will throw a _runtime_ exception. 

> Starting with version 6 we'll be requiring that a DI service provider be initialized to use CSLA .NET, allowing us to **_assume_** it exists within the CSLA implementation. (emphasis added)

A poor analogy of the new behavior is connection strings. Your codebase assumes the referenced connection string exists; however if the connection string is missing, a runtime exception is thrown.

rockfordlhotka:
Yes, that is exactly what will happen.

It will probably happen early, because I expect the data portal to use DI to configure itself, so the first attempt to call the data portal will result in runtime failure.

rockfordlhotka:
This discussion is valuable to consider when implementing this change: #1983

rockfordlhotka:
In this latest commit I have changed my approach a little.

[...]

PR #2028, opened 2021-01-10
Fall back to default provider if scoped is not available

rockfordlhotka:
#1738 Fall back to default provider if scoped is not available


dazinator (approved):
Looks fine to me. Noted you didn't use null coalescence for terseness and assume that's a style preference, i.e
`var provider = ApplicationContext.ScopedServiceProvider ?? ApplicationContext.DefaultServiceProvider`

PR #2085, opened 2021-02-03
Add extensibility to HttpProxy; add HttpCompressionProxy type

### LINKED ISSUE #2075: HttpProxy compression code doesn't work with custom HttpClient
The compression code added in #1881 causes a problem when using a custom `HttpClient` object.

> This new code broke our code at runtime with exception "Specified argument was out of range of valid values. Parameter name: Unhandled CSLA Known type was found"
> 
> This is because we override GetHttpClient() to create an own HttpClient to pass along the client's AccessToken.
@rockfordlhotka Perhaps add a boolean property to add the compression headers or not?

@rfcdejong I don't understand the "fix" in your comment in #1881, can you explain more completely?

Also, I wonder if there is a better solution for compression (and the token stuff) by using DI in CSLA 6? My thought there is to require that the `HttpClient` instance be injected, so any customization of the client would occur during app initialization.

rockfordlhotka:
Closes #2075 Add extensibility to HttpProxy; add HttpCompressionProxy type


BlagoCuljak (approved):
Seems fine.

PR #2166, opened 2021-03-17
[v5.x] - Expose PropertyName on Csla.Blazor.IPropertyInfo interface.

### LINKED ISSUE #2165: Expose PropertyName on Csla.Blazor.IPropertyInfo interface.
Currently unable to add a unique `id=""` attribute to HTML elements as `PropertyName` on the `Csla.Blazor.IPropertyInfo` interface is not exposed.

adrianwright109:
Fixes #2165

### DIFF
--- a/Source/Csla.Blazor/PropertyInfo.cs
+++ b/Source/Csla.Blazor/PropertyInfo.cs
@@ -21,10 +21,6 @@ namespace Csla.Blazor
     /// Gets the model
     /// </summary>
     protected object Model { get; }
-    /// <summary>
-    /// Gets the property name
-    /// </summary>
-    protected string PropertyName { get; }
 
     /// <summary>
     /// Creates an instance of the type.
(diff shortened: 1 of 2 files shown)


Source/Csla.Blazor/PropertyInfo.cs
```diff
@@ -21,10 +21,6 @@ public class PropertyInfo : IPropertyInfo, INotifyPropertyChanged
     /// Gets the model
     /// </summary>
     protected object Model { get; }
-    /// <summary>
-    /// Gets the property name
-    /// </summary>
-    protected string PropertyName { get; }
```
rockfordlhotka, line 27: There's no implementation here to actually get and return the property name, so how would this work?

I'd expect an implementation similar to the one in the `FriendlyName` property.

rockfordlhotka (changes requested):
I think the property needs an implementation.

Source/Csla.Blazor/PropertyInfo.cs
adrianwright109: Isn't it getting set in the `PropertyInfo` constructor?
https://github.com/MarimerLLC/csla/blob/5ab18303d4a0c9ae218edfcbb6e64e15dd72b568/Source/Csla.Blazor/PropertyInfo.cs#L33

rockfordlhotka: Oh, my bad!

rockfordlhotka (approved):

PR #2176, opened 2021-03-21
Remove ALL protected virtual DataPortal_XYZ methods

### LINKED ISSUE #2164: Remove DataPortal_XYZ methods from base classes
With the new data portal operation attributes (such as `Fetch` and `Insert`) the old DataPortal_XYZ methods are no longer required, and just complicate code in and out of the framework.

Remove all `protected virtual` DataPortal_XYZ operation methods.

adrianwright109:
Fixes #2164

### DIFF
--- a/Source/Csla.Validation.Test/Project.cs
+++ b/Source/Csla.Validation.Test/Project.cs
@@ -141,7 +141,8 @@ namespace Csla.Validation.Test
 
     #endregion
 
-    protected override void DataPortal_Create()
+    [Create]
+    protected void DataPortal_Create()
     {
       
     }
--- a/Source/Csla.test/AppContext/ExceptionRoot.cs
+++ b/Source/Csla.test/AppContext/ExceptionRoot.cs
@@ -83,14 +83,18 @@ namespace Csla.Test.AppContext
       Csla.ApplicationContext.GlobalContext["create"] = "create";
       throw new ApplicationException("Fail create");
     }
-    protected override void DataPortal_Insert()
+
+    [Insert]
+    protected void DataPortal_Insert()
     {
       //we would insert here
       Csla.ApplicationContext.GlobalContext["Root"] = "Inserted";
       Csla.ApplicationContext.GlobalContext["create"] = "create";
       throw new ApplicationException("Fail insert");
     }
-    protected override void DataPortal_Update()
+
+    [Update]
+    protected void DataPortal_Update()
     {
       Csla.ApplicationContext.GlobalContext["Root"] = "Updated";
       Csla.ApplicationContext.GlobalContext["create"] = "create";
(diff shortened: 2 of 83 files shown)


adrianwright109:
@rockfordlhotka - This PR is a WIP hence why it is in draft mode.

So far I have removed ALL `protected virtual void DataPortal_XYZ` methods (e.g. `Create`, `Fetch`, `Update`, `Insert`, `Delete`, `DeleteSelf` and `Execute`) from the various `Base` classes.

I'm not sure if this is enough or whether I need to do more so could do with some guidance?

I noticed that the `DataPortal_XYZ` methods are also used in these other places:

Should these another references I've found also be removed?

1. https://github.com/MarimerLLC/csla/blob/07597492c507e38c1d227202c498eadab5a7dffd/Source/Csla.Analyzers/Csla.Analyzers/CslaMemberConstants.cs#L25-L31

2. https://github.com/MarimerLLC/csla/blob/07597492c507e38c1d227202c498eadab5a7dffd/Source/Csla/Server/DataPortalTarget.cs#L235-L243

3. There are also some other `protected virtual void DataPortal_*` methods that I have left for example:
```
protected virtual void DataPortal_OnDataPortalInvoke(DataPortalEventArgs e)
protected virtual void DataPortal_OnDataPortalInvokeComplete(DataPortalEventArgs e)
protected virtual void DataPortal_OnDataPortalException(DataPortalEventArgs e, Exception ex)
```

4. Also the corresponding `private void DataPortal_XYZ` methods, I'm assuming these will be used by any method annotated with one if the following attributes: `[Create]`, `[Fetch]` etc.

rockfordlhotka:
@adrianwright109 thank you for the great work so far, sorry I've been slow in responding 😳

## 1. DP_XYZ constants (analyzers)

@JasonBock - I'm guessing we should leave the constants for the "DP_XYZ" method names for the analyzers, at least for now, because they are what allow the analyzers to find and fix missing data portal operation attributes - iirc?

## 2. DP_XYZ constants (data portal target)

@adrianwright109 I think that these can probably go away, because we're now relying on the use of data portal operation attributes to find all the methods. I suspect though, that this will require some work in terms of how methods are discovered in a factory class - probably simplifying the code by always using the dynamic method discovery instead of looking for specific method names.

## 3. DP_Invoke, etc.

@adrianwright109 I think those should remain for now. Removing them means adding some more operation attributes and code to find/invoke those methods. Which might be a good idea, but should be in its own backlog item.

## 4. Private DP_XYZ methods

I'm not sure what you are referring to here? Are there private DP_XYZ methods in the base classes of the framework?

Oh, yeah, probably in Csla.Security for some identity types, etc?

I'd expect those need to remain intact, because they are actually invoked by the data portal at runtime.

JasonBock:
Yes, the analyzers use those to find methods with that name and then flag them as issues. So, please don't remove those :).

adrianwright109:
Closing in favour of #2208

Issue #2095, opened 2021-02-08
DataPortal operation on repeatedly loaded assembly (using AssemblyLoadContext) causing cast exception - Can not cast [A]MyClass from MyAssembly.dll to [B]MyClass from MyAssembly.dll


BaHXeLiSiHg:
**Describe the bug**
Probably i'm doing something wrong, but will try to explain :\

Have main application with installed CSLA library. On request - loading dynamic assembly using AssemblyLoadContext and calling DataPortal.Fetch method from class within this dynamically loaded assembly.

On first call everything works fine, but after some time of inactivity flushing this not used anymore dynamic assembly, calling AssemblyLoadContext.Unload(), GC.Collect() and GC.WaitForPendingFinalizers().

After period of time trying to access same dynamic assembly via same scenario: loading in AssemblyLoadContext and calling DataPortal.Fetch, but this time getting error:

`
DataPortal.Create failed ([A]MyClass cannot be cast to [B]MyClass. Type A originates from 'MyAssembly.dll, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' in a byte array. Type B originates from 'MyAssembly.dll, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' in a byte array.)
`

That is call stack:

`
   at lambda_method(Closure , Object , Object[] )
   at Csla.Reflection.ServiceProviderMethodCaller.CallMethodTryAsync(Object obj, ServiceProviderMethodInfo method, Object[] parameters)
   at Csla.Reflection.ServiceProviderMethodCaller.CallMethodTryAsync(Object obj, ServiceProviderMethodInfo method, Object[] parameters)
   at Csla.Reflection.LateBoundObject.CallMethodTryAsyncDI[T](Boolean isSync, Object[] parameters)
   at Csla.Server.DataPortalTarget.InvokeOperationAsync[T](Object criteria, Boolean isSync)
   at Csla.Server.DataPortalTarget.FetchAsync(Object criteria, Boolean isSync)
   at Csla.Server.SimpleDataPortal.Fetch(Type objectType, Object criteria, DataPortalContext context, Boolean isSync)
   at Csla.DataPortal`1.Fetch(Type objectType, Object criteria)
   at Csla.DataPortal`1.Fetch(Object[] criteria)
   at Csla.DataPortal.Fetch[T](Object[] criteria)
   at MyClass.Fetch(MyFetchCriteria criteria)
`

[...]

PR #2179, opened 2021-03-24
Added support for loading/unloading of assemblies which are using CSLA ("AssemblyLoadContext", .Net5+, removing of obsolete cached types / re-caching).

BaHXeLiSiHg:
#2095

### DIFF
--- a/Source/Csla.Shared/ALC/ALCManager.cs
+++ b/Source/Csla.Shared/ALC/ALCManager.cs
@@ -1,0 +1,112 @@
+﻿#if NET5_0_OR_GREATER
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Runtime.Loader;
+
+namespace Csla.ALC
+{
+    /// <summary>
+    /// Provides common access to "AssemblyLoadContext" logic, which is used for "Plugins" concept of .NET 5+
+    /// </summary>
+    public static class ALCManager
+    {
+        /// <summary>
+        /// Simplified helper for "unloadable" cache item creation, linked to currently active "ContextualReflectionScope".
+        /// </summary>
+        /// 
+        /// <param name="objectType">Caching item source.</param>
+        /// <param name="cachingItem">Caching item.</param>
+        /// <param name="unloadAction">Action which will be performed after "AssemblyLoadContext" unload.</param>
+        /// <param name="excludeNonCollectible">
+        /// Provides possibility to not cache items which are referenced to main application - "AssemblyLoadContext.Default".
+        /// </param>
+        /// 
+        /// <typeparam name="TValue">Type of caching item.</typeparam>
+        /// 
+        /// <returns>Tuple structure where "Item1" is name of active "AssemblyLoadContext" and "Item2" - caching item.</returns>
+        /// 
+        /// <exception cref="ArgumentNullException">Throws if nor object type, nor caching item, not unload action was provided.</exception>
+        public static Tuple<string, TValue> CreateCacheInstance<TValue>(
+            Type objectType,
+            TValue cachingItem,
+            Action<AssemblyLoadContext> unloadAction,
+            bool excludeNonCollectible = false)
+        {
+            if (objectType == null)
+            {
+                throw new ArgumentNullException("objectType");
+            }
+
+            if (cachingItem == null)
+            {
+              throw new ArgumentNullException("cachingItem");
+            }
+
+            if (unloadAction == null)
+            {
+              throw new ArgumentNullException(nameof(Action<AssemblyLoadContext>));
+            }
+
+            string assemblyLoadContextName = null;
+
+            if ((!excludeNonCollectible || objectType.Assembly.IsCollectible)
+                && AssemblyLoadContext.CurrentContextualReflectionContext != null
+                && AssemblyLoadContext.CurrentContextualReflectionContext.Name != AssemblyLoadContext.Default.Name)
+            {
+                assemblyLoadContextName = AssemblyLoadContext.CurrentContextualReflectionContext.Name;
+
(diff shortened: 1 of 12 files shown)


rockfordlhotka:
Thank you for the PR @BaHXeLiSiHg.

Regarding timing, I want to communicate three things:

1. We use semver and this is a feature change, so it will go into version 5.5, not 5.4.x
2. This is a potentially risky change, which will need review, testing, a beta period, etc. - historically it is this sort of change that causes unforeseen edge case issues
3. Technically v5 is in maintenance mode. I'm making an exception for this by creating v5.5.0 for it, but really all new work is being done in v6; that means this needs to be implemented in v6 as well

The next release of anything will be 5.4.2, probably in the next 2-3 weeks. At which point all maintenance work will go into v5.5, along with this change. _That_ release will be in 2-3 months - again, giving plenty of time for folks to use a preview release so we don't get surprised by any threading/caching issues related to the changes you've made.

It is also the case that the first release of v6 will be in 3-4 months, and that will be the end of any work on v5.x.

Source/Csla.Shared/ALC/ALCManager.cs
```diff
@@ -0,0 +1,112 @@
+﻿#if NET5_0_OR_GREATER
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Runtime.Loader;
+
+namespace Csla.ALC
```
rockfordlhotka, line 8: I believe this should be in the Csla.Reflection namespace? I realize it is wrapping a `System.Runtime` concept, but it _feels_ more reflection-like than anything?

Also, by convention I think it should be `Alc` instead of `ALC`.

Source/Csla.Shared/Reflection/ServiceProviderMethodCaller.cs
```diff
@@ -214,7 +238,12 @@ public static ServiceProviderMethodInfo FindDataPortalMethod<T>(Type targetType,
               }
               else
               {
+#if NET5_0_OR_GREATER
```
rockfordlhotka, line 241: Isn't this #if block invoking the same line of code either way?

Source/Csla.Shared/Reflection/ServiceProviderMethodCaller.cs
```diff
@@ -236,7 +265,12 @@ public static ServiceProviderMethodInfo FindDataPortalMethod<T>(Type targetType,
             throw new TargetParameterCountException(cacheKey);
           else
           {
+#if NET5_0_OR_GREATER
```
rockfordlhotka, line 268: Isn't this #if block invoking the same line of code either way?

Source/Csla.Shared/ALC/ALCManager.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    /// <summary>
+    /// Provides common access to "AssemblyLoadContext" logic, which is used for "Plugins" concept of .NET 5+
+    /// </summary>
+    public static class ALCManager
```
rockfordlhotka, line 13: Type should be named `AlcManager`

rockfordlhotka (changes requested):

Source/Csla.Shared/Reflection/ServiceProviderMethodCaller.cs
BaHXeLiSiHg: Yep, value getter places have null-checks. Missed on refactoring.

For previous comment, resolved it without comment:

I have no complains) You can name it whatever you like. I'm not sure that it has something with reflection, active ```AssemblyLoadContext``` just provides "closest" context for types caching/searching. Contexts are linked together via "child" -> "parent" chain, up to main, "Default", application context (like legacy ```AppDomain```). So if type was registred in "Default" application context, but was accessed in some "dynamic" context, it will be looked up to "Default" context on this contexts chain, something like that.

Source/Csla.Shared/ALC/ALCManager.cs
BaHXeLiSiHg: I have no complains) You can name it whatever you like. I'm not sure that it has something with reflection, active AssemblyLoadContext just provides "closest" context for types caching/searching. Contexts are linked together via "child" -> "parent" chain, up to main, "Default", application context (like legacy AppDomain). So if type was registred in "Default" application context, but was accessed in some "dynamic" context, it will be looked up to "Default" context on this contexts chain, something like that.

Source/Csla.Shared/Reflection/ServiceProviderMethodCaller.cs
BaHXeLiSiHg: Same here as in previous place.

Source/Csla.Shared/ALC/ALCManager.cs
BaHXeLiSiHg: If you want so, named like this because "AssemblyLoadContext" - three words, but "Alc" seems like shortcut of one word, like "Alcohol" :D

Source/Csla.Shared/ALC/ALCManager.cs
rockfordlhotka: We could use a `Csla.Runtime` namespace to match the Microsoft convention, I'm fine with that as well.

Source/Csla.Shared/ALC/ALCManager.cs
rockfordlhotka: Perhaps `AssemblyLoadContextManager` is the best then - avoid confusing acronyms and make the code more readable overall.

Especially if you like my thought about a `Csla.Runtime` namespace, it would help be a little self-documenting.

Issue #2384, opened 2021-07-20
Load CSLA v5.4.2 API docs into cslanet.com website


rockfordlhotka:
For the most part the codebase includes decent API docs in comments. In the past I've used this (with ndoc and later sandcastle) to create chm or other help files.

These days what would be really nice is to use sandcastle or some other tool to generate static html content that we could add into the github repo under the `/docs` directory - maybe `/docs/api` or something?

This [SO thread](https://stackoverflow.com/questions/15316381/create-html-documentation-for-c-sharp-code) has some tooling options.

Brannos1970:
I have begun the documentation of the API.

Brannos1970:
An initial generation and preview of the documentation is uploaded and can be viewed at https://brannos1970.github.io/csla/5.4.2/html/. I am still working on this and would welcome any comments and suggestions to make it better.

rockfordlhotka:
I like the look of this. Some thoughts

* As it generates actual html, it probably can't go into the /docs directory, and instead should go into https://cslanet.com - maybe in a /docs directory there?
* I would like to _not_ see the various unit test projects in the API docs, as I think they just add confusion
* Probably we should not see the analyzers either?

Brannos1970:
I just released an updated preview of the API documentation. I removed the various unit test projects and the analyzers. The updated API is available at https://brannos1970.github.io/csla/5.4.2/html/. 

We can place the documentation anywhere. Right now, I have it set up to run from GitHub Pages which can be viewed at https://brannos1970.github.io/csla/. 

I am using Doxygen to generator the API documentation. If everyone is okay with that I will continue to do so as I work on the documentation.

rockfordlhotka:
I think this is a lot cleaner, thank you!

I see from the doxygen site that they've customized other info in the nav pane. It would be nice to keep that minimal, but also to at least add links to https://cslanet.com and the github repo - maybe also a specific one for the Discussions tab?

I think it is perfect if you continue to work in your fork until it is ready to roll out.

That does also give time to think about where it'll end up once we do roll it out.

Versioning is the challenge I see. Something like https://cslanet.com/docs would be nice - but inadequate, since usually there are at least 2 "active" versions at any time - the released/bugfix version, and whatever is in `main`. So maybe what you are doing with the version in the URL is the answer. That does imply (in my mind) the need for an intermediate landing page though?

* cslanet.com/docs/5.4.2
* cslanet.com/docs/5.5.0
[...]

PR #2415, opened 2021-08-08
Create .nojekyll file

Brannos1970:
API documentation linked pages were not working. Had to create a .nojekyll empty file to fix the issue.

This is for API documentation #2384.


rockfordlhotka:
Is that file in the root, or in the 5.4.2 directory? It looks like it might be in the root, which would prevent creation of index.html.

Brannos1970:
It is the root. When I tested it the index.html was still created.

rockfordlhotka:
@Brannos1970 looks good, yea!

PR #2446, opened 2021-08-23
Fix undo issue for lazy loaded property values

### LINKED ISSUE #2444: LazyLoad property and BeginEdit
### Discussed in https://github.com/MarimerLLC/csla/discussions/2395

<sup>Originally posted by **amselem** July 30, 2021</sup>
Hi,
I have a bo with some lazyload properties shown in a windows form app. 
Since the lazyload properties are heavy to load , I only load them if the user clicks on their respective tabpage. The user can click in the tabpage before or after beginning an edit in this form.
When the user begins the form edit I call my bo.BeginEdit()  so I can call bo.CancelEdit() if the user wants to rollback his changes.
However the Lazyload fields are rolled back only if they were loaded previously (i.e. if the user clicked in the tabpage). If the user begins an edit, then clicks on the tabpage, the lazyload property is not rolled back on an undo operation.  

What's the best/correct approach to solve this inconsistent behavior?

Kind regards

</div>

It seems that there are possible scenarios here.

1. In an undo the previous value was null - which your suggested change addresses by retaining the new value - _except_ I can imagine a scenario where it really _was_ null and should return to null - maybe your change needs to only happen if the property is marked for LazyLoad?
2. In an undo the previous value was _a different instance_ - so we should revert to that previous instance? That's maybe unrealistic, since it will always _technically_ be a different instance due to deserialization. Hmm...
3. In an undo the previous value was an instance and the _new value_ is null - so we should revert to that previous instance

This backlog issue is to **fix scenario number 1**.

rockfordlhotka:
Fixes #2444 Fix undo issue for lazy loaded property values

### DIFF
--- a/Source/Csla.Shared/Core/FieldManager/FieldDataManager.cs
+++ b/Source/Csla.Shared/Core/FieldManager/FieldDataManager.cs
@@ -564,6 +564,9 @@ namespace Csla.Core.FieldManager
               continue;
             }
           }
+          // restore IFieldData object into field collection
+          _fieldData[index] = oldItem;
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


rockfordlhotka:
#2444 Move lazy load check to correct location

Issue #2392, opened 2021-07-29
Csla should better handle/report on attempts to use synchronous data portal access from WebAssembly


TheCakeMonster:
** Problem Description **
I wrote a Data Annotations attribute that required data from an ASP.NET Core hosted backend. The attribute failed with a slightly difficult to find problem, with the exception message 'An exception occurred during a WebClient request'. Because of the structure of the code, I didn't know what part of my system was generating the exception, and it took quite a long time to track down the problem to my synchronous data access. I didn't know that WebClient was used for synchronous access under the covers, so I didn't even think to look to Csla as the source of the exception; I thought it must have been something else, because the data portal was working for everything else (using async data access methods.)

**The recommended solution**
I think Csla should generate a more user-friendly error message, by identifying before the call to WebClient that the operation will not work, and explaining that what the user is trying to do is not supported. This will hopefully make it clearer to developers that the error of their ways is to try synchronous data access.

**Alternatives I have considered**
I did wonder whether Csla should just 'do the right thing' by using a blocking call using an async method of HttpClient in this scenario - sync over async. However, this would further complicate the code and hide the underlying issue, potentially leading to undesirable side effects that come from sync over async (thread starvation, deadlocking, performance issues etc.) It's not really 'the right thing' because of the potential for side effects of this nature.

**Additional context**
Sometimes doing asynchronous data access is hard, forced on us by being compatible with patterns that were created before async/await existed. Ideally, we would support synchronous data access. However, we shouldn't take on the problems that providing this on both .NET FX and .NET (Core) since they behave differently.

TheCakeMonster:
Oh, right. We could just "do the right thing" even if we wanted to, as an attempt to call sync over async on WebAssembly generates the exception "Cannot wait on monitors on this runtime"

_sigh_

TheCakeMonster:
Using  System.Runtime.InteropServices.RuntimeInformation.OSDescription looks to be the best way to determine if the application is running on WASM. The string that is returned does vary a bit by version; in .NET 3.1 it returns "web" and in .NET 6 it returns "Browser". However, since this is only triggering the generation of an exception to help the developer, it is probably acceptable that this is a string-based comparison.

TheCakeMonster:
System.Runtime.InteropServices.RuntimeInformation.OSArchitecture would have been the most reliable option to determine the platform, if only it supported WASM under .NET 3.1. Sadly, WASM reports as Architecture.X86 under .NET 3.1

[...]

PR #2479, opened 2021-09-08
Check for whether synchronous WebClient calls are supported, Issue #2392

TheCakeMonster:
Added a check for whether synchronous WebClient calls are supported, to throw a more helpful exception when running on WebAssembly, Issue #2392

I chose not to complicate things with dependency injection, as this class is not testable anyway due to the level of implementation detail it already contains.


rockfordlhotka:
@TheCakeMonster 

This looks like a nice improvement for Blazor wasm, thank you.

I added some minor comments/requests for changes to the PR. Please let me know if you have any questions.

TheCakeMonster:
@rockfordlhotka Thanks for the review. I've incorporated all of the recommended changes.

PR #2487, opened 2021-09-14
Flow all context values to logical server in all scenarios

### LINKED ISSUE #2485: ApplicationContext.User is lost inside async DP methods (MVC app)
**Describe the bug**
`ApplicationContext.User` is not available inside DP **async** `Fetch` method after being set in an MVC controller action.

It is available in a non-async method call.

**Version and Platform**
CSLA version: 5.5.0
OS: Windows
Platform: ASP.NET Core, MVC, NETFX 4.8

**Code that Fails**

```
 public class HomeController : Controller
    {
        public async Task<ActionResult> Index()
        {
            var principal = new ClaimsPrincipal(new ClaimsIdentity(
                new[] { new Claim(ClaimTypes.Name, "Special"), new Claim(ClaimTypes.Role, "Destroyer") }));
            
            ApplicationContext.User = principal;
            var obj = await DataPortal.FetchAsync<PersonEdit>("abc");

            return View();
        }
    }

    [Serializable]
    public class PersonEdit : BusinessBase<PersonEdit>
    {
        public static readonly PropertyInfo<string> NameProperty = RegisterProperty<string>(nameof(Name));
        public string Name
        {
            get => GetProperty(NameProperty);
            set => SetProperty(NameProperty, value);
        }

        public static readonly PropertyInfo<string> ContextProperty = RegisterProperty<string>(nameof(Context));
        public string Context
        {
            get => GetProperty(ContextProperty);
            set => SetProperty(ContextProperty, value);
        }

        public static readonly PropertyInfo<AnotherRoot> AnotherProperty = RegisterProperty<AnotherRoot>(nameof(AnotherRoot));
        public AnotherRoot AnotherRoot
        {
            get => GetProperty(AnotherProperty);
            private set => LoadProperty(AnotherProperty, value);
        }

        //[Fetch]
        //private void Fetch(string name)
        //{
        //    var isInRole = ApplicationContext.User.IsInRole("Destroyer");
        //    Debug.WriteLine(isInRole); // TRUE
        //}

        [Fetch]
        private async Task FetchAsync(string name)
        {
            var isInRole = ApplicationContext.User.IsInRole("Destroyer");

            Debug.WriteLine(ApplicationContext.User.Identity.Name);
            Debug.WriteLine(isInRole); // FALSE
        }
    }
```

rockfordlhotka:
Fixes #2485

### DIFF
--- a/Source/Csla.Shared/DataPortalT.cs
+++ b/Source/Csla.Shared/DataPortalT.cs
@@ -29,47 +29,6 @@ namespace Csla
     /// </summary>
     public Csla.Core.ContextDictionary GlobalContext { get; set; }
 
-    private class DataPortalAsyncRequest
-    {
-      public object Argument { get; set; }
-      public System.Security.Principal.IPrincipal Principal { get; set; }
-      public Csla.Core.ContextDictionary ClientContext { get; set; }
-      public Csla.Core.ContextDictionary GlobalContext { get; set; }
-      public object UserState { get; set; }
-      // passes CurrentCulture and CurrentUICulture to the async thread
-      public CultureInfo CurrentCulture;
-      public CultureInfo CurrentUICulture;
-
-      public DataPortalAsyncRequest(object argument, object userState)
-      {
-        this.Argument = argument;
-        this.Principal = Csla.ApplicationContext.User;
-        this.ClientContext = Csla.ApplicationContext.ClientContext;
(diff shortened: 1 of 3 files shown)


j055 (approved):
Failing code https://github.com/MarimerLLC/csla/issues/2485 now passes.

Thank you

Issue #2148, opened 2021-03-11
Enable serialization of POCO types via MobileFormatter


rockfordlhotka:
There are times when it would be nice to include a simple POCO style type in an object graph. Right now the most basic serializable type is `MobileObject`, which requires manual work to serialize things like fields.

There are several thoughts to consider. One is documented here, which is to use a base class. Another would be to have MF trigger on an attribute that delegates to an arbitrary "serializer" that returns/accepts a `byte[]` representing that class. Yet another would be to pass in a list of serializers to MobileFormatter (via DI), where each of these serializers would have the opportunity to say that they can serialize a given type that would otherwise not be serialized by MF.

### MobilePoco base class

It is possible to create a `MobilePoco` type that uses reflection to get/set property and field values as needed by the MobileFormatter. This should be added to the Csla.Core namespace.

Here's a quick prototype of the idea:

```c#
using System;
using Csla.Serialization.Mobile;

namespace ConsoleApp3
{
  [Serializable]
  public abstract class MobilePoco : IMobileObject
  {
    void IMobileObject.GetChildren(SerializationInfo info, MobileFormatter formatter)
    {
    }

    void IMobileObject.SetChildren(SerializationInfo info, MobileFormatter formatter)
    {
    }

    private const System.Reflection.BindingFlags bindingFlags =
      System.Reflection.BindingFlags.Public |
      System.Reflection.BindingFlags.NonPublic |
      System.Reflection.BindingFlags.Instance;

    void IMobileObject.GetState(SerializationInfo info)
    {
      var fields = this.GetType().GetFields(bindingFlags);
      foreach (var item in fields)
        info.AddValue(item.Name, item.GetValue(this));
      var properties = this.GetType().GetProperties(bindingFlags);
      foreach (var item in properties)
        info.AddValue(item.Name, item.GetValue(this));
    }


    void IMobileObject.SetState(SerializationInfo info)
    {
      var fields = this.GetType().GetFields(bindingFlags);
      foreach (var item in fields)
        item.SetValue(this, info.GetValue<object>(item.Name));
      var properties = this.GetType().GetProperties(bindingFlags);
      foreach (var item in properties)
        item.SetValue(this, info.GetValue<object>(item.Name));
    }
  }
}
```

### SerializeUsing attribute

```c#
[SerializeUsing(typeof(MySerializer))]
public class MyPoco {}
```

Presumably the `MySerializer` type would implement `Csla.Serialization.ISerializationFormatter`.

[...]

PR #2499, opened 2021-09-23
Addition of Source Generator for auto serialization of POCOs

TheCakeMonster:
Addition of source generator to enable auto serialization of POCOs, for example where a list of items needs to be passed across the data portal as a parameter of a command object.

Now that source generators are possible, using a source generator seemed a better solution than a base class with reflection, as this avoids the expense of reflection on runtime performance. A source generator offers metadata programming with the only expense being incurred at build time.

For security, the behaviour is to serialize only public fields and properties by default. This limits the risk of vulnerabilities from insecure deserialization, as developers generally tend not to expect non-public fields to be manipulated outside of their control. Should a developer require the inclusion of other fields or properties, they can be included by adding the [AutoSerialized] attribute to them. The opposite, the [AutoNonSerialized] attribute, exists to allow any public field or property to be excluded from serialization, such as event handlers.

Reference for non-public fields and properties being a common target for abuse can be seen in the .NET section of the following article:
[https://klezvirus.github.io/Advanced-Web-Hacking/Serialisation/](https://klezvirus.github.io/Advanced-Web-Hacking/Serialisation/) (use "Find In Page" using the term ".NET: Binary Archive Format"; first example)

To serialize a type, mark it as partial and apply the [AutoSerializable] attribute to it. Add the [AutoSerialized] attribute to any non-public fields or properties which you wish to be included in serialization, and add the [AutoNonSerialized] attribute to any public fields or properties that are to be excluded from serialization.

Resolves issue #2148


TheCakeMonster:
I've added some scripts related to packaging and an update to Core.NuSpec, but I'm not confident that the packaging will be correct. It's not something I have any experience of

Issue #2544, opened 2021-10-19
Update unit tests for .NET 6


rockfordlhotka:
Now that the CSLA 6 codebase builds, it is time to bring the unit tests back into the `csla.test.sln` project.

Part of #2120

PR #2602, opened 2021-11-18
Refactoring of .NET Core tests for CSLA 6

TheCakeMonster:
Fixes to tests for CSLA 6 on .NET Core 3.1

Forms part of unit test fixes detailed in issue #2544

I don't think the tests in DataPortalClientTests.cs are relevant anymore as a result of the DataPortalProxyFactory class having been retired; they are commented out for review. I think we've also done away with the ability to use BinaryFormatter, so there is a test in SerializationTests.cs that is also commented out, pending review.


Source/csla.netcore.test/Configuration/FluentConfigTests.cs
```diff
@@ -22,159 +22,159 @@
...
-    public void FluentConfigSecurity()
-    {
-      new CslaConfiguration()
-        .Security().PrincipalCacheMaxCacheSize(123);
-
-      Assert.AreEqual(123, Csla.Security.PrincipalCache.MaxCacheSize, "MaxCacheSize");
-    }
-  }
+  //  [TestClass]
+  //  public class FluentConfigTests
```
rockfordlhotka, line 26: I agree with removing this code.

Source/csla.netcore.test/DataPortal/DataPortalClientTests.cs
```diff
@@ -27,86 +27,87 @@ public class DataPortalClientTests
...
-        { ProxyTypeName = typeof(Csla.DataPortalClient.HttpProxy).AssemblyQualifiedName, DataPortalUrl = "https://example.com/test" });
-      Csla.DataPortalClient.DataPortalProxyFactory.AddDescriptor("abc", fake);
-
-      var factory = new Csla.DataPortalClient.DataPortalProxyFactory();
-      var proxy = factory.Create(typeof(ResourceType));
-      Assert.IsNotNull(proxy, "proxy can't be null");
-      Assert.IsInstanceOfType(proxy, typeof(Csla.DataPortalClient.HttpProxy), "should be httpproxy");
-      Assert.AreEqual("https://example.com/test", ((Csla.DataPortalClient.HttpProxy)proxy).DataPortalUrl);
-    }
+    // I don't think these tests are relevant anymore, since the changes for CSLA 6 do away with DataPortalProxyFactory
```
rockfordlhotka, line 35: Agreed

Source/csla.netcore.test/Serialization/SerializationTests.cs
```diff
@@ -49,32 +52,37 @@ public void UseMobileFormatter()
...
-        Assert.AreEqual(ApplicationContext.SerializationFormatters.BinaryFormatter, serializer);
-        var s = Csla.Serialization.SerializationFormatterFactory.GetFormatter();
-        Assert.IsInstanceOfType(s, typeof(Csla.Serialization.BinaryFormatterWrapper));
-      }
-      finally
-      {
-        Csla.Configuration.ConfigurationManager.AppSettings["CslaSerializationFormatter"] = "MobileFormatter";
-      }
-    }
+    // BinaryFormatter dropped as of CSLA 6?
```
rockfordlhotka, line 55: Correct, no more BinaryFormatter, so related tests can go away

rockfordlhotka:
The test project(s) need to be added to the `csla.test.sln` solution to run as part of the CI build.

rockfordlhotka (approved):

TheCakeMonster:
There are a couple of outstanding issues here:

1. The configuration tests are currently disabled - tests in ReadConfigFromDotnetCoreSubsystemTests.cs. Because they affect global state, they break all of the other tests, which can run in parallel with them. We could try to solve this using some form of locking, but that would add constant complexity to the test solution. However, moving them into another test project might be the best approach.
2. Two of the dashboard tests fail - those associated with counting of failures. This might indicate that there is a bug with the dashboard code in the area of exception handling within the dataportal, but I've not yet had a chance to investigate this. These tests are disabled in DashboardTests.cs
3. We've got a problem with the build. Some of the tests that use DI fail as soon as Csla.AspNetCore is built. I haven't managed to work out why, despite a day of trying. Deleting the netcoreapp3.1 folder from the Bin folder is the temporary workaround if this occurs - but that's not a solution!

PR #2607, opened 2021-11-23
Fallback to no param in UpdateChild

### LINKED ISSUE #2559: Implement no-param fallback for UpdateChild data portal operation
Like we did in #2396 for the delete operations, for backward compatibility reasons the UpdateChild operation should fall back to a zero parameter data portal operation method if no other matches are found.

Basically, in `ServiceProviderMethodCaller`, the code should look like this:
```c#
          if (matches.Count() == 0
             && (typeOfOperation == typeof(DeleteSelfAttribute) || typeOfOperation == typeof(DeleteSelfChildAttribute)
             || typeOfOperation == typeof(UpdateChildAttribute)))
```

As per discussion thread https://github.com/MarimerLLC/csla/discussions/2449

Appropriate unit tests are also necessary.

rockfordlhotka:
Closes #2559

### DIFF
--- a/Source/Csla/Reflection/ServiceProviderMethodCaller.cs
+++ b/Source/Csla/Reflection/ServiceProviderMethodCaller.cs
@@ -202,7 +202,8 @@ namespace Csla.Reflection
             }
           }
           if (matches.Count() == 0 && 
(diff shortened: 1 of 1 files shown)


dchuen:
It looks like this didn't get included in 5.5.2 nuget package.

Issue #2644, opened 2021-12-10
Add example of authorization rules to BlazorExample


TheCakeMonster:
The BlazorExample sample doesn't include, or work with, authorization rules. When I added an authorization rule to PersonEdit, the sample did not work, throwing exceptions on the Add Person screen. As this partial sample is something that people will use as a reference when creating their own applications, it seems beneficial to have the limitations of the sample, and the framework, resolved.

This is a mixture between a feature request - an improvement to the sample to enable a more advanced example of how CSLA 6 works, it is also a bugfix to resolve issues that I have found whilst working on something else.

There are several problems in this area of the code:
1. The sample makes use of a custom authentication state provider that does not set ApplicationContext.Principal, so any attempt to apply authorization rules to objects results in an exception on the Add Person page. As it happens, the exception is difficult to diagnose, as the ViewModel captures, and does not display, the resultant exception (but this is not a bug; it stops the application from crashing, and so it's a nice addition.)
2. The built-in authorization rules attempt to use an ApplicationContext captured at the time that the rule was created, rather than the context information that forms part of the Execute method's existing parameters.

Note that the issues are twofold, and it will take two separate updates to resolve this:
1. There is a small bug in the BlazorExample application itself, which can be resolved immediately, and directly
2. There is a bug in the framework. Whilst this can be fixed as part of the same PR, it does not take effect immediately, as the BlazorExample solution references a Nuget package. The package needs to be republished and the references updated before the application will work correctly - but I have made changes in a re-referenced version to confirm that the solutions result in a working application.

PR #2645, opened 2021-12-10
Addition of code that enables authorization rules to work in BlazorExample

TheCakeMonster:
This change includes the addition of authorization rules to the sample, plus the resolution of two bugs that prevent authorization from working in it. One of these bugs - a failure to impact ApplicationContext when logging in, affects just this sample application. However, the other - use of the wrong ApplicationContext instance on authorization rules (the instance captured [or not, as it happens] at the time of creation of the shared rule object instance, instead of the one provided during execution) - has wider impact than just this sample code, and is worthy of much more significant scrutiny.

Resolution for new issue #2644

### DIFF
--- a/Samples/BlazorExample/BlazorExample/Client/CurrentUserService.cs
+++ b/Samples/BlazorExample/BlazorExample/Client/CurrentUserService.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Security.Claims;
+using Csla;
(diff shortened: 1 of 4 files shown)


rockfordlhotka (approved):
Nice improvements to the code, thank you!

rockfordlhotka:
Closes #2644

Issue #2649, opened 2021-12-12
Server-side data portal needs option to run rules before insert/update/delete operations


rockfordlhotka:
The server-side data portal should provide an option to run business rules (`CheckRules`) on each domain object instance before invoking the insert/update/delete operation methods on root and child objects.

If, after calling `CheckRules`, the object's `IsValid` property is invalid, then a validation exception should be thrown so the data portal operation is stopped and any database changes rolled back.

Another way to implement this would be to walk the object graph prior to invoking the _root_ insert/update/delete operation. This is probably more efficient than checking each object prior to lower-level operations, because checking at the top level would prevent _all_ attempts at updating the database, so no rollback would be required.

There is a question as to whether this behavior should be the default (a breaking change) or an option people can enable.

Either way, this should _not_ happen by default if the server-side data portal is run via `LocalProxy`, because in this case the data portal is already running in the context of the client, so re-checking all the rules would be a waste of resources.

rockfordlhotka:
This is based on an email thread with @TheCakeMonster

rockfordlhotka:
@TheCakeMonster I am sure I didn't do everything in the issue in my PR, but I made some good progress replicating what you'd done (sorry!).

PR #2651, opened 2021-12-13
Server-side object revalidation using extension to the existing IInterceptDataPortal seam

TheCakeMonster:
Reworked the existing extensibility point based on the IInterceptDataPortal interface to support use of multiple interceptors. The new implementation also makes use of DI for instantiation of the interceptors, so it is more compatible with other parts of CSLA 6.

The submission also includes a revalidating interceptor as the first built-in, example implementation.

There is a small breaking change here. While the existing, public IInterceptDataPortal interface remains unchanged, the way it is configured by anybody who is using it will need to change as they upgrade to CSLA 6. However, the mechanism to allow them to add their own interceptors is complete, and the change is very straightfoward and in keeping with the new configuration mechanism in this version of the framework.

At the moment, the RevalidatingInterceptor class attempts revalidation on all of the DataPortal operations. This could be restricted to just the Update operation; however, the implementation supports users who might choose to use an editable object as a parameter for one of the other operations, should that be useful - for example, to gain the ability to apply complex validation rules to the parameters if they are to be edited in a user interface prior to submission.

The new interceptors framework could be improved by changing the interface, or creating a new one, to support the return of a value from the interception methods. Such a return value could be used to bypass the remainder of the DataPortal operation if deemed necessary by an interceptor. At the moment there is no return from the Initialize or Complete methods, so the only way to short-circuit a DataPortal operation is to throw an exception, which is generally very slow in .NET.

Intended to resolve Issue #2649


rockfordlhotka (changes requested):
To avoid use of the old appsettings concept, I think the configuration for interceptors should look like this (in `Program.cs` or similar):

```c#
        .DataPortal()
          .AddServerSideDataPortal(o => o
            .InterceptorProviders.Add(typeof(Csla.Server.RevalidationInterceptor))));
```

TheCakeMonster:
> To avoid use of the old appsettings concept, I think the configuration for interceptors should look like this (in `Program.cs` or similar):
> 
> ```cs
>       services.AddCsla(options => options
>         .DataPortal()
>           .AddServerSideDataPortal(o => o
>             .InterceptorProviders.Add(typeof(Csla.Server.RevalidationInterceptor))));
> ```

Sorry, I didn't realise that we were trying to do away with the older structure entirely, so that's fair comment.

The code didn't start off the way I submitted it. Originally I added a new property to CslaDataPortalConfiguration for this value, but then I found this didn't work - the setting was lost. The static DataPortal() extension method returns a new instance of CslaDataPortalConfiguration on each call, and as we already have more than one call to that method in the codebase, and ended up with 3 the way I coded it, I got in a pickle. As a result, I looked for how this had been solved elsewhere, and happened upon the use of AppSettings.

The way we return a new instance of CslaDataPortalConfiguration on each call caused a bug in my code. I think it is likely that such bugs will arise again, as this is easy to overlook. I would suggest we change this.

The change I propose is to invert the dependency between CslaDataPortalConfiguration and CslaConfiguration/ICslaConfiguration, and have CslaConfiguration expose a read-only property of type CslaDataPortalConfiguration that the DataPortal() method can return. That way, in all of the extension methods, we can be confident that we are always working with the same instance of the configuration object, and therefore can store data within it, which keeps any consuming code simpler.

This proposal is not perfect, in that it couples CslaConfiguration/ICslaConfiguration to CslaDataPortalConfiguration, which slightly breaks one of the intentions of using extension methods. However, the change is very minor - everything else remains extension methods - and keeps new code simpler in the future.

I wonder if the suffix provider in InterceptorProviders would add any value? In general, my perception of a provider is that it is a single, replaceable entity, and this new mechanism supports the use of multiple interceptors, so it doesn't feel lexically correct anyway.

I suggest the methods to add and remove interceptors be generic. Generic methods are more in keeping with the syntax used for DI, so they read a little better, as well as avoiding the need for the typeof operator. Importantly they can also have constraints applied - especially that the type implements the required interface. This gives users compile-time checking that the types they are registering are compatible with the extensibility point they are using, rather than a runtime exception.

One final thought is that using an action as part of the AddServerSideDataPortal method to provide the configuration might mentally mislead users into thinking that interceptors only run on the server in a remote data portal configuration, which isn't strictly true. Whilst the only interceptors we have written are restricted to server-side execution, there is no reason why someone shouldn't make use of the extensibility point in the local data portal. Furthermore, the AddServerSideDataPortal method is called internally within the AddCsla method, so the order of execution of the code could result in confusing bugs. This is why I kept it separate.

This is a lot of text. I think the best approach is for me to submit the change I am proposing for further review. I think it achieves what you are looking for, but let's see.

TheCakeMonster:
I wonder if the ICslaConfiguration interface is doing us any favours. In my case, it stopped me doing what I intended.

I wanted to add an internal property onto CslaConfiguration to expose the CslaDataPortalConfiguration instance, but because we are using an interface, I had to expose it publicly instead.

rockfordlhotka:
I think ICslaConfiguration is _not_ helping us, and so I'm removing it - good point.

Source/Csla/Configuration/Fluent/DataPortalConfigurationExtensions.cs
```diff
@@ -58,7 +58,11 @@ public static DataPortalClientOptions AddServerSideDataPortal(this DataPortalCli
       options?.Invoke(opt);
       var services = builder.Services;
       services.TryAddScoped(typeof(IAuthorizeDataPortal), opt.AuthorizerProviderType);
-      services.TryAddScoped((p) => opt.InterceptorProviders);
+      foreach (Type interceptorType in opt.InterceptorProviders)
```
rockfordlhotka, line 61: Won't this throw an exception when more than one interceptor is provided? I don't think you can register more than one entry for a type. This is why I registered the list of providers, so the interceptormanager could unpack the list.

TheCakeMonster: No, it doesn't. TryAdd is slightly a misnomer, and there is a special behaviour for multiple registrations.

If you use Add twice, the first registration sort of overwrites the first, and there is no exception. TryAdd checks if an existing registration exists and avoids registering if there is one, so as not to 'overwrite' the first. In my opinion it's a slightly confusing name, because it's not used in the sense that Try is used in the rest of the framework.

However, in reality, I don't think Add _actually_ overwrites a registration anyway. I think both registrations are stored, but if you ask for a single instance then you get the most recent. However, if you ask for an IEnumerable, then you get an instance of each one registered.

The code I have shown does work, but it works because InterceptorManager receives an IEnumerable<IInterceptDataPortal> on its constructor. It's a really nice feature of DI, but it seems to be one that is not all that well known. I only know about it because I chanced upon it whilst researching this very feature. I think I was lucky; if I'd got a different answer to a very slightly different Google search, I'd not have found it.

TheCakeMonster: I tried running the code you used, and I got back an empty enumerator. I think that is mainly because it isn't associated with the interface I am asking for.

TheCakeMonster: I suppose for completeness I should call out that the IEnumerable<T> approach in constructors works for zero or more registrations. If nothing is registered it returns an empty enumerator. It's a really nice, straightforward approach because of that behaviour - a simple way to add optional extensibility using a pipeline-style (or decorator-style) pattern.

PR #2657, opened 2021-12-16
Major updates to data portal configuration and IAuthorizeDataPortal

### LINKED ISSUE #2633: Remove the context/connection manager types from Csla.Data
These types are no longer required, because database contexts, connections, and transactions can be managed via dependency injection.

OR we could consider changing these types to be DI-friendly, and maybe have them still provide value within the context of dependency injection:

* Stop reference counting (they'd be a singleton)
* Have them work more like `IHttpClientFactory`, where they get injected and then provide a named instance of the manager type

### LINKED ISSUE #2650: Add IAuthorizeDataPortal implementation that verifies per-type authn rules
It has been possible, nearly forever, for people to implement an `IAuthorizeDataPortal` provider that checks any authorization rules before allowing the server-side data portal pipeline to really get started.

The default provider has always just allowed all requests. The default _should_ enforce the basic per-type authn rules for the root type. The context provided to the authorizer includes the type of the root object and the data portal operation being requested, so it should be possible to invoke the `HasPermission` method on `Csla.Rules.BusinessRules` to check the authn rule for the type-operation combination.

The `IAuthorizeDataPortal` provider should _also_ be configured via DI as a service that is injected into the server-side data portal.

So this work item is as follows:

1. Add a provider that checks the type-operation permissions for all inbound requests
2. Remove the existing provider loading configuration based on app.config, etc.
3. Have the server-side `DataPortal` class require the authorizer service in its constructor
4. Enhance the `AddCsla` method to default to using the new authorizer
5. Do not remove the null authorizer, as people can use that for backward compatibility if desired

Here's an untested prototype of the authorizer:

```c#
using Csla.Security;

namespace Csla.Server
{
  public class ActiveAuthorizer : IAuthorizeDataPortal
  {
    public ActiveAuthorizer(ApplicationContext applicationContext)
    {
      ApplicationContext = applicationContext;
    }

    private ApplicationContext ApplicationContext { get; }

    public void Authorize(AuthorizeRequest clientRequest)
    {
      if (clientRequest.Operation == Csla.DataPortalOperations.Create)
      {
        if (!Csla.Rules.BusinessRules.HasPermission(
          ApplicationContext, Csla.Rules.AuthorizationActions.CreateObject, clientRequest.ObjectType))
          throw new SecurityException($"{clientRequest.ObjectType.Name}.{clientRequest.Operation}");
        return;
      }
      if (clientRequest.Operation == Csla.DataPortalOperations.Fetch)
      {
        if (!Csla.Rules.BusinessRules.HasPermission(
          ApplicationContext, Csla.Rules.AuthorizationActions.GetObject, clientRequest.ObjectType))
          throw new SecurityException($"{clientRequest.ObjectType.Name}.{clientRequest.Operation}");
        return;
      }
      if (clientRequest.Operation == Csla.DataPortalOperations.Update)
      {
        if (!Csla.Rules.BusinessRules.HasPermission(
          ApplicationContext, Csla.Rules.AuthorizationActions.EditObject, clientRequest.ObjectType))
          throw new SecurityException($"{clientRequest.ObjectType.Name}.{clientRequest.Operation}");
        return;
      }
      if (clientRequest.Operation == Csla.DataPortalOperations.Delete)
      {
        if (!Csla.Rules.BusinessRules.HasPermission(
          ApplicationContext, Csla.Rules.AuthorizationActions.DeleteObject, clientRequest.ObjectType))
          throw new SecurityException($"{clientRequest.ObjectType.Name}.{clientRequest.Operation}");
        return;
      }
      if (clientRequest.Operation == Csla.DataPortalOperations.Execute)
      {
        if (!Csla.Rules.BusinessRules.HasPermission(
          ApplicationContext, Csla.Rules.AuthorizationActions.EditObject, clientRequest.ObjectType))
          throw new SecurityException($"{clientRequest.ObjectType.Name}.{clientRequest.Operation}");
        return;
      }
    }
  }
}
```

The configuration for the server-side data portal with authorizer should look like this:

```c#
        .DataPortal()
          .AddServerSideDataPortal(o => o
            .AuthorizationProvider(typeof(Csla.Server.NullAuthorizer))));
```

rockfordlhotka:
Closes #2650
Closes #2633
Affects #2649


Source/Csla/Configuration/Fluent/CslaDataPortalConfiguration.cs
```diff
@@ -85,10 +125,53 @@ public CslaDataPortalConfiguration AutoCloneOnUpdate(bool value)
...
+    /// Gets a list of the IInterceptDataPortal instances
+    /// that should be executed by the server-side data portal.
+    /// injection.
+    /// </summary>
+    public List<IInterceptDataPortal> InterceptorProviders { get; } = new List<IInterceptDataPortal>();
+
+    /// <summary>
+    /// Gets or sets the type of the ExceptionInspector.
+    /// </summary>
+    public Type ExceptionInspectorType { get; set; }
```
TheCakeMonster: Instead of a public property of type Type, I wonder if we should use a generic method, perhaps RegisterExceptionInspector<T>(). Generic methods have the advantage that they can have constraints applied. This results in compile-time checks that the type implements the required interface, which avoids runtime exceptions caused by configuration using a type that fails to meet the required contract.

This applies to the public ExceptionInspectorType, ActivatorType and ObjectFactoryLoaderType properties.

Source/Csla/Server/ActiveAuthorizer.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// <summary>
+    /// Checks authorization rules for the request.
+    /// </summary>
+    /// <param name="clientRequest">
+    /// Client request information.
+    /// </param>
+    public void Authorize(AuthorizeRequest clientRequest)
+    {
+      if (clientRequest.Operation == DataPortalOperations.Create)
```
TheCakeMonster, line 38: The code in the AuthorizeRequest class which does this job is more succinct. All we have to decide is if it less readable or less secure because of its brevity?

I was surprised to see the code in AuthorizeRequest; I would expect this to be a POCO, but it contains functionality. I'm wondering if there is any benefit to the shorter code in AuthorizeRequest; if there is then maybe we should use that here - and then maybe delete it from AuthorizeRequest

Source/Csla/Configuration/Fluent/CslaDataPortalConfiguration.cs
```diff
@@ -31,13 +33,36 @@ public static CslaDataPortalConfiguration DataPortal(this ICslaConfiguration con
...
-      services.TryAddSingleton(typeof(Server.Dashboard.IDashboard), typeof(Csla.Server.Dashboard.NullDashboard));
+      services.AddScoped((p) => opt);
+      services.AddScoped(typeof(IAuthorizeDataPortal), opt.AuthorizerProviderType);
+      services.AddScoped((p) => opt.InterceptorProviders);
+      if (opt.ObjectFactoryLoaderType != null)
+        services.AddScoped(typeof(IObjectFactoryLoader), opt.ObjectFactoryLoaderType);
+      if (opt.ActivatorType != null)
+        services.AddTransient(typeof(IDataPortalActivator), opt.ActivatorType);
+      if (opt.ExceptionInspectorType != null)
+        services.AddScoped(typeof(IDataPortalExceptionInspector), opt.ExceptionInspectorType);
```
TheCakeMonster: I wonder if the code might be simpler and less error-prone if there was always an exception inspector, by creating a NullExceptionInspector for scenarios where the less secure option of allowing exceptions to be propagated was required.

Source/Csla/Server/DataPortal.cs
```diff
@@ -29,63 +30,44 @@ public class DataPortal : IDataPortalServer
...
-
+      Options = options;
+      Authorizer = authorizer;
+      InterceptorList = interceptors;
+      FactoryLoader = factoryLoader;
+      Activator = activator;
+      ExceptionInspector = exceptionInspector;
     }
 
     #region Data Access
```
TheCakeMonster, line 73: I think we need to think about the order of execution of the latest set of behaviours we have created. I have used an interceptor for triggering revalidation, but interceptors currently run prior to authorization being performed.

This might put the data portal at risk of amplification attacks, where a small action by an unauthenticated user/malicious agent - namely submission of an invalid request - can cause a significant amount of work - namely revalidation of all of the business rules.

I think authorization should come as early as possible - and that means before revalidation.

I can think of three options for solving this:

1. Move the call to AuthorizeRequest into the Initialize method. However, this makes it less visible.
2. Move the call to the interceptors out of the Initialize method and into its own method that is called after AuthorizeRequest.
3. Implement authorization using an interceptor instead.

Option 3 is the most flexible, because it allows people to put an interceptor before authorization if they so desire. However, this is less secure, in that people may not understand the risk of doing so. Generally, security should top flexibility if all else is equal, so option 2 seems the best to me purely from a security perspective.

Source/Csla/Server/ActiveAuthorizer.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    private ApplicationContext ApplicationContext { get; set; }
+
+    /// <summary>
+    /// Checks authorization rules for the request.
+    /// </summary>
+    /// <param name="clientRequest">
+    /// Client request information.
+    /// </param>
+    public void Authorize(AuthorizeRequest clientRequest)
```
TheCakeMonster, line 36: Should this only be applied when the location and logical location values are Server, so that it doesn't run when the data access operation is run locally, such as with [RunLocal]?

Source/Csla/Configuration/Fluent/DataPortalServerOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// injection.
+    /// </summary>
+    public Type AuthorizerProviderType { get; set; } = typeof(ActiveAuthorizer);
+
+    /// <summary>
+    /// Gets a list of the IInterceptDataPortal instances
+    /// that should be executed by the server-side data portal.
+    /// injection.
+    /// </summary>
+    public List<IInterceptDataPortal> InterceptorProviders { get; } = new List<IInterceptDataPortal>();
```
TheCakeMonster, line 32: This seems to accept constructed instances of interceptors. I don't think that is what we would want; instead we would want a list of types that we want to be instantiated by DI when we need them - so that they are provided any dependencies by the container at the point they are instantiated.

Source/Csla/Configuration/Fluent/DataPortalServerOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// <summary>
+    /// Gets or sets the type of the ExceptionInspector.
+    /// </summary>
+    public Type ExceptionInspectorType { get; set; } = typeof(DefaultExceptionInspector);
```
TheCakeMonster, line 37: As above, I think exposing generic methods to enable configuration is a better approach, as the people performing the configuration get compile-time checking of their config because we can apply constraints in the definition of the type parameter accepted by the method.

This applies to ExceptionInspectorType, ActivatorType and ObjectFactoryLoaderType

Source/Csla/Server/DataPortal.cs
```diff
@@ -29,63 +30,44 @@ public class DataPortal : IDataPortalServer
...
+      IDashboard dashboard, 
+      DataPortalServerOptions options,
+      IAuthorizeDataPortal authorizer,
+      List<IInterceptDataPortal> interceptors,
+      IObjectFactoryLoader factoryLoader,
+      IDataPortalActivator activator,
+      IDataPortalExceptionInspector exceptionInspector)
     {
       ApplicationContext = applicationContext;
       Dashboard = dashboard;
```
TheCakeMonster, line 64: Did you intend to be setting what is currently a static property from a constructor, or did you intend this to change to an instance property? The static declaration is shown above - it has changed, but only in its use of namespacing.

Source/Csla/Server/DataPortal.cs
```diff
@@ -195,7 +177,7 @@ private Reflection.ServiceProviderMethodCaller ServiceProviderMethodCaller
           error = ex;
         var fex = DataPortal.NewDataPortalException(
             ApplicationContext, "DataPortal.Create " + Resources.FailedOnServer,
-            new DataPortalExceptionHandler().InspectException(objectType, criteria, "DataPortal.Create", error),
+            new DataPortalExceptionHandler(ExceptionInspector).InspectException(objectType, criteria, "DataPortal.Create", error),
```
TheCakeMonster, line 180: I wonder if DataPortalExceptionHandler should be injected, and pick up the exception inspector automatically? However, that does mean we are not using a new instance on each and every call, so the lifetime is different.

I say this only because it would simplify the code a little - but then making the NewDataPortalException method an instance method could also be used to simplify it.

Source/Csla/Server/DataPortal.cs
```diff
@@ -577,17 +530,17 @@ internal void Complete(InterceptArgs e)
         Dashboard.CompleteCall(e);
       }
 
-      if (_interceptor != null)
-        _interceptor.Complete(e);
+      foreach (var interceptor in InterceptorList)
```
TheCakeMonster, line 533: In my implementation, I chose to iterate the interceptors backwards for the Complete method. This felt slightly more appropriate in that the earlier in the list they are, the more influence they are able to have on the result.

Source/Csla/Server/FactoryDataPortal.cs
```diff
@@ -17,43 +17,24 @@ namespace Csla.Server
...
-      }
+      ApplicationContext = applicationContext;
+      FactoryLoader = factoryLoader;
+      Inspector = inspector;
     }
 
-    #endregion
+    private ApplicationContext ApplicationContext { get; set; }
+    private IObjectFactoryLoader FactoryLoader { get; set; }
+    private IDataPortalExceptionInspector Inspector { get; set; }
```
TheCakeMonster, line 37: The IDataPortalExceptionInspector property was called ExceptionInspector in the DataPortal class; might it be more consistent if we named it that here, too?

Source/Csla/Server/SimpleDataPortal.cs
```diff
@@ -16,19 +16,24 @@ namespace Csla.Server
...
     {
       ApplicationContext = applicationContext;
+      Activator = activator;
+      Inspector = exceptionInspector;
     }
 
-    ApplicationContext Core.IUseApplicationContext.ApplicationContext { get => ApplicationContext; set => ApplicationContext = value; }
     private ApplicationContext ApplicationContext { get; set; }
+    private IDataPortalActivator Activator { get; set; }
+    private IDataPortalExceptionInspector Inspector { get; set; }
```
TheCakeMonster, line 36: Again, ExceptionInspector might be a slightly more consistent and clearer name.

Source/Csla/Server/DefaultExceptionInspector.cs
```diff
@@ -0,0 +1,30 @@
...
+// <summary>Default implementation of exception inspector.</summary>
+//-----------------------------------------------------------------------
+using System;
+
+namespace Csla.Server
+{
+  /// <summary>
+  /// Default implementation of exception inspector.
+  /// </summary>
+  internal class DefaultExceptionInspector : IDataPortalExceptionInspector
```
TheCakeMonster, line 15: I would like to create a sanitizing exception inspector too - but I can do that as a separate task.

Source/Csla/Configuration/Fluent/DataPortalClientOptions.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// Gets or sets a value indicating whether objects should be
+    /// automatically cloned by the data portal Update()
+    /// method when using a local data portal configuration.
+    /// </summary>
+    public bool AutoCloneOnUpdate
+    {
+      get
+      {
+        bool result = true;
+        string setting = ConfigurationManager.AppSettings["CslaAutoCloneOnUpdate"];
```
TheCakeMonster, line 26: I'm imagining the continued use of AppSettings here is an intentional decision - that there is a reason why this cannot, or should not, be defined in the startup code at compile time.

Source/Csla/Configuration/Fluent/CslaDataPortalConfiguration.cs
```diff
@@ -31,13 +45,33 @@ public static CslaDataPortalConfiguration DataPortal(this ICslaConfiguration con
...
+      return AddServerSideDataPortal(builder, null);
+    }
+
+    /// <summary>
+    /// Add services required to host the server-side
+    /// data portal.
+    /// </summary>
+    /// <param name="builder"></param>
+    /// <param name="options"></param>
+    public static CslaDataPortalConfiguration AddServerSideDataPortal(this CslaDataPortalConfiguration builder, Action<DataPortalServerOptions> options)
```
TheCakeMonster, line 57: This method was getting called automatically from the CslaConfigurationExtensions type's AddCsla method. I haven't seen this removed, so I wonder if we may get unusual behaviour from it being called twice - and with different instances of the configuration data, as the DataPortal() extension method returns a new instance of its type on each call.

This behaviour of the DataPortal() method caused me some pain when I was working on my solution, as it led to some nasty bugs in my code.

TheCakeMonster (commented):
Reviewed as requested.

I'm less comfortable with this version of the interceptor code, as it seems not to instantiate new instances of the interceptors, and therefore is not providing access to appropriately scoped instances of any dependencies. However it might be that it uses a DI technique that I am not aware of.

I have recommended use of generic methods in a number of places. I like that generic methods allow constraints to be applied to the type parameters, so that users get compile-time checks that their types implement the right interface.

Source/Csla/Configuration/Fluent/CslaDataPortalConfiguration.cs
rockfordlhotka: How would this work exactly? We don't want the calling code to create an instance, and I don't think we can restrict the type of the Type being provided?

Source/Csla/Server/DataPortal.cs
rockfordlhotka: I am trying to remember why the interceptor Initialize method is called before authorization. On the surface it surely seems like authorization should be called before initialize.

I wonder if the thought was that someone might write authorizer code that required the DI container be initialized, and so we called initialize first. I'm thinking that was the reason for the order that exists now.

However, there's no longer a valid scenario for someone to be initializing a DI container, because they should be integrating into the .NET DI subsystem. So the original purpose of the initialize/complete interceptor methods is entirely gone, and has morphed into a new purpose: things like running rules, etc.

So I think the answer is simply to move the `AuthorizeRequest` method above the call to `Initialize`, restoring authorization to the first thing that happens after deserialization.

Source/Csla/Configuration/Fluent/DataPortalServerOptions.cs
rockfordlhotka: I didn't _set out_ to replace your work - but I was in the thick of changing all the other code, and I thought I could save us both some effort :)

Source/Csla/Configuration/Fluent/CslaDataPortalConfiguration.cs
rockfordlhotka: You are right, in that in this latest bit of code I put in some Add methods instead of TryAdd. I've changed those now, so the second call (if there is one) of the `AddServerSideDataPortal` method doesn't really do anything.

Source/Csla/Configuration/Fluent/CslaDataPortalConfiguration.cs
rockfordlhotka: Doh! I see 😳

Source/Csla/Configuration/Fluent/DataPortalServerOptions.cs
```diff
@@ -40,7 +39,35 @@ public class DataPortalServerOptions
     /// that should be executed by the server-side data portal.
     /// injection.
     /// </summary>
-    public List<IInterceptDataPortal> InterceptorProviders { get; } = new List<IInterceptDataPortal>();
+    internal List<Type> InterceptorProviders { get; } = new List<Type>();
```
rockfordlhotka, line 42: @TheCakeMonster I think you'll like this approach for registering interceptor types better - let me know what you think.

Source/Csla/Configuration/Fluent/DataPortalClientOptions.cs
rockfordlhotka: Removed appsettings throughout

Source/Csla/Configuration/Fluent/DataPortalServerOptions.cs
TheCakeMonster: Yes, that looks good. Coupled with a generic method or two, plus the registration of the type against the interface, that should do us nicely :-)

PR #2664, opened 2021-12-21
Data portal exception sanitization inspector, implementing IDataPortalExceptionInspector

### LINKED ISSUE #2662: Create sanitizing exception inspector for server-side dataportal operations
Remote data portals are capable of exposing sensitive information to a client outside of the trust boundary by returning exceptions that occur during data access operations. Such information might include connection information, server-side data store names and so on - or even custom data that developers add to exceptions that they create, unaware of the potential risk that might be incurred.

It should be possible to filter the output from the server-side of data portal operations to stop any such information leaving the server. However, exception details should not be lost entirely, as this makes fault finding far too difficult. Instead the exception should be logged on the server including a unique identifier, and the return from the data portal should instead be a generic exception message containing that unique identifier. Developers can then look up the sensitive exception details of the server-side failure from their server-side logging subsystem using the unique identifier provided.

We could restrict exception sanitization to non-development environments; that is, exceptions could be thrown in development, but restricted in all other environments. This might offer a useful balance between developer productivity and good security. However, might this prove to be too confusing for people?

Ideally, this security-enhanced behaviour should be enabled by default, although this is the controversial element of the proposal.

**Proposed Solution:**
Create a class implementing the IDataPortalExceptionInspector interface, which accepts in its constructor an instance of ILogger to which to log the exception. Generate a unique identifier to be logged in the server-side exception message, using a format that fits with structured logging.

Create a generic exception message, but containing the unique Id added to the log entry. Rethrow this exception to have the data portal provide this to the caller on the client. Inclusion of the unique Id allows IT team members to tie the client and server exceptions together to help with fault finding and diagnosis.

The generic exception message should be drawn from resources to enable localization.

**Alternatives Considered**
An alternative would be to create one or more implementations of IInterceptDataPortal - perhaps one that logs the exception and another that performs sanitization. However, the IDataPortalExceptionInspector interface exists specifically to offer this exception inspection behaviour, so it seems a better approach.

I have also considered using the Activity class from .NET 5 onwards to enable use of the request's trace Id as the unique identifier for the exception. However, this seems to add complexity because it is only available in quite limited scenarios (it is currently restricted to .NET 5+ and only for Http requests to ASP.NET hosts.) For the moment I am suggesting that we take the simplest approach possible - use of a Guid - with the aim of gathering feedback. If users request it, we can then look at using Activity to gain access to the trace Id/correlation Id as an enhancement.

TheCakeMonster:
Sanitizing exception inspector, intended to reduce the risk of leaking of sensitive system information outside of the trust boundary of a remote data portal host.

Not enabled by default, for the moment. Enabling this places a dependency on the host to configure an implementation of both ILogger and IHostEnvironment, which feels slightly too heavy-handed at the moment. It is especially problematic for .NET Framework hosts, where DI is not such a baked in concept, and things like logging may not be configured by default.

This submission also includes fixes to the test projects which resolves the previous failures when trying to run them all at the same time.

Closes #2662

### DIFF
--- a/Source/Csla/Server/SanitizedServerSideDataPortalException.cs
+++ b/Source/Csla/Server/SanitizedServerSideDataPortalException.cs
@@ -16,8 +16,10 @@ namespace Csla.Server
   /// avoid the transmission of sensitive server-side information
   /// to the client in remote data portal operations
   /// </summary>
-  public class SanitizedServerSideDataPortalException : ApplicationException
+  public class SanitizedServerSideDataPortalException : Exception
   {
+
+    private const string IdentifierKey = "identifier";
 
     /// <summary>
     /// Constructor.
@@ -25,7 +27,27 @@ namespace Csla.Server
     /// <param name="identifier">The unique identifier for this exception</param>
     public SanitizedServerSideDataPortalException(string identifier) : base(Properties.Resources.SanitizedServerSideDataPortalException)
     {
-      this.Data.Add("identifier", identifier);
+      Data.Add(IdentifierKey, identifier);
+    }
+
+    /// <summary>
+    /// The unique identifier for the request that failed
+    /// Use this identifier to look for the exception details in server logs
+    /// </summary>
+    public string RequestIdentifier 
+    { get 
+      {
+        return Data[IdentifierKey].ToString();
+      }
+    }
+
+    /// <summary>
+    /// Override of the ToString method to insert the unique identifier
+    /// </summary>
+    /// <returns>String representation of the custom exception, including the custom tokens</returns>
+    public override string ToString() 
+    {
+      return string.Format(Properties.Resources.SanitizedServerSideDataPortalDetailedException, RequestIdentifier);
     }
   }
 }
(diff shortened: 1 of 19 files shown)


Source/Csla/Server/SanitizedServerSideDataPortalException.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Text;
+
+namespace Csla.Server
+{
+  /// <summary>
+  /// Sanitized server-side data portal exception; used to 
+  /// avoid the transmission of sensitive server-side information
+  /// to the client in remote data portal operations
+  /// </summary>
+  public class SanitizedServerSideDataPortalException : ApplicationException
```
rockfordlhotka: I think maybe just `ServerException` would be a better name? Short and to the point?

It'll end up in a `Csla.DataPortalException` on the client in any case - so just indicating that its origin is a general server exception (like a 500 http result) seems reasonable.

rockfordlhotka (changes requested):

Source/Csla/Server/SanitizedServerSideDataPortalException.cs
TheCakeMonster: Yes, it's certainly a bit long, now you mention it!

I agree, ServerException would work. However, I do wonder if that would be cybersquatting, in the sense that I would be using a name that might be useful for something else.

You'll note that internally this type uses a resource string which makes its function very specific. It's not really all that reusable. I suppose i could change that so that the message is passed in, but the token replacement that is performed does mean that the format of the string is important.

SanitizedException is an alternative I suppose.

I'll think on it for a few minutes before diving in and making the change.

rockfordlhotka (approved):

PR #2686, opened 2022-01-09
Recreation/upgrade of Csla.test project for detailed testing of Csla 6

TheCakeMonster:
Large rework of the code in the former Csla.test project to make it multiplatform and upgraded to compile and run on the Csla 6 codebase.

This is a large submission I'm afraid; I know that will cause you a problem with the review, sorry! My recommendation would be to limit your review to the changes outside of the new Csla.Tests project to begin with, with the plan that it is merged and then all of the remaining issues resolved between us.

I have created a new test project Csla.Tests.csproj, so that we can still refer to the old one Csla.test, and so that the old one can still run in the CI actions, if that is happening. I have also created a new, temporary solution Csla.test.Csla6.sln that includes the new test project. That allows any CI based on the original solution to continue running and pass all of the tests that exist other than in the new project. Once we are happy with Csla.Tests.csproj, we can add it to Csla.test.sln and drop the temporary Csla.test.Csla6.sln

There are a lot of these tests that are failing - but I need help to move forward. Some are for the same couple of bugs in the framework showing up repeatedly. Some are because I haven't managed to understand how to do the Csla 6 conversion for them - some of the concepts have changed since the tests were written, not all of which I understand. Hopefully none of them are for bugs in the tests, but you never know.

The submission includes a couple of small fixes in the Csla codebase, when it was obvious why the tests were failing due to bugs.

There is a TODO: explaining each of the places in code within Csla.Tests.csproj where I don't know how to finish the conversion/upgrade process. You can use Task Explorer in Visual Studio to examine each and make decisions on the right approach forwards in each case.

Forms part of issue #2544


rockfordlhotka:
How do you think we should proceed here?

We could accept this PR as-is and then go through the TODO comments in subsequent PRs.

Or we can analyze and work on each TODO as part of the review process for this PR. The drawback with this is that the PR could remain open for quite some time, and doing so might slow down other work.

I'm leaning toward accepting the PR as-is, but want to get your thoughts @TheCakeMonster .

TheCakeMonster:
@rockfordlhotka Yes, it sounds like we are thinking along the same lines. I was imagining that we merge it as-is and work on it from there.

It is unusual to merge imperfect code into main but the purpose of the tests is to help us track down the remaining bugs in the framework, so in this case it seems acceptable.

Merging now also makes the reviews easier for you; the bigger it gets, the harder it might be to make sense of the reason for each change. Small and steady might win this particular race.

We would be in no worse a position with the tests merged than we were before, as I have taken steps to ensure that they don't run as part of the CI process until we are ready. I've done that by adding the test project into a new, separate and temporary solution so that the status quo is maintained for now.

PR #2708, opened 2022-01-17
Blazor user/context management updates

### LINKED ISSUE #2682: Add a GetDataPortal method to the rule/auth rule context types
### Discussed in https://github.com/MarimerLLC/csla/discussions/2681

<sup>Originally posted by **mtavares628** January  7, 2022</sup>
Can someone give an example of how to utilize the DataPortal in a business rule for CSLA 6? Now that the DataPortal is being injected and not static, I'm not sure how to use it.  So for example I have this rule:

```
        private class IsLinkedToClient : BusinessRuleAsync
        {
            private IPropertyInfo NameProperty;

            public IsLinkedToClient(IPropertyInfo primaryProp, IPropertyInfo NameProp) : base(primaryProp)
            {
                NameProperty = NameProp;
                InputProperties = new List<IPropertyInfo>() { PrimaryProperty, NameProperty };
                RuleUri.AddQueryParameter("Name", NameProperty.Name);
            }

            protected override async Task ExecuteAsync(IRuleContext context)
            {
                int ID = context.GetInputValue<int>(PrimaryProperty);
                string Name = context.GetInputValue<string>(NameProperty);
                if (ID > 0)
                {
                    AnswersLinkedToClientCommand cmd = await DataPortal.CreateAsync<AnswersLinkedToClientCommand>(ID, (AnswersLinkedToClientCommand.FieldEnum)Enum.Parse(typeof(AnswersLinkedToClientCommand.FieldEnum), PrimaryProperty.Name));
                    cmd = await DataPortal.ExecuteAsync(cmd);
                    if (cmd is object && !cmd.IsLinked)
                    {
                        string desc;
                        desc = string.Format("{0} selected ({1}) is no longer linked for use with clients", PrimaryProperty.FriendlyName, Name);
                        context.AddWarningResult(desc);
                    }
                }
            }
        }
```
How can that be rewritten in CSLA 6?

Additionally, I have concerns with no longer being able to utilize static methods in my BOs.  I've been supporting a project utilizing CSLA since v3, and have been updating that project with each new iteration, but past iterations still allowed for me to utilize some older methodologies, when I couldn't quite make things fit with the newer ones. For example, a lot of my per type authorization rules needed more than just a user's role and required checking a database for additional security that may have been put on a record.  To make this work, I kept utilizing the old CanGetObject static functions from v3 and passed in the necessary parameters.  I would then call the static authorization methods from within the static factory methods. Here's an example:

```
        public static async Task<bool> CanGetObjectAsync(int clientID)
        {
            var result = default(bool);
            if (BusinessRules.HasPermission(AuthorizationActions.GetObject, typeof(Client)))
            {
                HasPermissionForClientCommand cmd = await DataPortal.CreateAsync<HasPermissionForClientCommand>(clientID, AccessPermissions.ReadAccess);
                cmd = await DataPortal.ExecuteAsync(cmd);
                result = cmd.HasPermission;
            }

            return result;
        }

        public async static Task<Client> GetClientAsync(int clientID)
        {
            var canGet = await CanGetObjectAsync(clientID);
            if (!canGet)
            {
                throw new System.Security.SecurityException("User not authorized to load this Client");
            }

            return await DataPortal.FetchAsync<Client>(clientID);
        }
```
While it is extra work, it is flexible enough so that I have the authorization I need on my objects.  However, with the new injected DataPortal, I'm not sure how to make this work anymore, since I can no longer use the static methods.  This was why I was inquiring about async authorization rules last month  (#2643), because the current out of the box solution for CSLA 6 won't work in blazor which requires async dataportal methods.  Is there some workaround that I can do until async auth rules are an option?

Finally, I feel like while moving to an injected DataPortal seems to be necessary, the tradeoff is losing a lot of the nice encapsulation that CSLA's BOs provided. It was nice being able create static methods so that when instantiating a business object, you knew what parameters were needed with the static factory method. Now it looks like those parameters are just a param array being called from the UI's injected DataPortal, which loses the ability to pick up bugs at compile time.  

Additionally, things like additional methods that are pertinent to the BO, but outside the scope of Create/Fetch/Insert/Update/Delete can now no longer be easily encapsulated. For example on the ProjectEdit BO in your Project Tracker Project:

```
    public static async Task<bool> ExistsAsync(int id)
    {
      var cmd = await DataPortal.CreateAsync<ProjectExistsCommand>(id);
      cmd = await DataPortal.ExecuteAsync(cmd);
      return cmd.ProjectExists;
    }
```

I assume that now, that check will have to be made directly from the UI and inject the DataPortal with the ProjectExistsCommand. So instead of the user just needing to work with ProjectEdit, they have to work with the command BO directly. 

This could also just be my lack of understanding with the new methodologies, and there is a way to use injection while still maintaining this encapsulation. Please let me know if there is a way to do any of the 3 things I've mentioned.

I also want to state that I'm not complaining here. I just want to know how to make this stuff work. I think CSLA is amazing, and have been using it for the past 17 years. I appreciate all the thought and work that goes into it. Hopefully, I will have a better understanding of how to make these things work once more examples are available.</div>

### LINKED ISSUE #2695: Server-side data portal should use a single DI scope
### Discussed in https://github.com/MarimerLLC/csla/discussions/2685

The server-side data portal _should_ create a DI scope on first call from the client. Subsquent data portal calls from _within the existing server-side context_ should run within that same DI scope.

Right now it appears that a new DI scope is created on each data portal call from within a server-side data portal operation method.

In other words, the client calls something like `FetchAsync`, resulting in code running on the server in a DI scope. The data portal operation method _then_ calls something like `FetchAsync`, which also runs on the server. But right now that second call is resulting in a new DI scope, which is not the correct behavior.

<sup>Originally posted by **swegele** January  8, 2022</sup>
I have a DI service which wraps an instance of my DAL db connection object.  I want this wrapper to be scoped to ensure that I have only one DB connection open at a time for this particular user's call to the DataPortal.  
```cs
public class JTDataAdapterManager
{
    private JTDataAccessAdapter _adapter;  //DAL connection
    private Guid guid = Guid.NewGuid();

    //Use DI
    public JTDataAdapterManager(ApplicationContext applicationContext, DBConfiguration dbConfigOptions)
    {
        //Set DAL Config options here with dbConfigOptions
        ...

            // open connection if we have crossed the logical and/or physical dataportal so it is safe to access the database
            if (applicationContext.LogicalExecutionLocation == ApplicationContext.LogicalExecutionLocations.Server)
            {
                _adapter = new JTDataAccessAdapter(true, connectionTimeoutInt);
            }
            else
            {
                throw new Exception("The Database can only be accessed on the server side of DataPortal");
            }
    }

```

The service is defined in my server side Blazor app as "Scoped"
```cs
builder.Services.AddScoped<LLBLGenDataUtilities.JTDataAdapterManager>();

```

I assumed since the service is defined as "Scoped" that the first call to this on the server side of the DataPortal would create a new one of these...but subsequent chained calls (still on the server side) would pick up the existing created one.  Example, here is an instance of a server side [Fetch] method that gets called first: 
```cs
public class TranslationRecordInfoList : BusinessBaseClasses.JTReadOnlyListBase<TranslationRecordInfoList, TranslationRecordInfo>
{
   ...

  [Fetch]
  private void DP_Fetch(
      LLBLGenDataUtilities.Criteria.LLBLGenFilterCriteria criteria,
      [Inject] LanguageInfoListFactory languageInfoListFactory,
      [Inject] TranslationRecordInfoFactory translationInfoFactory, 
      [Inject] LLBLGenDataUtilities.JTDataAdapterManager cxnManager)
{
      RaiseListChangedEvents = false;
      IsReadOnly = false;

      using (EntityCollection<TranslationEntity> list = new EntityCollection<TranslationEntity>())
      {
          // load values
          cxnManager.ActiveAdapter.FetchEntityCollection(list, criteria.Filter, criteria.MaxRecordsToReturn, criteria.SortExpression);

          if (list.Count > 0)
          {
              LanguageInfoList languages = languageInfoListFactory.GetAll();
              foreach (var language in languages)
              {
                  foreach (var entity in list.Where(ent => ent.LanguageId == language.Id))
                  {
                      Add(TranslationRecordInfo.Load(translationInfoFactory, entity, language));
                  }
              }
          }
      }

      IsReadOnly = true;
      RaiseListChangedEvents = true;
  }
```

This is the first method that gets called in a chain of calls...JTDataAdapterManager is injected so this is where one should get created.  A few lines into that method, I fetch another business object I need
```cs
    LanguageInfoList languages = languageInfoListFactory.GetAll();
```

That call ends up here in the LanguageInfoList object
```cs
public class LanguageInfoList : BusinessBaseClasses.JTReadOnlyListBase<LanguageInfoList, LanguageInfo>
{
  ...

[Fetch]
private void DP_Fetch(
    LLBLGenDataUtilities.Criteria.LLBLGenFilterCriteria criteria,
    [Inject] LanguageInfoFactory languageInfoFactory,
    [Inject] LLBLGenDataUtilities.JTDataAdapterManager cxnManager)
{
	RaiseListChangedEvents = false;
	IsReadOnly = false;

```

The problem is I check the "guid" property of the cxnManager in the second call and it is a new GUID (different than the one in the first fetch method.  I was not expecting that.  I thought once you cross the DataPortal, the LocalProxy had created a scope which would apply for DI until you end that original call into the DP.

I used have this class just like Rocky's where I did ref counting and stuck an instance of it inside the ApplicationContext.  That worked fine and I "could" keep doing that, but I figured I could get rid of all that extra stuff and really simplify this class by just defining it as scoped.
Obviously "scoped" is not behaving the way I expected so I'm missing something.  
</div>

### LINKED ISSUE #2702: LocalProxy creates a new instance of the built-in ApplicationContextManager, resulting in loss of context
Whilst debugging the strange behaviour of a business object on which IsSavable was false despite IsValid and IsDirty both being true, I've discovered that the application context gets lost during the execution of a request through LocalProxy.

My object has the correct ApplicationContext - exposed by a custom ApplicationContextManager - until line 41 of LocalProxy.cs, at which point the default, AsyncLocal implementation of ApplicationContextManager is instantiated. At this point, the ApplicationContext is lost, because the new context manager doesn't have it.

It appears from the comment that the creation of a separate context manager is intentional. Should this new implementation of IContextManager be created from the scoped service provider created a few lines above rather than being created using the constructor of a specific type? Or is the use of a different and very specific context manager intentional, and therefore it should have its context set from the ambient/client-side instance at that moment, copying it from one place to another?

### LINKED ISSUE #2706: Csla.Blazor needs its own ApplicationContextManager type
Blazor relies on the `AuthenticationStateProvider` service to get the current user, not HttpClient, and so it needs an `ApplicationContextManager` type that uses this service to obtain the current user.

`ApplicationContextManager` is a scoped service, and so should maintain local/client context values as private properties.

Lots of discussion in #2687

### LINKED ISSUE #2707: Rework the Blazor authentication types in Csla.Blazor
The Csla.Blazor.Client.Authentication types should be in a Csla.Blazor.Authentication namespace.

They can/should be written to work on client and server.

The `CslaUserService` type should do nothing more than maintain a reference to the current user, and raise an event when that value changes. No reference to any other type or instance.

The `CslaAuthenticationStateProvider` should maintain a reference to an injected `CslaUserService` instance, listening for the changed event, and raising appropriate events based on the changes to the authentication state. No references to other instances.

The `Csla.Blazor.AuthenticationStateManager`, as per #2706 will get the current user from the registered `AuthenticationStateManager` service, which might be this simple CSLA one, or another one from Microsoft or other implementation providers.

rockfordlhotka:
Closes #2706 
Closes #2707 
Closes #2682 
Closes #2702 
Closes #2695


rockfordlhotka:
@swegele you may want to review these changes to see if they are meeting your needs.

swegele:
> @swegele you may want to review these changes to see if they are meeting your needs.

Looks like this does the trick.  :-)

Source/Csla/DataPortalClient/LocalProxy.cs
```diff
@@ -30,19 +30,18 @@ public LocalProxy(ApplicationContext applicationContext, LocalProxyOptions optio
       ApplicationContext = applicationContext;
       Options = options;
 
-      if (Options.CreateScopePerCall)
+      if (Options.CreateScopePerCall && ApplicationContext.ExecutionLocation == ApplicationContext.ExecutionLocations.Server)
```
rockfordlhotka, line 33: I am thinking this is not right.

I think a new scope should be created only if the _logical_ execution location is Client. After that first call, the data portal will have set the logical execution location to server, and subsequent calls through LocalProxy would then not create a new scope.

swegele: Gosh I thought that but didn't say anything!  Ugh - trust your gut Sean.  Glad you caught it.

swegele: That is such a nice feature being able to distinguish the diff between logical and physical :-)

Source/Csla.Blazor/ApplicationContextManager.cs
```diff
@@ -43,15 +44,25 @@ public class ApplicationContextManager : IContextManager, IDisposable
...
       AuthenticationStateProvider.AuthenticationStateChanged += AuthenticationStateProvider_AuthenticationStateChanged;
-      try
-      {
-        CurrentPrincipal = AuthenticationStateProvider.GetAuthenticationStateAsync().Result.User;
-      }
-      catch
+      InitializeUser();
+    }
+
+    private async void InitializeUser()
```
rockfordlhotka, line 52: @TheCakeMonster I think this will work better - it avoids the sync over async issue.

Source/Csla.Blazor/Authentication/CslaAuthenticationStateProvider.cs
```diff
@@ -10,6 +10,7 @@ namespace Csla.Blazor.Client.Authentication
   public class CslaAuthenticationStateProvider : AuthenticationStateProvider
```
rockfordlhotka, line 10: @TheCakeMonster you were wondering if we should even have the CSLA authenticationstateprovider and user service.

I think we do need them, b/c _some_ `AuthenticationStateProvider` must be registered for CSLA to initialize at all (app context requires the service). If I just create a Blazor app with no authn at all, I don't believe any provider is registered, which would prevent your app from running, due to a missing required service.

So at the very least, I think we need to provide simplistic service implementations that support the bare minimum Blazor project templates.

rockfordlhotka: Your original comment:

> I have slightly mixed feelings about the presence of CslaAuthenticationStateProvider and CslaUserService. Their presence does support simpler demo code, but do not offer the level of detail of the built in state providers that come in the box with Blazor. It also starts to project Csla into the security space in a way that isn't necessary, putting the framework at risk of containing vulnerabilities.
> When people create a solution using either of the Blazor templates, they already get a state provider, so it might even be more confusing to have Csla versions available. People who don't know this security model well might think that they need to make use of these implementations, when that is not the case.
> The benefit of the Blazor ApplicationContextManager is that it can work with any state provider; as long as it responds to the AuthenticationStateChanged event, then the ones that are provided by Microsoft are perfectly appropriate.
> The only thing I am not absolutely sure of is whether the steps detailed in the book to create the multi-headed solution creates the state providers; I guess this is based on project templates and not solution templates. I know for sure that creating standalone solutions for either Blazor Server or Blazor WebAssembly definitely does create the state providers.

You say here that you are confident that the minimal Blazor project templates (w/o any authn) do register some state provider. If that's the case, then I tend to agree with you that we should use the "dummy" Microsoft ones rather than providing a duplicate "dummy" implementation.

Source/Csla.Blazor/ApplicationContextManager.cs
TheCakeMonster: Hmm, not sure. It's using async void - and I think it's being called synchronously. I think that might risk the same problem, although we're quite far down into the weeds of how async/await works, I know. But I'd say that will still deadlock.

I think attaching a continuation (I think of these like a callback) using ContinueWith on the task that is returned by the GetAuthenticationStateAsync() method avoids the risk. Don't await the task, just tack a continuation/callback onto the end and let that get executed when the task has completed.

Source/Csla.Blazor/ApplicationContextManager.cs
```diff
@@ -49,19 +49,14 @@ public ApplicationContextManager(AuthenticationStateProvider authenticationState
       InitializeUser();
     }
 
-    private async void InitializeUser()
+    private void InitializeUser()
```
rockfordlhotka, line 52: @TheCakeMonster I think this is what you suggested? It does seem cleaner and probably safer.

Source/Csla.Blazor/Authentication/CslaAuthenticationStateProvider.cs
rockfordlhotka: I just did a quick test. It appears that some `AuthenticationStateProvider` is registered and available. It contains a null `AuthenticationState` reference.

So I think you are right, we can remove the CSLA implementations of these types and work with the default implementation.

Source/Csla.Blazor/ApplicationContextManager.cs
TheCakeMonster: Yes, that looks good.

I wrote it out as a separate method to make the readability as good as it could get, but the lambda works too.

Source/Csla.Blazor/ApplicationContextManager.cs
```diff
@@ -0,0 +1,162 @@
...
+        if (task.IsCompletedSuccessfully && task.Result != null)
+          CurrentPrincipal = task.Result.User;
+        else
+          CurrentPrincipal = UnauthenticatedPrincipal;
+      });
+    }
+
+    private void AuthenticationStateProvider_AuthenticationStateChanged(System.Threading.Tasks.Task<AuthenticationState> task)
+    {
+      CurrentPrincipal = task.Result.User;
```
TheCakeMonster, line 65: I think this probably should use the same technique of accessing the Result only after a continuation has run. I don't know for sure that the even would be raised before the result is available, but I think it's possible based upon the design of the API.

Source/Csla/DataPortalClient/LocalProxy.cs
```diff
@@ -30,19 +30,19 @@ public LocalProxy(ApplicationContext applicationContext, LocalProxyOptions optio
...
+        ApplicationContext.LogicalExecutionLocation == ApplicationContext.LogicalExecutionLocations.Client)
       {
         // create new DI scope and provider
         _scope = ApplicationContext.CurrentServiceProvider.CreateScope();
         var provider = _scope.ServiceProvider;
 
         // create and initialize new "server-side" ApplicationContext and manager
         var parentContext = applicationContext;
-        var ctx = new Csla.Core.ApplicationContextManager();
+        var ctx = provider.GetRequiredService<Core.IContextManager>();
```
TheCakeMonster, line 42: I wonder if the ctx variable would be better called mgr; I think this is the context manager and not the context at this point. This makes no difference to execution, only to whether there is a chance for greater understanding for those who come after us.

TheCakeMonster (commented):
There are a couple of changes that I don't feel able to validate here, particularly around lazy loading of properties. This isn't something I understand yet, as I have never used it. There are also some changes to rules; they seem to entirely make sense, to enable data access, but I can't say I feel qualified to yay or nay specifically.

The Blazor stuff is starting to look rock solid reliable and usable in both server-side and WASM scenarios; great progress. However, I have raised a question about the event handler. I'd be tempted to do that with a continuation as well, to be sure that the task is complete before accessing the property. To some extent this might be unnecessary; the idea that something or someone would raise an event before knowing the result of the action seems odd - but unfortunately the API allows for it, so I think we should handle it, so that the framework users don't suffer odd edge cases from custom state provider implementations.

PR #2709, opened 2022-01-18
Fix issue with lazy loading; clean up implementation

### LINKED ISSUE #2705: Lazy Load Properties throwing InvalidOperationException
**Describe the bug**
Lazy-loaded properties throw InvalidOperationException when the fetch method is not fast and the property info's relationship type is "LazyLoad". 
It happens because ReadProperty in Csla.Core.BusinessBase does not check if the property is loading before thrown the exception.
```c#
if (((propertyInfo.RelationshipType & RelationshipTypes.LazyLoad) == RelationshipTypes.LazyLoad) && !FieldManager.FieldExists(propertyInfo))
        throw new InvalidOperationException(Resources.PropertyGetNotAllowed);
```

**Version and Platform**
CSLA version: 5.5.2
OS: Windows
Platform: WinForms, WPF

**Code that Fails**
```c#
namespace Csla.Test.Basic
{
  [TestClass]
  public class LazyLoadTests
  {
    [TestMethod]
    public void LazyGet()
    {
      IDataPortal<LazyLoadTest> dataPortal = DataPortalFactory.CreateDataPortal<LazyLoadTest>();

      var obj = dataPortal.Fetch("root");
      Assert.AreEqual("root", obj.Name);

      // This code works becase property is not marked as Lazy Load
      var childWorking = obj.ChildWorking;
      Assert.IsNull(childWorking);

      // This code fails - the delay inside Fetch is important to reproduce the error
      var childException = obj.ChildException;
      Assert.IsNull(childException);
    }
  }

  [Serializable]
  public class LazyLoadTest
    : BusinessBase<LazyLoadTest>
  {
    public static readonly PropertyInfo<string> NameProperty = RegisterProperty<string>(nameof(Name));
    public string Name
    {
      get => GetProperty(NameProperty);
      set => SetProperty(NameProperty, value);
    }

    public static readonly PropertyInfo<LazyLoadTest> ChildWorkingProperty = RegisterProperty<LazyLoadTest>(nameof(ChildWorking));
    public LazyLoadTest ChildWorking
    {
      get => LazyGetPropertyAsync(ChildWorkingProperty, GetDataPortal<LazyLoadTest>().FetchAsync("ChildWorking", 500));
      set => SetProperty(ChildWorkingProperty, value);
    }

    public static readonly PropertyInfo<LazyLoadTest> ChildDelayedProperty = RegisterProperty<LazyLoadTest>(nameof(ChildException), RelationshipTypes.LazyLoad);
    public LazyLoadTest ChildException
    {
      get => LazyGetPropertyAsync(ChildDelayedProperty, GetDataPortal<LazyLoadTest>().FetchAsync("ChildException", 500));
      set => SetProperty(ChildDelayedProperty, value);
    }

    [Fetch]
    private void Fetch(
      string name
      )
    {
      using (BypassPropertyChecks)
      {
        Name = name;
      }
    }

    [Fetch]
    private async Task Fetch(
      string name,
      int millisecondsDelay
      )
    {
      using (BypassPropertyChecks)
      {
        Name = name;

        await Task.Delay(millisecondsDelay);
      }
    }
  }
}
```

rockfordlhotka:
Closes #2705

### DIFF
--- a/Source/Csla/Core/BusinessBase.cs
+++ b/Source/Csla/Core/BusinessBase.cs
@@ -1864,7 +1864,14 @@ namespace Csla.Core
     /// </remarks>
     protected P GetProperty<P>(PropertyInfo<P> propertyInfo, Security.NoAccessBehavior noAccess)
     {
-      P result = default(P);
+      if (((propertyInfo.RelationshipType & RelationshipTypes.LazyLoad) == RelationshipTypes.LazyLoad) && !FieldManager.FieldExists(propertyInfo))
+      {
+        if (PropertyIsLoading(propertyInfo))
+          return default;
+        throw new InvalidOperationException(Resources.PropertyGetNotAllowed);
+      }
+
+      P result = default;
       if (_bypassPropertyChecks || CanReadProperty(propertyInfo, noAccess == Csla.Security.NoAccessBehavior.ThrowException))
         result = ReadProperty<P>(propertyInfo);
       else
@@ -1884,7 +1891,7 @@ namespace Csla.Core
     /// </remarks>
     protected object GetProperty(IPropertyInfo propertyInfo)
     {
-      object result = null;
+      object result;
       if (_bypassPropertyChecks || CanReadProperty(propertyInfo, false))
       {
         // call ReadProperty (may be overloaded in actual class)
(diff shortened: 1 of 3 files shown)


Eduardo-Micromust (commented):
When the property isn't loaded the "default" is returned:
```
if (PropertyIsLoading(propertyInfo))
          return default;
```

Is the code below better in this case?

```
if (PropertyIsLoading(propertyInfo))
          return propertyInfo.DefaultValue;
```

For instance, I could return an empty list while the property is loading.

rockfordlhotka:
Thank you @Eduardo-Micromust, you are absolutely correct!

PR #2717, opened 2022-01-21
Blazor ViewModel: call OnModelChanged

peranborkett:
Seems like this line was missing as there were no usages of OnModelChanged


rockfordlhotka:
Thank you for catching this and submitting a fix!

PR #2726, opened 2022-01-24
Fluent configuration API updates

### LINKED ISSUE #2692: Csla 6 fluent configuration - methods on the DataPortalServerOptions class should support method chaining
During configuration using the Csla 6 AddCsla method, it's not possible to chain configuration using methods that are exposed by the DataPortalServerOptions class, as they do not expose the type on which they are placed.

The following is not supported, for example - this would not compile, as RegisterActivator is a void method:
```csharp
.DataPortal(options => options.AddServerSideDataPortal(cfg => cfg
.RegisterActivator<CustomActivator>()
.AddInterceptorProvider<CustomInterceptor>()
));
```

This:
```csharp
public void AddInterceptorProvider<T>() where T: IInterceptDataPortal
{
      InterceptorProviders.Add(typeof(T));
}
```
would become this:
```csharp
public DataPortalServerOptions AddInterceptorProvider<T>() where T: IInterceptDataPortal
{
      InterceptorProviders.Add(typeof(T));
      return this;
}
```
Optionally, I could also move the methods to a new file called DataPortalServerConfigurationExtensions and make them into extension methods if we'd prefer to make them a little more consistent with other parts of the codebase.

I can't see any downside to this suggestion, so I have created this directly as an issue because I think it likely that you will be happy with the change and want me to make it.

### LINKED ISSUE #2693: Csla 6 fluent configuration - should the builder parameters be renamed config?
This is a very minor point, but I've found it slightly confusing while working on the Csla fluent configuration code (inside of the framework, while I had the hood open) that some of the parameters in the fluent configuration classes are called builder, but we don't have a configuration builder that I can see.

I wonder if this parameter would be better named config, as it is the parameter that contains configuration data.

This applies to methods in classes including DataPortalConfigurationExtensions.cs. I thought it applied to more than that, although I can't see many other references at the moment. If I change it, I'll change it consistently.

```csharp
public static DataPortalClientOptions AddServerSideDataPortal(this DataPortalClientOptions builder, Action<DataPortalServerOptions> options)
{
  var opt = builder.DataPortalServerOptions;
  options?.Invoke(opt);
  return builder;
}
```
I associate the word builder with a class that has a Build method, or at least something that outputs a build artefact (and instance of a class.) Currently the AddCsla() method in ConfigurationExtensions.cs is acting as the way to define and build configuration.

I don't think this is a breaking change as it's the name of the parameter indicating the type being extended, but I'm not entirely confident about that.

### LINKED ISSUE #2727: Add RegisterContextManager method to fluent config
It would be more clear to have an explicit method for people to register a custom context manager than the current approach, where they must register an `IContextManager` service manually.

rockfordlhotka:
Closes #2692 
Closes #2693 
Closes #2727

### DIFF
--- a/Source/Csla/Configuration/ConfigurationExtensions.cs
+++ b/Source/Csla/Configuration/ConfigurationExtensions.cs
@@ -54,13 +54,13 @@ namespace Csla.Configuration
       services.TryAddTransient((p) => new Channels.Local.LocalProxyOptions());
       services.AddTransient<Channels.Local.LocalProxy, Channels.Local.LocalProxy>();
       // Add server-side data portal; this is always necessary to support RunLocal
-      cslaOptions.DataPortal().ApplyConfiguration();
+      cslaOptions.DataPortal(options => options.ApplyConfiguration());
 
       // Default to using LocalProxy and local data portal
       var proxyInit = services.Where(i => i.ServiceType.Equals(typeof(IDataPortalProxy))).Any();
       if (!proxyInit)
       {
-        cslaOptions.DataPortal().UseLocalProxy();
+        cslaOptions.DataPortal((options) => options.UseLocalProxy());
       }
 
       return services;
--- a/Source/Csla/Configuration/ConfigurationExtensions.cs
+++ b/Source/Csla/Configuration/ConfigurationExtensions.cs
@@ -50,11 +50,7 @@ namespace Csla.Configuration
       services.TryAddTransient(typeof(IDataPortal<>), typeof(DataPortal<>));
       services.TryAddTransient(typeof(IChildDataPortal<>), typeof(DataPortal<>));
 
-      // LocalProxy is always necessary to support RunLocal
-      services.TryAddTransient((p) => new Channels.Local.LocalProxyOptions());
-      services.AddTransient<Channels.Local.LocalProxy, Channels.Local.LocalProxy>();
-      // Add server-side data portal; this is always necessary to support RunLocal
-      cslaOptions.DataPortal(options => options.ApplyConfiguration());
+      cslaOptions.DataPortalClientOptions.AddRequiredDataPortalServices();
 
       // Default to using LocalProxy and local data portal
       var proxyInit = services.Where(i => i.ServiceType.Equals(typeof(IDataPortalProxy))).Any();
--- a/Source/Csla/Configuration/Fluent/CslaOptions.cs
+++ b/Source/Csla/Configuration/Fluent/CslaOptions.cs
@@ -32,6 +32,19 @@ namespace Csla.Configuration
     }
 
     /// <summary>
+    /// Registers a specific ApplicationContext manager service,
+    /// overriding the default service registration.
+    /// </summary>
+    /// <typeparam name="T">IContextManager implementation type.</typeparam>
+    /// <returns></returns>
+    public CslaOptions RegisterContextManager<T>() 
+      where T : IContextManager
+    {
+      Services.AddScoped(typeof(IContextManager), typeof(T));
+      return this;
+    }
+
+    /// <summary>
     /// Sets a value indicating whether CSLA
     /// should fallback to using reflection instead of
     /// System.Linq.Expressions (true, default).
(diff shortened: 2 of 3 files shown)


Source/Csla/Configuration/ConfigurationExtensions.cs
```diff
@@ -54,13 +54,13 @@ public static IServiceCollection AddCsla(this IServiceCollection services, Actio
       services.TryAddTransient((p) => new Channels.Local.LocalProxyOptions());
       services.AddTransient<Channels.Local.LocalProxy, Channels.Local.LocalProxy>();
       // Add server-side data portal; this is always necessary to support RunLocal
-      cslaOptions.DataPortal().ApplyConfiguration();
+      cslaOptions.DataPortal(options => options.ApplyConfiguration());
```
TheCakeMonster, line 57: Hmm, yeah, that looks a bit unusual, doesn't it?

Can we do something like this instead - solely from _inside_ of Csla, never from outside:

```csharp
cslaOptions.DataPortalClientOptions.ApplyConfiguration();
```

rockfordlhotka: `ApplyConfiguration` is not public, so yes, we should be able to do that.

Source/Csla/Configuration/Fluent/CslaOptions.cs
```diff
@@ -31,6 +31,19 @@ public CslaOptions(IServiceCollection services)
...
+    /// Registers a specific ApplicationContext manager service,
+    /// overriding the default service registration.
+    /// </summary>
+    /// <typeparam name="T">IContextManager implementation type.</typeparam>
+    /// <returns></returns>
+    public CslaOptions RegisterContextManager<T>() 
+      where T : IContextManager
+    {
+      Services.AddScoped(typeof(IContextManager), typeof(T));
+      return this;
```
TheCakeMonster, line 44: Yeah, this is great. Very discoverable from a framework user's perspective.

Long term, we should maybe consider what should be an extension method in a separate file, and what should be within the options class itself. 

There is an option to keep the options classes as simple as possible by having only properties on them. All of the methods could be extension methods. I think this is maybe what is more normal in this type of code. However, this new addition is entirely consistent with the existing methods in this class, so we should probably run with it for now and refactor at a later date, if we get the desire.

Source/Csla/Configuration/ConfigurationExtensions.cs
```diff
@@ -54,13 +54,13 @@ public static IServiceCollection AddCsla(this IServiceCollection services, Actio
...
 
       // Default to using LocalProxy and local data portal
       var proxyInit = services.Where(i => i.ServiceType.Equals(typeof(IDataPortalProxy))).Any();
       if (!proxyInit)
       {
-        cslaOptions.DataPortal().UseLocalProxy();
+        cslaOptions.DataPortal((options) => options.UseLocalProxy());
       }
```
TheCakeMonster, line 64: As above, I think there is a difference between how it is accessed by the end user and how we access it internally. There is nothing wrong with this code, but there is something about the way it reads that makes it look like we are mixing up config from application of config.

Most of the time, mixing the two concepts up doesn't matter; TryAdd<blah> sorts out the confusion. However, occasionally, it does matter if you mix them up. The interceptors code didn't work when we mixed these two concepts. In that scenario we are registering multiple types, and calling the same code twice caused each type to be registered twice; instead of 2 interceptors, we had 4.

TheCakeMonster (commented):
Yeah, nice consistency being achieved in the external interface as a result of these changes I think. Good enhancement.

Source/Csla/Configuration/Fluent/CslaOptions.cs
rockfordlhotka: I originally thought about an extension method - but then there's no access to the services collection - so we'd need to sort that out to support extension methods.

Source/Csla/Configuration/ConfigurationExtensions.cs
rockfordlhotka: Not anymore - now end users will be writing this code too. Because the `DataPortal` method no longer returns a unique type, we can't write extension methods over it. It now returns a `CslaOptions` instance and so resets to the base level of extension methods.

TheCakeMonster: Sorry, I think my comment was maybe a bit ambiguous. I think what I meant was that the following might read better from _within_ the framework:

```csharp
cslaOptions.DataPortalClientOptions.UseLocalProxy();
```
 
I know this has changed anyway, but the point was about the usage/selection of the coding styles. The lambdas look like they set config and are available to end users, and the direct method call looks like it applies that config and can only be done inside of the framework.

No matter.

PR #2739, opened 2022-02-01
Handle aspnetcore, SSB, Blazor wasm context and current user separately

### LINKED ISSUE #2724: Add IContextManager implementation for hybrid Blazor/WebAPI, Blazor/MVC or Blazor/Razor Pages scenarios
We have a new Blazor ApplicationContextManager which plays very nicely with Blazor. However, if a Blazor Server-side site has any WebAPI controllers, MVC controllers or Razor Pages then the context manager will not work for those sections of the ASP.NET site. If someone needs to access anything via Csla from outside of Blazor then they will not be able to add authorization rules, as the context of the user would be unavailable in the non-Blazor sections - in other words, the GetUser() method of the context manager would return an unauthenticated principal to a WebAPI or MVC controller or Razor Page, even if a valid authentication cookie is presented with the HTTP request to the controller/page.

It is specifically - and almost exclusively - the behaviour of GetUser() that I think should behave better in hybrid scenarios. I am not suggesting that any state should be shared between Blazor and non-Blazor as such, but that all code in ASP.NET that is outside of Blazor should function as it always has - based on HttpContext, including/especially HttpContext.Current.User - whilst everything inside of Blazor should function as we have now placed into the new Blazor context manager.

**Proposed Solution**
I suggest we provide a hybrid context manager which determines whether it is within Blazor or outside of it, and within its constructor news up an instance of either the Blazor context manager (based on scoped objects from DI) or the ASP.NET one (based on HttpContext.) All of the IContextManager implementation is then delegated to the internal context manager that was created in the constructor.

The hybrid provider solely needs to make the determination of which other context manager to create, then delegate to it.

**Alternatives I Have Considered**
We could leave this out, and expect the end user to implement it. However, I think this would lead to more requests for help. I think it would be better to provide the implementation inside of the framework; then all we need do is point people to the correct implementation for their hybrid scenario.

**Additional context**
A full implementation of the solution was mocked up in a demo available at https://github.com/TheCakeMonster/Examples/tree/master/CSLA6/BlazorAuthentication

This demo includes the Blazor context manager and the hybrid one. The Blazor context manager has now been included in the framework, so the only thing remaining to be done that can be seen in the demo is the hybrid context manager.

rockfordlhotka:
Closes #2724

### DIFF
--- a/Source/Csla.AspNetCore/ApplicationContextManager.cs
+++ b/Source/Csla.AspNetCore/ApplicationContextManager.cs
@@ -3,38 +3,53 @@
 //     Copyright (c) Marimer LLC. All rights reserved.
 //     Website: https://cslanet.com
 // </copyright>
-// <summary>Application context manager that uses HttpContextAccessor</summary>
+// <summary>Application context manager that adapts for AspNetCore/SSB</summary>
 //-----------------------------------------------------------------------
 using Csla.Core;
+using System;
+#if NET5_0_OR_GREATER
+using Csla.AspNetCore.Blazor;
+using Microsoft.Extensions.DependencyInjection;
+#else
 using Microsoft.AspNetCore.Http;
-using System.Security.Claims;
+#endif
 
 namespace Csla.AspNetCore
 {
   /// <summary>
-  /// Application context manager that uses HttpContextAccessor when 
-  /// resolving HttpContext to store context values.
+  /// Application context manager that adapts for AspNetCore/SSB.
   /// </summary>
-  public class ApplicationContextManager : IContextManager
+  public class ApplicationContextManager : IContextManager, IDisposable
   {
+    private IContextManager ActiveContextManager { get; set; }
 
-    private const string _localContextName = "Csla.LocalContext";
-    private const string _clientContextName = "Csla.ClientContext";
-
+#if NET5_0_OR_GREATER
     /// <summary>
-    /// Creates an instance of the object, initializing it
-    /// with the required IServiceProvider.
+    /// Creates an instance of the object.
     /// </summary>
-    /// <param name="httpContextAccessor">HttpContext accessor</param>
+    /// <param name="provider"></param>
+    /// <param name="activeCircuitHandler"></param>
+    public ApplicationContextManager(IServiceProvider provider, ActiveCircuitHandler activeCircuitHandler)
+    {
+      if (activeCircuitHandler.IsServerSideBlazor)
+      {
+        ActivatorUtilities.CreateInstance(provider, typeof(ApplicationContextManagerBlazor));
+      }
+      else
+      {
+        ActivatorUtilities.CreateInstance(provider, typeof(ApplicationContextManagerHttpContext));
+      }
+    }
+#else
+    /// <summary>
+    /// Creates an instance of the object.
+    /// </summary>
+    /// <param name="httpContextAccessor"></param>
     public ApplicationContextManager(IHttpContextAccessor httpContextAccessor)
     {
-      HttpContext = httpContextAccessor.HttpContext;
+      ActiveContextManager = new ApplicationContextManagerHttpContext(httpContextAccessor);
     }
-
-    /// <summary>
-    /// Gets the current HttpContext instance.
-    /// </summary>
-    protected virtual HttpContext HttpContext { get; private set; }
+#endif
 
     /// <summary>
     /// Gets a value indicating whether this
@@ -43,7 +58,7 @@ namespace Csla.AspNetCore
     /// </summary>
     public bool IsValid
     {
-      get { return HttpContext != null; }
+      get { return ActiveContextManager.IsValid; }
     }
 
     /// <summary>
@@ -51,13 +66,7 @@ namespace Csla.AspNetCore
     /// </summary>
     public System.Security.Principal.IPrincipal GetUser()
     {
-      var result = HttpContext?.User;
-      if (result == null)
-      {
-        result = new Csla.Security.CslaClaimsPrincipal();
-        SetUser(result);
-      }
-      return result;
+      return ActiveContextManager.GetUser();
     }
 
     /// <summary>
@@ -66,7 +75,7 @@ namespace Csla.AspNetCore
     /// <param name="principal">Principal object.</param>
     public void SetUser(System.Security.Principal.IPrincipal principal)
     {
-      HttpContext.User = (ClaimsPrincipal)principal;
+      ActiveContextManager.SetUser(principal);
     }
 
     /// <summary>
@@ -74,7 +83,7 @@ namespace Csla.AspNetCore
     /// </summary>
     public ContextDictionary GetLocalContext()
     {
-      return (ContextDictionary)HttpContext?.Items[_localContextName];
+      return ActiveContextManager.GetLocalContext();
     }
 
     /// <summary>
@@ -83,7 +92,7 @@ namespace Csla.AspNetCore
     /// <param name="localContext">Local context.</param>
     public void SetLocalContext(ContextDictionary localContext)
     {
-      HttpContext.Items[_localContextName] = localContext;
+      ActiveContextManager.SetLocalContext(localContext);
     }
 
     /// <summary>
@@ -92,7 +101,7 @@ namespace Csla.AspNetCore
     /// <param name="executionLocation"></param>
     public ContextDictionary GetClientContext(ApplicationContext.ExecutionLocations executionLocation)
     {
-      return (ContextDictionary)HttpContext?.Items[_clientContextName];
+      return ActiveContextManager.GetClientContext(executionLocation);
     }
 
     /// <summary>
@@ -102,10 +111,10 @@ namespace Csla.AspNetCore
     /// <param name="executionLocation"></param>
     public void SetClientContext(ContextDictionary clientContext, ApplicationContext.ExecutionLocations executionLocation)
     {
-      HttpContext.Items[_clientContextName] = clientContext;
+      ActiveContextManager.SetClientContext(clientContext, executionLocation);
     }
 
-    private const string _applicationContextName = "Csla.ApplicationContext";
+    private bool disposedValue;
 
     /// <summary>
     /// Gets or sets a reference to the current ApplicationContext.
@@ -114,12 +123,40 @@ namespace Csla.AspNetCore
     {
       get
       {
-        return (ApplicationContext)HttpContext?.Items[_applicationContextName];
+        return ActiveContextManager.ApplicationContext;
       }
       set
       {
-        HttpContext.Items[_applicationContextName] = value;
+        ActiveContextManager.ApplicationContext = value;
       }
+    }
+
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="disposing"></param>
+    protected virtual void Dispose(bool disposing)
+    {
+      if (!disposedValue)
+      {
+        if (disposing)
+        {
+          if (ActiveContextManager is IDisposable disposable)
+            disposable.Dispose();
+        }
+
+        disposedValue = true;
+      }
+    }
+
+    /// <summary>
+    /// Dispose this instance.
+    /// </summary>
+    public void Dispose()
+    {
+      // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+      Dispose(disposing: true);
+      GC.SuppressFinalize(this);
     }
   }
 }
--- a/Source/Csla.AspNetCore/ApplicationContextManager.cs
+++ b/Source/Csla.AspNetCore/ApplicationContextManager.cs
@@ -28,10 +28,10 @@ namespace Csla.AspNetCore
     /// Creates an instance of the object.
     /// </summary>
     /// <param name="provider"></param>
-    /// <param name="activeCircuitHandler"></param>
-    public ApplicationContextManager(IServiceProvider provider, ActiveCircuitHandler activeCircuitHandler)
+    /// <param name="activeCircuitState"></param>
+    public ApplicationContextManager(IServiceProvider provider, ActiveCircuitState activeCircuitState)
     {
-      if (activeCircuitHandler.IsServerSideBlazor)
+      if (activeCircuitState.CircuitExists)
       {
         ActivatorUtilities.CreateInstance(provider, typeof(ApplicationContextManagerBlazor));
       }
--- a/Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
+++ b/Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
@@ -30,6 +30,8 @@ namespace Csla.AspNetCore
     public ApplicationContextManagerHttpContext(IHttpContextAccessor httpContextAccessor)
     {
       HttpContext = httpContextAccessor.HttpContext;
+      if (HttpContext == null)
+        throw new NullReferenceException(nameof(HttpContext));
     }
 
     /// <summary>
--- a/Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
+++ b/Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
@@ -1,0 +1,126 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ApplicationContextManager.cs" company="Marimer LLC">
+//     Copyright (c) Marimer LLC. All rights reserved.
+//     Website: https://cslanet.com
+// </copyright>
+// <summary>Application context manager that uses HttpContextAccessor</summary>
+//-----------------------------------------------------------------------
+using Csla.Core;
+using Microsoft.AspNetCore.Http;
+using System;
+using System.Security.Claims;
+
+namespace Csla.AspNetCore
+{
+  /// <summary>
+  /// Application context manager that uses HttpContextAccessor when 
+  /// resolving HttpContext to store context values.
+  /// </summary>
+  public class ApplicationContextManagerHttpContext : IContextManager
+  {
+
+    private const string _localContextName = "Csla.LocalContext";
+    private const string _clientContextName = "Csla.ClientContext";
+
+    /// <summary>
+    /// Creates an instance of the object, initializing it
+    /// with the required IServiceProvider.
+    /// </summary>
+    /// <param name="httpContextAccessor">HttpContext accessor</param>
+    public ApplicationContextManagerHttpContext(IHttpContextAccessor httpContextAccessor)
+    {
+      HttpContext = httpContextAccessor.HttpContext;
+    }
+
+    /// <summary>
+    /// Gets the current HttpContext instance.
+    /// </summary>
+    protected virtual HttpContext HttpContext { get; private set; }
+
+    /// <summary>
+    /// Gets a value indicating whether this
+    /// context manager is valid for use in
+    /// the current environment.
+    /// </summary>
+    public bool IsValid
+    {
+      get { return HttpContext != null; }
+    }
+
+    /// <summary>
+    /// Gets the current principal.
+    /// </summary>
+    public System.Security.Principal.IPrincipal GetUser()
+    {
+      var result = HttpContext?.User;
+      if (result == null)
+      {
+        result = new Csla.Security.CslaClaimsPrincipal();
+        SetUser(result);
+      }
+      return result;
+    }
+
+    /// <summary>
+    /// Sets the current principal.
+    /// </summary>
+    /// <param name="principal">Principal object.</param>
+    public void SetUser(System.Security.Principal.IPrincipal principal)
+    {
+      HttpContext.User = (ClaimsPrincipal)principal;
+    }
+
+    /// <summary>
+    /// Gets the local context.
+    /// </summary>
+    public ContextDictionary GetLocalContext()
+    {
+      return (ContextDictionary)HttpContext?.Items[_localContextName];
+    }
+
+    /// <summary>
+    /// Sets the local context.
+    /// </summary>
+    /// <param name="localContext">Local context.</param>
+    public void SetLocalContext(ContextDictionary localContext)
+    {
+      HttpContext.Items[_localContextName] = localContext;
+    }
+
+    /// <summary>
+    /// Gets the client context.
+    /// </summary>
+    /// <param name="executionLocation"></param>
+    public ContextDictionary GetClientContext(ApplicationContext.ExecutionLocations executionLocation)
+    {
+      return (ContextDictionary)HttpContext?.Items[_clientContextName];
+    }
+
+    /// <summary>
+    /// Sets the client context.
+    /// </summary>
+    /// <param name="clientContext">Client context.</param>
+    /// <param name="executionLocation"></param>
+    public void SetClientContext(ContextDictionary clientContext, ApplicationContext.ExecutionLocations executionLocation)
+    {
+      HttpContext.Items[_clientContextName] = clientContext;
+    }
+
+    private const string _applicationContextName = "Csla.ApplicationContext";
+
+    /// <summary>
+    /// Gets or sets a reference to the current ApplicationContext.
+    /// </summary>
+    public virtual ApplicationContext ApplicationContext
+    {
+      get
+      {
+        return (ApplicationContext)HttpContext?.Items[_applicationContextName];
+      }
+      set
+      {
+        HttpContext.Items[_applicationContextName] = value;
+      }
+    }
+  }
+}
--- a/Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
+++ b/Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
@@ -21,7 +21,7 @@ namespace Csla.AspNetCore.Blazor
     /// Gets a value indicating whether the current scope
     /// is running in server-side Blazor.
     /// </summary>
-    public bool IsServerSideBlazor { get; private set; }
+    public bool IsServerSideBlazor { get; private set; } = false;
 
     /// <summary>
     /// Handler for the OnCircuitOpenedAsync call
--- a/Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
+++ b/Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
@@ -18,10 +18,15 @@ namespace Csla.AspNetCore.Blazor
   public class ActiveCircuitHandler : CircuitHandler
   {
     /// <summary>
-    /// Gets a value indicating whether the current scope
-    /// is running in server-side Blazor.
+    /// Creates an instance of the type.
     /// </summary>
-    public bool IsServerSideBlazor { get; private set; } = false;
+    /// <param name="activeCircuitState"></param>
+    public ActiveCircuitHandler(ActiveCircuitState activeCircuitState)
+    {
+      ActiveCircuitState = activeCircuitState;
+    }
+
+    private ActiveCircuitState ActiveCircuitState { get; set; }
 
     /// <summary>
     /// Handler for the OnCircuitOpenedAsync call
@@ -30,7 +35,7 @@ namespace Csla.AspNetCore.Blazor
     /// <param name="cancellationToken">The cancellation token provided by the runtime</param>
     public override Task OnCircuitOpenedAsync(Circuit circuit, CancellationToken cancellationToken)
     {
-      IsServerSideBlazor = true;
+      ActiveCircuitState.CircuitExists = true;
       return base.OnCircuitOpenedAsync(circuit, cancellationToken);
     }
   }
--- a/Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
+++ b/Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
@@ -1,0 +1,38 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ActiveCircuitHandler.cs" company="Marimer LLC">
+//     Copyright (c) Marimer LLC. All rights reserved.
+//     Website: https://cslanet.com
+// </copyright>
+// <summary>Circuit handler indicating if code in server-side Blazor</summary>
+//-----------------------------------------------------------------------
+#if NET5_0_OR_GREATER
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Components.Server.Circuits;
+
+namespace Csla.AspNetCore.Blazor
+{
+  /// <summary>
+  /// Circuit handler indicating if code in server-side Blazor.
+  /// </summary>
+  public class ActiveCircuitHandler : CircuitHandler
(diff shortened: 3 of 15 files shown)


Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
```diff
@@ -0,0 +1,126 @@
...
+      return result;
+    }
+
+    /// <summary>
+    /// Sets the current principal.
+    /// </summary>
+    /// <param name="principal">Principal object.</param>
+    public void SetUser(System.Security.Principal.IPrincipal principal)
+    {
+      HttpContext.User = (ClaimsPrincipal)principal;
```
TheCakeMonster, line 70: This doesn't handle the possibility of a null HttpContext like the other code. I can't decide if that would be relevant by this point. Would this already have crashed before that? I don't _think_ so, so maybe we should handle the possibility?

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
```diff
@@ -0,0 +1,126 @@
...
+      return (ContextDictionary)HttpContext?.Items[_localContextName];
+    }
+
+    /// <summary>
+    /// Sets the local context.
+    /// </summary>
+    /// <param name="localContext">Local context.</param>
+    public void SetLocalContext(ContextDictionary localContext)
+    {
+      HttpContext.Items[_localContextName] = localContext;
```
TheCakeMonster, line 87: No null handling here either, should there be?

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /// <summary>
+    /// Sets the client context.
+    /// </summary>
+    /// <param name="clientContext">Client context.</param>
+    /// <param name="executionLocation"></param>
+    public void SetClientContext(ContextDictionary clientContext, ApplicationContext.ExecutionLocations executionLocation)
+    {
+      HttpContext.Items[_clientContextName] = clientContext;
```
TheCakeMonster, line 106: No null handling here again, should there be? I'm starting to think that you have ignored the possibility on purpose.

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// </summary>
+    public virtual ApplicationContext ApplicationContext
+    {
+      get
+      {
+        return (ApplicationContext)HttpContext?.Items[_applicationContextName];
+      }
+      set
+      {
+        HttpContext.Items[_applicationContextName] = value;
```
TheCakeMonster, line 122: Yeah, OK, I'm getting doubtful now! I guess it's inconsistent, let's put it like that.

Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+  /// <summary>
+  /// Circuit handler indicating if code in server-side Blazor.
+  /// </summary>
+  public class ActiveCircuitHandler : CircuitHandler
+  {
+    /// <summary>
+    /// Gets a value indicating whether the current scope
+    /// is running in server-side Blazor.
+    /// </summary>
+    public bool IsServerSideBlazor { get; private set; }
```
TheCakeMonster, line 24: Personally I might be tempted to actually explicitly set the property to false here. I know that the default is false, so the difference at the point of execution is nothing, but I tend it put it in to explicitly indicate that I intend for it to remain unset/be unmodified in certain circumstances, just for readability.

Source/Csla.AspNetCore/ApplicationContextManager.cs
```diff
@@ -3,38 +3,53 @@
...
     /// <summary>
-    /// Creates an instance of the object, initializing it
-    /// with the required IServiceProvider.
+    /// Creates an instance of the object.
     /// </summary>
-    /// <param name="httpContextAccessor">HttpContext accessor</param>
-    public ApplicationContextManager(IHttpContextAccessor httpContextAccessor)
+    /// <param name="provider"></param>
+    /// <param name="activeCircuitHandler"></param>
+    public ApplicationContextManager(IServiceProvider provider, ActiveCircuitHandler activeCircuitHandler)
```
TheCakeMonster, line 32: I'm not sure this is a great idea, but I'll mention it anyway ...!

I think there might be an opportunity to avoid the need to duplicate the Blazor context manager by not depending on the circuit handler here. If we instead depended on a separate class that the circuit handler is able to influence, the context manager itself is independent of server-side Blazor. That change might allow the circuit handler to live in Csla.AspNetCore, but the hybrid context manager to live in Csla.Blazor.

The only problem with this is that the type on which this class depends would have to live somewhere that is available in both flavours of Blazor, but also accessible from Csla.AspNetCore. Does that mean it would have to go in the core Csla assembly? It doesn't really feel right for such a class to go in the root Csla assembly.

Source/Csla.AspNetCore/Csla.AspNetCore.csproj
```diff
@@ -28,10 +28,12 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net5.0'">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Microsoft.AspNetCore.Components.Authorization" Version="5.0.11" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Microsoft.AspNetCore.Components.Authorization" Version="5.0.11" />
```
TheCakeMonster, line 36: Presumably this would ideally be 6.0.1, but that would require an update to all of the related packages for net6.0 throughout, which you might be delaying on purpose?

TheCakeMonster (commented):
Yes, I think that will work - but for one caveat. I didn't see the circuit handler for SSB registered anywhere. Sadly, the circuit handler needs to be registered with DI for it to be used. I think that needs adding to the AddServerSideBlazor method in Csla.Blazor/ConfigurationExtensions.cs.

I have suggested an alternative - disconnecting the hybrid context manager from SSB using a separate class that the circuit handler can influence, when it is in play. That would allow the circuit handler code to exist in Csla.AspNetCore, but the hybrid context manager to exist in Csla.Blazor, which I think would remove the need for the duplication of the Blazor context manager into two assemblies.

For that to work, we'd have to override the registered context manager when support for SSB is enabled - which admittedly is a little messier in that regard than what you have done, because the one chosen is based solely on the code in the central RegisterContextManager selection method in the core assembly.

Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
```diff
@@ -0,0 +1,38 @@
+﻿//-----------------------------------------------------------------------
```
swegele, line 1: Two CircuitManagers cannot be registered...so I'm wondering what happens when I register my own custom circuit handler, won't that step on this guys' toes?

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -0,0 +1,166 @@
...
+    {
+      AuthenticationStateProvider = authenticationStateProvider;
+      CurrentPrincipal = UnauthenticatedPrincipal;
+      AuthenticationStateProvider.AuthenticationStateChanged += AuthenticationStateProvider_AuthenticationStateChanged;
+      InitializeUser();
+    }
+
+    private void InitializeUser()
+    {
+      AuthenticationStateProvider_AuthenticationStateChanged(AuthenticationStateProvider.GetAuthenticationStateAsync());
```
swegele, line 55: I'm still thinking this line will blow up with that GetState called before SetState error when the instigating code is started in a HostedService?

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
rockfordlhotka: "Handle" would mean throwing an exception. Which could be valid, as we could throw something more instructive than a generic null reference exception maybe.

However, this really is a null reference scenario.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /// <summary>
+    /// Attempts to set the current principal on the registered
+    /// IHostEnvironmentAuthenticationStateProvider service.
+    /// </summary>
+    /// <param name="principal">Principal object.</param>
+    public virtual void SetUser(IPrincipal principal)
+    {
+      CurrentPrincipal = principal;
```
swegele, line 94: Could be a dumb suggestion...but I found that your LocalProxy was calling this so often as I jumped around on the server side of the dataportal (many child objects) and so I put a check in here to see if the CurrentPrincipal == principal then skip the whole SetAuthenticationState which seemed to me to just cause needless consumption of compute resources.

Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
TheCakeMonster: My understanding - or maybe assumption - was that you could have as many circuit handlers as you want. I think it probably uses the same IEnumerable<T> feature of DI as we use in Csla 6 for interceptors. Have you found some documentation that says otherwise?

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
TheCakeMonster: I'm pretty sure you were seeing that because you were outside of Blazor. The point of these changes is to avoid the use of this code in that scenario.

Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
swegele: In the demo I'm working on (haven't uploaded it yet), I have a "scoped" circuit handler that I'm making heavy use of.  Yesterday as I was adding more features, I was wondering "can I have a singleton circuit handler too?".  So I made one and registered it in Program.cs.  
All of a sudden all my "scoped" ones were null.  So I just assumed DI wouldn't allow two.  But maybe it had to do with the different scopes?

TheCakeMonster: No, a singleton circuit handler wouldn't work - or it would do something extremely strange, at the very least. At a guess there would be about a 15% chance that it would create a black hole that would engulf the Earth and grow to the size of the universe. Something like that. I wouldn't risk it ;-)

The concept of a circuit handler only makes sense within a circuit - which represents a scope. The circuit manager (or another, related class) creates a scope for the circuit as it is created. From there on, it's scopes all the way down.

TheCakeMonster:
I suppose the other thing that is worth calling out here is that the new hybrid context manager is being used for all ASP.NET scenarios, whether there is any Blazor in place or not. I think that's intentional in order to keep the registration of the context manager scoped to the RegisterContextManager method.

There is an alternative, which is to register the hybrid context manager from the AddServerSideBlazor method in Csla.Blazor/ConfigurationExtensions.cs. This does make it a bit more difficult for people to reason about which context manager will be used though, which is not ideal.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
swegele: I will double check whether the error I saw only happens from HostedService calls...and I will report back here.
In the meantime, I did get around all the issues by registering my own `BlazorApplicationContextManager` (a copy of Rocky's) with the 3 small tweaks I mentioned in [this discussion](https://github.com/MarimerLLC/csla/discussions/2687#discussioncomment-2051923) and the small fix a couple comments later.

Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
swegele: LOL - Superman 3

I agree.  I was surprised when I saw [this Microsoft article](https://github.com/MarimerLLC/csla/discussions/2687#discussioncomment-2051923) and decided to try it for myself...wondering "how the heck would this work?!"  Notice they register it as a singleton.  Perhaps the documentation person just made a mistake??

Source/Csla.AspNetCore/Csla.AspNetCore.csproj
rockfordlhotka: Yeah, there's a need to go through and check all the versions. dependabot is nice, but falls down when your projects target multiple versions of dotnet that need different versions of dependencies.

Source/Csla.AspNetCore/ApplicationContextManager.cs
rockfordlhotka: I struggled with that, and didn't want to add such a type to core Csla.dll. 

There is no duplicate Blazor context manager now though, because Csla.Blazor no longer contains a context manager.

There is a "duplicate" in that Csla.Blazor.WebAssembly now contains a context manager, because the aspnetcore one has access to that server-hosted authentication state provider interface, and the wasm one does not. That feels a little dirty to me, but maybe it is better to have SSB and wasm be cleanly separate in their implemenations.

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
rockfordlhotka: We have to throw some exception - nullreference seems valid?

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
rockfordlhotka: We have to throw some exception - nullreference seems valid?

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
swegele: Nope as I thought...the same unhandled error happens 
```cs        
     [2022-02-01T20:57:59.663Z] Error: System.InvalidOperationException: GetAuthenticationStateAsync was called before SetAuthenticationState.
 at Microsoft.AspNetCore.Components.Server.ServerAuthenticationStateProvider.GetAuthenticationStateAsync()
 at Csla.Blazor.ApplicationContextManager.InitializeUser()
```

simply from injecting the Csla.ApplicationContext into the Counter blazor page.
```cs
@inject ApplicationContext _appContext;
```

Project only references Csla.Blazor (not Csla.AspNet.Core).

Here is Program.cs
```cs
builder.Services.AddRazorPages();
builder.Services.AddServerSideBlazor();
builder.Services.AddHttpContextAccessor();

builder.Services.AddCsla(options =>
{
    options.AddServerSideBlazor();
});

```

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
TheCakeMonster: Yeah, I guess so. The only thing I don't like about NRE is that it often doesn't tell you _what_ was null. If we are able to add that extra info, then I agree that's the best we can do.

Source/Csla.AspNetCore/ApplicationContextManager.cs
swegele: Rocky I humbly remember here what you told me over 10+ years ago in person at a MS conference in San Fran...when I was struggling with duplicated code in different objects.  I was trying to figure out how to share Property Getters between read-only and writeable objects.  You said two things:
1.  "Behavior driven design" and you had me go buy David Wests book :-)
2.  "If your not using code generation then your an idiot"  You said it nicely :-)

The application context manager that serves WASM is different from SSB according to that sage wisdom....only because Microsoft is doing different things under the hood with them (their version of ServerAuthenticationStateProvider)

Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
TheCakeMonster: Hmm, yeah, interesting. You linked back to us, not the article, but I think I see the one you mean.

Here is CircuitHost, it accepts an array of circuit handlers into its constructor and iterates through all of them in the appropriate places, so there should be no reason why there can't be lots.

https://github.com/dotnet/aspnetcore/blob/main/src/Components/Server/src/Circuits/CircuitHost.cs

I was expecting to see `IEnumerable<CircuitHandler>` but I guess an array does the same thing. I know they are somewhat interchangeable in places.

swegele: Sorry...fixed the link above.

TheCakeMonster: Thanks. Yeah, it was the one I found.

It makes sense that you could use a handler across circuits I suppose, as long as you don't expect it to do anything specific to any single one.

I wonder if you might have got caught out by captive dependencies somehow in what you were doing. It's just a guess, but it might have explained what you were seeing. https://blog.ploeh.dk/2014/06/02/captive-dependency/

Anyway, it seems .NET allows for there being multiple, which is what we were concerned about.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
swegele: @rockfordlhotka and @TheCakeMonster - I've updated my [demo](https://github.com/swegele/Csla6BlazorSS) so it takes < 5 minutes to see this error.  

Totally removed HostedService.  Nothing funky.
Follow instructions on readme.md

1.  Run app, go to Counter page and see unhandled error (F12)
2. Stop app
3. Comment out injection line of my business object factory on Counter.razor
3. Run app again and see everything works

I'm not even calling the business objects...simply injecting one factory.  So weird.

Source/Csla.AspNetCore/Blazor/ActiveCircuitHandler.cs
rockfordlhotka: Thank you both for the deep thinking here.

Just so I'm clear: what you discovered is that we can register this circuit handler, and people can register other ones too, and they'll be honored (if they are scoped the same)?

(I wonder if the confusion is that in Blazor wasm Singleton and Scoped have the same effect in most cases, but in SSB they always have different effects? Maybe the sample author was confused about this?)

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
rockfordlhotka: @swegele what is (in summary) the solution?

rockfordlhotka:
The Csla.Blazor project has no reference to the Csla.AspNetCore project, so I was unable to (easily) register the type from there.

I think it could be done by dynamically loading the type (`Type.GetType()`), and maybe that is a better solution?

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
swegele: It's a workaround for [this issue](https://github.com/dotnet/aspnetcore/issues/28684).  I openly admit I do not understand why it is happening in that very simple and clean demo.  If we can't figure that out, some order of operations thing I'm guessing, then we need a workaround (at least until Microsoft fixes that).

**Summary of workaround for a simple/default BSS contextmanager (not hybrid)** 

1. Make a version of an `ApplicationContextManager` for Blazor SS (separate from WASM) so we can count on the `AuthenticationStateProvider` implementing `IHostEnvironmentAuthenticationStateProvider`
2. Trap the error during `InitializeUser()` and call `SetAuthenticationState`.  That is what the error instructs us to do  
3. Also call the `SafeSetCurrentUser` method from SetUser

Credit to @TheCakeMonster as this is a slightly tweaked version of his code last week...so far it fixes everything in all scenarios for BSS that I've done so far.

```cs
private void InitializeUser()
{
        try
        {
            var getUserTask = AuthenticationStateProvider.GetAuthenticationStateAsync();
            AuthenticationStateProvider_AuthenticationStateChanged(getUserTask);
        }
        catch (InvalidOperationException ioex)
        {
            //Test for this specific error
            string message = ioex.Message;
            if (message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync))
                && message.Contains(nameof(IHostEnvironmentAuthenticationStateProvider.SetAuthenticationState)))
            {
                //error tells us no one has called set yet...so let's call it
                SafeSetCurrentUser(AuthenticationStateProvider, UnauthenticatedPrincipal);
            }
            else
            {
               thow;
            }
        }
    }
}

private void SafeSetCurrentUser(IHostEnvironmentAuthenticationStateProvider hostEnvironmentAuthProvider, ClaimsPrincipal principal)
{
    var task = new Task<AuthenticationState>(() => new AuthenticationState(principal));
    hostEnvironmentAuthProvider.SetAuthenticationState(Task.FromResult(authState));
}

public virtual void SetUser(IPrincipal principal)
{   
    if (!CurrentPrincipal.Equals(principal))
    {
        CurrentPrincipal = principal;
        if (CurrentPrincipal is ClaimsPrincipal)
        {
            SafeSetCurrentUser(AuthenticationStateProvider, (ClaimsPrincipal)CurrentPrincipal);
        }
        else
        {
            //??Throw some exception here??
        }
    }
}

```

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
rockfordlhotka: I'm now throwing a "good" exception in the constructor if the value is null.

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
```diff
@@ -30,6 +30,8 @@ public class ApplicationContextManagerHttpContext : IContextManager
     public ApplicationContextManagerHttpContext(IHttpContextAccessor httpContextAccessor)
     {
       HttpContext = httpContextAccessor.HttpContext;
+      if (HttpContext == null)
```
rockfordlhotka, line 33: @TheCakeMonster this should fix the null reference issue in the whole class?

swegele:
I will do a test tomorrow to see if I can prove multiple circuit handlers
(all scoped) and see if it works. Will let ya know.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -52,18 +52,44 @@ public ApplicationContextManagerBlazor(AuthenticationStateProvider authenticatio
 
     private void InitializeUser()
     {
-      AuthenticationStateProvider_AuthenticationStateChanged(AuthenticationStateProvider.GetAuthenticationStateAsync());
+      Task<AuthenticationState> task = default;
```
rockfordlhotka, line 55: @swegele I think this captures the fix to the issue of setting the user if it hasn't already been set?

Source/Csla.AspNetCore/ConfigurationExtensions.cs
```diff
@@ -36,6 +40,9 @@ public static CslaOptions AddAspNetCore(this CslaOptions config, Action<AspNetCo
     {
       var localOptions = new AspNetCoreConfigurationOptions();
       options?.Invoke(localOptions);
+#if NET5_0_OR_GREATER
+      config.Services.AddScoped(typeof(CircuitHandler), typeof(ActiveCircuitHandler));
```
rockfordlhotka, line 44: @TheCakeMonster the circuit handler is now being registered.

@swegele I read a few things about this. It does appear that if someone registered a singleton CircuitHandler that it would brake CSLA. It appears to be a limitation of .NET DI, so I don't know how to fix the issue.

Were someone to choose to implement a singleton handler, they'd have to figure out a way to implement a singleton solution to the CSLA requirement as well - and maybe that's something we'll solve if/when someone actually does such a thing :)

swegele: Well before bed, I hopped on PC and ran a test.  I made an exact copy of my "scoped" circuit handler and named it with 2.  Then I registered it and put a breakpoint in OnCircuitOpenedAsync in both.

When debugging **they both get called** (interesting)...but it seems that DI will only inject the **LAST ONE** that is registered in Program.cs.  So when you have this in your page...`@inject CircuitHandler _circuitHandler;` you get the last one.  

So when I have them like this - then `UserScopedCircuitHandler` ends up `null` when I try to cast it.
```cs
builder.Services.AddScoped<CircuitHandler, UserScopedCircuitHandler>();
builder.Services.AddScoped<CircuitHandler, UserScopedCircuitHandler2>();
```


But when I reverse it, then I do get the cast just fine.
```cs
builder.Services.AddScoped<CircuitHandler, UserScopedCircuitHandler2>();
builder.Services.AddScoped<CircuitHandler, UserScopedCircuitHandler>();
```

So I'm not sure how you would ever end up injecting yours in the framework if mine is registered after it.

rockfordlhotka: So there can be only one circuit handler registered period - like most services. This, then, is fine with me. If someone needs a custom circuit handler, they'll have to incorporate the CSLA functionality into their implementation as well.

rockfordlhotka: Oh wait. I just read what you said more closely.

I think @TheCakeMonster solved this already - if you want to have more than one injected you need to request an array of the type rather than just the type.

```c#
// constructor
public MyType(CircuitHandler[] handlerList)
{
}
```

swegele: One additional thought, since all instances get called but only one gets served up via DI...you could have your circuit handler inject to another service (e.g. `CircuitWatcher` ).  That could be the service that you use internally which could be the messenger and safe to use.  
Then us user's wouldn't have to implement anything in our circuit handlers.

swegele: When I try
`public MyType(CircuitHandler[] handlerList)`

I get this error:
`Unable to resolve service for type 'Microsoft.AspNetCore.Components.Server.Circuits.CircuitHandler[]'`

swegele: Viola!  Just tried my idea and I think it will work for you.  

Given this:
```cs
builder.Services.AddScoped<CircuitHandler, CslaScopedCH>();
builder.Services.AddScoped<CslaCHMessenger>();
builder.Services.AddScoped<CircuitHandler, MyScopedCH>();
```

and this additional service:
```cs
public class CslaCHMessenger
{
    public string CircuitIdFromCslaCHMessenger { get; set; }

    public CslaCHMessenger()
    {
    }
}
```

Anywhere in my app where I call `@inject CircuitHandler _circuitHandler;` I get a `MyScopedCH`.  Good.

But your `CslaScopedCH` still gets called just when the `OnCircuitOpenedAsync` event fires.

So in `CslaScopedCH` constructor you can inject `CslaCHMessenger`.  Just as a test I set the `CircuitIdFromCslaCHMessenger`  property of it.  In the Counter.razor page I inject `CslaCHMessenger` and I see it's value is set.

You could do this and have the messenger class obtain and carry whatever values out of your circuit handler that you need to test.  And you can always count on it working.

HTH

TheCakeMonster: Loads of questions and limited time, so I will try to summarise.

1. Yes, there can be multiple circuit handlers registered. The events on all of them will be raised by the underlying infrastructure.
2. If you ask for `IEnumerable<CircuitHandler>` to be injected, you get an enumerable of all of them. We know this works, as this is what we do with Interceptors (and is proven to work by the results of the Csla.Tests project.)
3. It looks like `CircuitHost` gets passed an array of `CircuitHandler` and my guess was that this was coming from DI. However, I don't recognise the syntax being used, so I can't be sure that will definitely work. Maybe CircuitHost isn't created by DI; I remember researching it and writing it all down on this repo somewhere, but I don't know exactly where TBH.
3. If you ask for a type of CircuitHandler to be injected, you get the last one registered, whatever that was. That makes it very hit and miss what you will get, so I wouldn't try this.
4. If you ask for your own derived type by name, I'm not sure what happens. At a guess it would need to be registered twice, and it might not do what you expect.

I know that my demo code worked, so it's worth checking over that again. I used a separate class to carry the data from one place to another; at a glance this also seems to be what @swegele is suggesting. Somehow that feels more correct anyway, to break the dependencies apart, but I'm not sure that I can explain _exactly_ why in the limited amount of time I have at the moment.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
TheCakeMonster: I continue to be surprised that this is a problem. I do wonder if it's a peculiarity of something else, such as the host page of the Blazor island, or a left over bit of an experiment that is still in place and forgotten. However, I haven't had a chance to check @swegele 's demo code, so I can't be sure.

I'd be tempted to create a new SSB solution and try the code without this change, to see if it's really still a problem.

Like I've said before, SSB never exists in isolation, as something has to host it. I'm wondering if the tyre kicking being done doesn't take this into account, but I haven't managed to make time to check myself, sorry.

swegele: That should do it - yes

Source/Csla.AspNetCore/ConfigurationExtensions.cs
rockfordlhotka: It think @TheCakeMonster's solution with the messenger class is correct. I was trying to simplify things by removing it, but clearly it is necessary (or is the simplest solution).

I'll make that change to the codebase and see what we all think.

Source/Csla.AspNetCore/Blazor/ActiveCircuitState.cs
```diff
@@ -0,0 +1,25 @@
+﻿//-----------------------------------------------------------------------
```
rockfordlhotka, line 1: Thank you for all the testing @swegele - hopefully this latest commit fits into what you discovered?

swegele: Wunder Bar!  This little trick seems like a nice pattern to have in the back pocket...thanks to you both.

rockfordlhotka: I was thinking the same - a nice pattern to keep in mind for the future.

PR #2773, opened 2022-02-15
Bug fixes 

### LINKED ISSUE #2770: Fix IDataPortalFactory DI error
### Discussed in https://github.com/MarimerLLC/csla/discussions/2769

<sup>Originally posted by **swegele** February 14, 2022</sup>
Using your latest preview from Nuget...when I try to use the IDataPortalFactory in my business class factory like so:
```cs
public class BetaInfoListFactory : DemoReadOnlyListBaseFactory<BetaInfoList, BetaInfo>, IBusinessObjectListFactory
{
    public BetaInfoListFactory(IDataPortalFactory portal, ApplicationContext applicationContext)
    {
        Portal = portal.GetPortal<BetaInfoList>();

```

I get this error:
```System.InvalidOperationException: A suitable constructor for type 'Csla.IDataPortal`1[BusinessLayer.BetaInfoList]' could not be located. Ensure the type is concrete and all parameters of a public constructor are either registered as services or passed as arguments. Also ensure no extraneous arguments are provided.```

Based on some SO answers...I changed this (line 388) in your ApplicationContext.cs:
```cs
result = ActivatorUtilities.CreateInstance(CurrentServiceProvider, objectType, parameters);
```

to be this:
```cs
result = CurrentServiceProvider.GetService(objectType);
```

Then it works.  Figured I would share with you my findings.  

**EDIT:**  

@rockfordlhotka  Confirmed I get the same error in your BlazorExample, by changing as follows:

```cs
@inject Csla.IDataPortalFactory portalFactory
```

and below OnParametersSetAsync() to this:

```cs
var personListPortal = portalFactory.GetPortal<PersonList>();
await vm.RefreshAsync(() => personListPortal.FetchAsync());
```

</div>

### LINKED ISSUE #2771: 'Csla.Blazor.WebAssembly.dll' shouldn't be part of Csla.Blazor package
### Discussed in https://github.com/MarimerLLC/csla/discussions/2767

<sup>Originally posted by **swegele** February 14, 2022</sup>
1.  Create a new Blazor SS project in VS 2022 (no auth)
2.  Reference these two nuget packages (Csla.Blazor and Csla.AspNetCore)  
    choose latest version '6.0.0-R22020902'
3.  Build the solution
Notice the "Csla.Blazor.WebAssembly.dll" in the bin directory.  Why is it there?

That WebAssembly.dll being there by default is causing Csla to use the `Csla.Blazor.WebAssembly.ApplicationContextManager` 

If I delete that dll, then Csla properly uses the `Csla.AspNetCore.ApplicationContextManager`.  Every code-change/build puts that dll back again.

</div>

### LINKED ISSUE #2775: LocalProxy should use a consistent app context type
Now that we are past all that, I see two issues related to the `Csla.AspNetCore.ApplicationContextManager`:

1.  During a HostedService instantiation of business layer, the code errantly selects the `ApplicationContextManagerHttpContext` which simply throws a NULL error for httpcontext.  This is because a HostedService call happens with no circuit and no httpcontext.  IMO in that scenario the code should select the `ApplicationContextManagerBlazor.`  This works great if you edit as follows:

```cs
if (activeCircuitState.CircuitExists || httpContextAccessor.HttpContext is null)
{
  ActiveContextManager = (IContextManager)ActivatorUtilities.CreateInstance(provider, typeof(ApplicationContextManagerBlazor));
}
else
{
  ActiveContextManager = (IContextManager)ActivatorUtilities.CreateInstance(provider, typeof(ApplicationContextManagerHttpContext));
}
```

2. Now, for a normal instantiation of a business layer call (which is inside a circuit), I noticed this constructor gets called twice.  The first time , from injection on a BSS razor page/component, it is called and the CircuitExists property is true so I get an instance of `ApplicationContextManagerBlazor` as expected.  
But then on logical server side of the dataportal this gets called again because of the new scope created.  This time CircuitExists property is false, so I get an instance of `ApplicationContextManagerHttpContext`.  That seems wrong to get one on client side of dataportal and a different one on logical server side.  I'm using LocalProxy of course for my testing.

_Originally posted by @swegele in https://github.com/MarimerLLC/csla/issues/2773#issuecomment-1039934424_

### LINKED ISSUE #2791: No GetDataPortal method in IRuleContext and IAuthorizationContext
**Describe the bug**
`IDataPortal<T> GetDataPortal<T>()` method not declared in IRuleContext and IAuthorizationContext interfaces

**Version and Platform**
CSLA version: 6.0.000 Preview 4
OS: Windows, Linux, iOS, Android, etc.
Platform: WinForms, WPF, ASP.NET Core, MVC, Xamarin, etc.

**Additional context**
Issue #2682 added GetDataPortal methods to rule context and authorization context, but not in interface

### LINKED ISSUE #2793: LocalProxy needs to property create (or not) DI scope in aspnetcore
In an aspnetcore project that hosts both server-side Blazor _and also_ MVC/Razor Pages content, and which uses the local data portal, it is necessary for `LocalProxy` to properly switch between creating a new DI scope (for server-side Blazor), and not (for aspnetcore).

Right now such a host will use the server-side Blazor configuration: always creating a new scope for each data portal request. This _works_, but isn't optimal because this is unnecessary work for aspnetcore requests.

The answer is probably to stop setting the value as a static configuration bool, and instead get the value from a service. Such a service can work similarly to the way the aspnetcore application context manager works to detect whether to use the Blazor or Http based context manager. In this case, the service can return true/false to indicate whether to create a new DI scope in LocalProxy.

### LINKED ISSUE #2794: aspnetcore ApplicationContextManager doesn't pick correct mgr type
In `Csla.AspNetCore.ApplicationContextManager` to keep a HostedService in a BSS app from blowing up immediately looking for `HttpContext`.

```c#
if (activeCircuitState.CircuitExists || httpContextAccessor.HttpContext is null)
```

cc @swegele

rockfordlhotka:
Closes #2770 
Closes #2771 
Closes #2791 
Closes #2775 
Closes #2793 
Closes #2794

### DIFF
--- a/NuGet/Definition/Csla.Blazor.NuSpec
+++ b/NuGet/Definition/Csla.Blazor.NuSpec
@@ -37,11 +37,17 @@
   </metadata>
   <files>
     <!-- NetStandard Assembly -->
-    <file src="..\..\bin\Release\netstandard2.1\**\Csla.Blazor.*" target="lib\netstandard2.1" />
+    <file src="..\..\bin\Release\netstandard2.1\**\Csla.Blazor.dll" target="lib\netstandard2.1" />
+    <file src="..\..\bin\Release\netstandard2.1\**\Csla.Blazor.pdb" target="lib\netstandard2.1" />
+    <file src="..\..\bin\Release\netstandard2.1\**\Csla.Blazor.xml" target="lib\netstandard2.1" />
     <!-- Net5.0 Assembly -->
-    <file src="..\..\bin\Release\net5.0\**\Csla.Blazor.*" target="lib\net5.0" />
+    <file src="..\..\bin\Release\net5.0\**\Csla.Blazor.dll" target="lib\net5.0" />
+    <file src="..\..\bin\Release\net5.0\**\Csla.Blazor.pdb" target="lib\net5.0" />
+    <file src="..\..\bin\Release\net5.0\**\Csla.Blazor.xml" target="lib\net5.0" />
     <!-- Net6.0 Assembly -->
-    <file src="..\..\bin\Release\net6.0\**\Csla.Blazor.*" target="lib\net6.0" />
+    <file src="..\..\bin\Release\net6.0\**\Csla.Blazor.dll" target="lib\net6.0" />
+    <file src="..\..\bin\Release\net6.0\**\Csla.Blazor.pdb" target="lib\net6.0" />
+    <file src="..\..\bin\Release\net6.0\**\Csla.Blazor.xml" target="lib\net6.0" />
     <!-- resources -->
     <file src="..\..\Support\Logos\csla.png" target="images\" />
   </files>
--- a/Source/Csla.AspNetCore/ApplicationContextManager.cs
+++ b/Source/Csla.AspNetCore/ApplicationContextManager.cs
@@ -62,6 +62,12 @@ namespace Csla.AspNetCore
     }
 
     /// <summary>
+    /// Gets a value indicating whether the current runtime
+    /// is stateful (e.g. WPF, Blazor, etc.)
+    /// </summary>
+    public bool IsStatefulRuntime => ActiveContextManager.IsStatefulRuntime;
+
+    /// <summary>
     /// Gets the current principal.
     /// </summary>
     public System.Security.Principal.IPrincipal GetUser()
--- a/Source/Csla.AspNetCore/ApplicationContextManager.cs
+++ b/Source/Csla.AspNetCore/ApplicationContextManager.cs
@@ -7,6 +7,7 @@
 //-----------------------------------------------------------------------
 using Csla.Core;
 using System;
+using Microsoft.AspNetCore.Http;
 #if NET5_0_OR_GREATER
 using Csla.AspNetCore.Blazor;
 using Microsoft.Extensions.DependencyInjection;
@@ -29,15 +30,20 @@ namespace Csla.AspNetCore
     /// </summary>
     /// <param name="provider"></param>
     /// <param name="activeCircuitState"></param>
-    public ApplicationContextManager(IServiceProvider provider, ActiveCircuitState activeCircuitState)
+    /// <param name="httpContextAccessor"></param>
+    public ApplicationContextManager(IServiceProvider provider, ActiveCircuitState activeCircuitState, IHttpContextAccessor httpContextAccessor)
     {
       if (activeCircuitState.CircuitExists)
       {
         ActiveContextManager = (IContextManager)ActivatorUtilities.CreateInstance(provider, typeof(ApplicationContextManagerBlazor));
       }
-      else
+      else if (httpContextAccessor.HttpContext is not null)
       {
         ActiveContextManager = (IContextManager)ActivatorUtilities.CreateInstance(provider, typeof(ApplicationContextManagerHttpContext));
+      }
+      else
+      {
+        ActiveContextManager = (IContextManager)ActivatorUtilities.CreateInstance(provider, typeof(ApplicationContextManagerAsyncLocal));
       }
     }
 #else
--- a/Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
+++ b/Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
@@ -50,6 +50,12 @@ namespace Csla.AspNetCore
     }
 
     /// <summary>
+    /// Gets a value indicating whether the current runtime
+    /// is stateful (e.g. WPF, Blazor, etc.)
+    /// </summary>
+    public bool IsStatefulRuntime => false;
+
+    /// <summary>
     /// Gets the current principal.
     /// </summary>
     public System.Security.Principal.IPrincipal GetUser()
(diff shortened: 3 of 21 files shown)


rockfordlhotka:
I think this fixes your latest reports @swegele ?

swegele:
Tested, and all those issues are fixed!  
 :-)  Hooray, seeing light at the end of the tunnel.

Now that we are past all that, I see two issues related to the `Csla.AspNetCore.ApplicationContextManager`:
@TheCakeMonster 
1.  During a HostedService instantiation of business layer, the code errantly selects the `ApplicationContextManagerHttpContext` which simply throws a NULL error for httpcontext.  This is because a HostedService call happens with no circuit and no httpcontext.  IMO in that scenario the code should select the `ApplicationContextManagerBlazor.`  This works great if you edit as follows:

```cs
if (activeCircuitState.CircuitExists || httpContextAccessor.HttpContext is null)
{
  ActiveContextManager = (IContextManager)ActivatorUtilities.CreateInstance(provider, typeof(ApplicationContextManagerBlazor));
}
else
{
  ActiveContextManager = (IContextManager)ActivatorUtilities.CreateInstance(provider, typeof(ApplicationContextManagerHttpContext));
}
```


2. Now, for a normal instantiation of a business layer call (which is inside a circuit), I noticed this constructor gets called twice.  The first time , from injection on a BSS razor page/component, it is called and the CircuitExists property is true so I get an instance of `ApplicationContextManagerBlazor` as expected.  
But then on logical server side of the dataportal this gets called again because of the new scope created.  This time CircuitExists property is false, so I get an instance of `ApplicationContextManagerHttpContext`.  That seems wrong to get one on client side of dataportal and a different one on logical server side.  I'm using LocalProxy of course for my testing.

swegele:
@rockfordlhotka feel free to put those last two bits into a separate issue if you want.  They are not showstoppers as I can make my own as a workaround.

rockfordlhotka:
I think for the number 1 case, it should not use the Blazor context manager, b/c that one relies on the authentication state provider - which won't be available right? It needs a context manager that stores everything (including the principal) as a private field.

swegele:
Just ran a test where my tweak makes the HostedService call the ApplicationContextManagerBlazor.  I set a breakpoint in the constructor and the AuthenticationStateProvider is NOT NULL.  

It is of the proper type `{Microsoft.AspNetCore.Components.Server.ServerAuthenticationStateProvider}`

rockfordlhotka:
OK, that's interesting. Did that authentication state provider give you the correct user identity?

swegele:
I'd have to test that...since, as of yet, I'm not logging in a user during that Hosted Service call, but rather just getting some readonly list utility data.  What do you suggest?

swegele:
@rockfordlhotka  OK it took me a minute to test this....it works.  

I did the following:
1. In the Hosted Service I got through DI the `var authStateProvider = sp.GetRequiredService<AuthenticationStateProvider>();`
2. Then I created a claims principal
3. Created a task `var task = Task.FromResult(new AuthenticationState(principal));`
4. Called `(authStateProvider as IHostEnvironmentAuthenticationStateProvider).SetAuthenticationState(task);`
5. Then proceeded to call my business object

On the logical server side of the local dataportal I checked `ApplicationContext.User` and it was set correctly.

TheCakeMonster:
@swegele I think you'd need to be very careful about use of a scoped service outside of a scope. You'll have to investigate exactly what happens in that scenario. I have heard it said that creating an instance of a scoped service from the root service provider (instead of from a service provider created by starting a scope) makes the created instance a singleton - and because of that, there is only a single instance.

I can understand this idea, but I'm not exactly clear what it means in fine detail. I'm not sure if that same, singleton instance would be provided in circumstances where the instantiation of a consuming class is done inside of a scope as well, or if the one singleton instance is only shared by code outside of any particular scope.

Sharing a single instance across all scopes is definitely not what you are after!

swegele:
@TheCakeMonster Thank you.  Your comment just gave me the epiphany that I needed.  I think you are saying it is EXPECTED that the scoped services ARE DIFFERENT on the logical server side of the Local Portal (because of the new scope option).  I've been struggling with that because it didn't seem right.  I know it seems basic, but honestly that is a new thought to me that I am trying to come to terms with.

Now I'm asking myself, why would I want a new scope - e.g. different instances of services/classes - when I'm simply using a Local Portal?  Why is the default to have the "logical" server-side be in it's own scope?  What purpose does that serve?  
I can turn that option off, but it begs the question why it was put there and default to true.

swegele:
@TheCakeMonster Regardless of the answer to the previous question (which I am still very curious about)...  

I feel like the `CircuitExists=true` on the client side of the local dataportal then turning into `CircuitExists=false` on the logical server side causes an unintended side effect of the `ApplicationContextManagerHttpContext` being selected when it shouldn't.  In reality, in BSS, the HttpContext should not be used as you found in that article...yes it's present but it should not to be used or counted on when doing things from a BSS circuit.  right?

swegele:
@rockfordlhotka - to keep this PR from spidering off into oblivion by my questions...I will attempt to summarize to expedite closure/progress, and to move questions to "Discussions"

Regarding the original 1 and 2 issues I raised above:

**Number 1** - is solved by adding the extra "if" clause: 
`if (activeCircuitState.CircuitExists || httpContextAccessor.HttpContext is null)`
This allows a HostedService to work without error because it uses the `ApplicationContextManagerBlazor`
After testing it looks like the AuthenticationStateProvider still works fine in that scenario.  

**Number 2** - is perhaps best discussed and if need-be enhanced on issue #2775 - and should be limited to the question of how to make it consistently pick the correct ContextManager (which I will comment on more there).

PR #2780, opened 2022-02-17
Pr/swegele/2773

swegele:
This is my suggested attempt to fix two remaining issues:

1. HostedService always throws error because HttpContext is null AND circuit is null.  So I added an if clause to make the Blazor app context manager get selected in this scenario

2. When a BSS razor page accesses a business object (in context of a circuit), the hybrid application context manager correctly selects the Blazor app context manager.  BUT on the logical server side in LocalProxy, a new scope is created, which causes the `CircuitExists` property to now be default/false.  Subsequently this causes the selection of the Http app context manager which is discouraged by microsoft when in BSS context.  Both sides of dataportal should get the Blazor app context manager in this scenario.

I struggled for hours to figure out a way to send a signal to the `Csla.AspNetCore.ApplicationContextManager` so it knows, on logical server side, to select the correct active context manager.  Chicken/egg scenario over and over.  The only way I could accomplish this was to bump the ActiveCircuitState service up into Csla.Core.Blazor.  Maybe not ideal but fairly minor change.  This allows the LocalProxy to fetch it from service provider before and after new scope in order to copy the boolean value across for CircuitExists property.  Maybe you can think of a better way to get that done.  But this is tested and works great.

### DIFF
--- a/NuGet/Definition/Csla.Blazor.NuSpec
+++ b/NuGet/Definition/Csla.Blazor.NuSpec
@@ -37,11 +37,17 @@
   </metadata>
   <files>
     <!-- NetStandard Assembly -->
-    <file src="..\..\bin\Release\netstandard2.1\**\Csla.Blazor.*" target="lib\netstandard2.1" />
+    <file src="..\..\bin\Release\netstandard2.1\**\Csla.Blazor.dll" target="lib\netstandard2.1" />
+    <file src="..\..\bin\Release\netstandard2.1\**\Csla.Blazor.pdb" target="lib\netstandard2.1" />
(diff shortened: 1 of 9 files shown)


swegele:
The fact that I had to move something Blazor related down into Csla.Core in order to get DI to fire up the correct ContextManager felt weird.  It "works" and it was "minimal" but it just seemed like something was off.  I wondered if that "something" might grow nastier over time. 

I believe @TheCakeMonster is right on his [comment](https://github.com/MarimerLLC/csla/issues/2775#issuecomment-1042785929) where he puts his finger on the source of the issue.  So feel free to toss this PR in the "swegele learning round file" :-)

swegele:
This PR is no longer relevant given other changes.

PR #2804, opened 2022-03-02
2798-localproxy-seantweak

swegele:
I used your main idea and tweaked it to:

- Make all scenarios work including calls to bl from (hosted service, cshtml pages, and razor pages);  
- HttpContextManager now knows it's NOT valid if a circuit exists or if a localproxy scope has been created;
- Simplified the runtimeinfo to only track if localproxy scope exists;
- AspNetCore and Blazor configurationextensions no longer fight with eachother;

I believe this covers all the scenarios...all the ones i've been testing anyway and it does seem a bit cleaner of a design.


swegele:
Well shoot, not sure what I did wrong here.  I meant to be able to create a PR to your rockfordlhotka/csla branch that you've got going.  I'll try again and ditch this.

Issue #2798, opened 2022-02-23
DI still giving wrong IContextManager within new scope in LocalProxy (IsStatefulRuntime=true)


swegele:
@rockfordlhotka  @TheCakeMonster 
All testing good in preview 6 except one last issue - it's baaaack :-(   

AFAIK - this issue only shows up when
1.  LocalProxy is being used
2.  A new scope is in play (e.g. for ApplicationContextManager instances with `IsStatefulRuntime=true`)

As you know, in `LocalProxy,` if in a new scope, we now manually create `ApplicationContextManagerAsyncLocal`.  
But since DI doesn't create it, DI doesn't yet have a scoped instance of `IContextManager` to give others.  So the next time (while in that scope) any code uses DI to get an `IContextManager`, a new instance is given for whatever `IContextManager` type is registered.  (In BSS this ends up being a new instance of `ApplicationContextManagerHttpContext`)

**BUG:**  So you  end up with LocalProxy having an  `ApplicationContextManagerAsyncLocal` while my code gets a separate `ApplicationContextManagerHttpContext`

Two different IContextManagers on logical server side is a problem.  We need to devise a way where, **during the new scope in `LocalProxy`**, only one is used and it is always `ApplicationContextManagerAsyncLocal`.

Back to the chicken-egg scenario:  

Maybe you'll think of a better way, because this is rough, but I think it will work.  After brainstorming on this for hours, it's the only way I can think of that respects DI:

FIRST:
1. Create a new scoped service called "LocalProxyState" which has a bool property "NewScopeExists".
2. See about making an abstract base class called `ApplicationContextManagerFacade` which implements `IContextManager`.  This can then be used for 3 below.
2. Make each `IContextManager` implementation where `IsStatefulRuntime=true` be a class that inherits from `ApplicationContextManagerFacade`, and which is a facade wrapper with an `ActiveContextManager` just like the current  `Csla.AspNetCore.ApplicationContextManager` class.
3. Each of these facade context managers needs to use the `LocalProxyState` in their CTOR to know if they should create the  `ActiveContextManager` as a `ApplicationContextManagerAsyncLocal`

SECOND:
1. Immediately after LocalProxy creates the new scope, it will fetch a fresh instance of this service
    `var lpScopState = provider.GetRequiredService<LocalProxyState>()` 
    then it sets
    `lpScopeState.NewScopeExists= true`
2. LocalProxy should then use DI to get it's `IContextManager` and everyone gets one happy scoped instance

[...]

PR #2830, opened 2022-03-14
Set service provider and context for LocalProxy and return value

### LINKED ISSUE #2808: ApplicationContext and ServiceProvider reference issues when stateful runtime because new scope
This is really hard to describe, I'll do my best.  I didn't notice it till I had my whole huge business layer converted and started running automated tests with login and whoopsy!   If I'm correct, it is a really nasty issue.
Using latest CSLA 6 repo (you don't have the nuget preview package for it yet.  

_I will say up front, if you disable the LocalProxy behavior of creating a new Scope, then all the trouble goes away and everything works just like it did in Csla 5.  You recently removed the LocalProxyOption to be able disable that via config? So I had to manually edit the CSLA to not do new new scope for testing.

To see the behavior, do the following:
1. Add a GUID property to Csla.ApplicationContext (helps you be able to quickly identify different instances)
2. Use a simple BSS project and inject an ApplicationContext and a simple ObjectAFactory
3. Call the ObjectAFactory.GetObjectA method which calls IDataPortal<ObjectA>Fetch() and populates / returns the instance.
4. Immediately after this call, notice the reference of original ApplicationContext DOES NOT equal the ObjectA.ApplicationContext
5. The "CurrentServiceProvider" references of the two instances are also different and thus even using .GetService(typeof(ApplicationContext)) returns two different instances

Big problem having two versions of ApplicationContext on the blazor page...the original AND the one inside your AObject.  
- If you need to add a key/value to ApplicationContext.ClientContext...which one do you change?  
- If you set the .User property on the original, the one inside your business object doesn't see that.

This trouble compounds significantly if ObjectA has other business objects as properties inside itself fetched while on the server side of the data portal or after in subsequent calls.  Each of those go through the dataportal on the server side and get out of sync even more.  Depending on the scenario you can get MANY different instances of ApplicationContext / CurrentServiceProvider.

There is no way to get a trustworthy source of truth about ContextDictionary values, User principal, etc.  There is no formal mechanism to set the ApplicationContext and CurrentServiceProvider references to a business object **and it's whole tree** when the invoke of LocalPortal is complete.

Like I mentioned, if you turn off the new scope feature in LocalProxy then everything works fine because you always just have ONE instance of ApplicationContext and it's associated CurrentServiceProvider.   But if we ditch the new scope feature, then we have to figure out some other way to make multiple calls from a stateful client UI technology like Blazor/WPF, etc. or at least some mechanism for a queuing/waiting such that one finishes before another starts (which sucks but at least it works).  I realized that I never had the new scope behavior before in CSLA and I was fine.  I knew that adding asyncronous was troublesome because you could end up trying to open two database connections in the same context and it would break.  But we got along just fine with that.  I was really excited about having that solved but the how to keep AppContext in sync?

I hope that is clear.  I'm stumped as to what to do.

rockfordlhotka:
#2798 
Closes #2808

### DIFF
--- a/Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
+++ b/Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
@@ -75,8 +75,6 @@ namespace Csla.AspNetCore
     {
       get
       {
-        var returnVal = false;
-
         if (HttpContext is null)
           return false;
 
@@ -92,6 +90,10 @@ namespace Csla.AspNetCore
       }
     }
 
+    /// <summary>
+    /// Gets a value indicating whether the context manager
+    /// is stateful.
+    /// </summary>
     public bool IsStatefulContext => false;
 
     /// <summary>
--- a/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
+++ b/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
@@ -109,6 +109,10 @@ namespace Csla.AspNetCore.Blazor
       get { return ActiveCircuitState.CircuitExists; }
     }
 
+    /// <summary>
+    /// Gets a value indicating whether the context manager
+    /// is stateful.
+    /// </summary>
     public bool IsStatefulContext => true;
 
     /// <summary>
--- a/Source/Csla/ApplicationContext.cs
+++ b/Source/Csla/ApplicationContext.cs
@@ -10,7 +10,6 @@ using System.Security.Principal;
 using Csla.Core;
 using System.Security.Claims;
 using Microsoft.Extensions.DependencyInjection;
-using System.Collections.Generic;
 
 namespace Csla
 {
@@ -23,38 +22,20 @@ namespace Csla
     /// <summary>
     /// Creates a new instance of the type
     /// </summary>
-    /// <param name="contextManagerList">List of registered IContextManager</param>
-    /// <param name="serviceProvider">Current service provider</param>
-    /// <param name="runtimeInfo"></param>
-    public ApplicationContext(IEnumerable<IContextManager> contextManagerList, IServiceProvider serviceProvider, Runtime.IRuntimeInfo runtimeInfo)
+    /// <param name="applicationContextAccessor"></param>
+    public ApplicationContext(ApplicationContextAccessor applicationContextAccessor)
     {
-      _serviceProvider = serviceProvider;
-      RuntimeInfo = runtimeInfo;
-
-      foreach (var context in contextManagerList)
-      {
-        if (context.IsValid)
-        {
-          ContextManager = context;
-          break;
-        }
-      }
-      if (ContextManager is null)
-      {
-        ContextManager = new Core.ApplicationContextManagerAsyncLocal();
-      }
-
-      ContextManager.ApplicationContext = this;
+      ApplicationContextAccessor = applicationContextAccessor;
     }
 
-    private Runtime.IRuntimeInfo RuntimeInfo { get; set; }
+    internal ApplicationContextAccessor ApplicationContextAccessor { get; set; }
 
     /// <summary>
     /// Gets the context manager responsible
     /// for storing user and context information for
     /// the application.
     /// </summary>
-    public IContextManager ContextManager { get; internal set; }
+    public IContextManager ContextManager => ApplicationContextAccessor.GetContextManager();
 
     /// <summary>
     /// Get or set the current ClaimsPrincipal
@@ -378,16 +359,10 @@ namespace Csla
     /// </summary>
     public bool IsAStatefulContextManager => ContextManager.IsStatefulContext;
 
-    private IServiceProvider _serviceProvider;
-
     /// <summary>
-    /// Sets the service provider scope for this application context.
+    /// Gets the service provider scope for this application context.
     /// </summary>
-    internal IServiceProvider CurrentServiceProvider
-    {
-      get => _serviceProvider;
-      set => _serviceProvider = value;
-    }
+    internal IServiceProvider CurrentServiceProvider => ApplicationContextAccessor.ServiceProvider;
 
     /// <summary>
     /// Creates an object using dependency injection, falling back
--- a/Source/Csla/Configuration/ConfigurationExtensions.cs
+++ b/Source/Csla/Configuration/ConfigurationExtensions.cs
@@ -64,6 +64,9 @@ namespace Csla.Configuration
 
     private static void RegisterContextManager(IServiceCollection services)
     {
+      services.AddScoped<Core.ApplicationContextAccessor>();
+      services.TryAddScoped(typeof(Core.IContextManagerLocal), typeof(Core.ApplicationContextManagerAsyncLocal));
+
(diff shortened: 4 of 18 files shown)


rockfordlhotka:
@swegele I think I have it now!

This code works in my testing, and avoids having to walk the entire tree, because it just fixes up the `ApplicationContext` created by the child DI scope - the one used by all the objects created on the logical server.

Likewise, on the way in, it fixes up the `ApplicationContext` created on the logical client.

I suspect this won't work well if `AutoCloneOnUpdate` is turned off - I haven't tested that yet. And that's (imo) such an edge case, that people who still rely on that being turned off can also turn off the use of child scopes in `LocalProxy`.

swegele:
Getting closer!  But as luck would have it, the first test in my automated suite still shows a problem.  

Here's the flow:  
_(BTW - don't be distracted by my use of a custom identity...i'm still using it for login, but once fetched I use it to create a new ClaimsP and ClaimsI as you suggest on #2806)_

1. Login test calls static Login method on `MyIdentity` which is a `ReadOnlyBase`.  
2. `MyIdentity` calls portal.Fetch(string username, string password)
3. Logic on server side detects that a password change is required due to first login
4. A `ForcePasswordChange` object is create (which is a `ChildBusinessBase`).  It is populated and returned to the client side by throwing a custom exception `LoginFailedPasswordChangeRequiredException`
5. This custom exception has a `PasswordChange` property which holds that child object.
6. The test catches this login failed exception and grabs the `ForcePasswordChange` child object which has editable `NewPassword` and `NewPasswordConfirm` properties.  It also has the business rules that check for password strength and password history etc.
8. The test properly fills out the new password change fields
9. The test calls another static method  `MyIdentity.LoginWithForcePasswordChange(string username, string password, ForcePasswordChange forcePasswordChange)`
10.  `MyIdentity` calls `portal.Fetch(username, password, forcePasswordChange)` sending this child object across the dataportal of as a parameter


When I arrive on the server side:
- MyIdentity.ApplicationContext.LogicalExecutionLocation == Server
- The criteria object contains the ForcePasswordChange object who's ApplicationContext.LogicalExecutionLocation == Client

This shows that **objects passed across the portal as parameters** are not properly getting their AppContext's sync'd.  
I think we still need to walk the tree of parameters looking for and resetting IUseApplicationContext.

rockfordlhotka:
Thanks @swegele - it had occurred to me that I'd need to fully swap the `ApplicationContext` itself in any case, to avoid async timing issues between client and logical server.

My latest commit now swaps the `ApplicationContext` of the criteria/object graph as it comes into `LocalProxy`, and then swaps it back on the way out, so I think (🤞) that this should work?

swegele:
A potential issue - and a slap on my hand :-) 

**Potential Issue:**
If I am reading the code right in LocalProxy, I don't think an obj of type `CriteriaBase` will enter into your logic for the recursive loop because it does not implement `IManageProperties`.  
It inherits from `ManagedObjectBase` which does have `protected FieldDataManager FieldManager`
So I think you could make this:
`if (obj is IManageProperties target)`

into this:
`if (obj is IManageProperties target || obj is ManagedObjectBase)` 


**Slap on my hand**
Looking at this made me realize I've made a bunch of custom criteria objects that do not inherit from `CriteriaBase` (whoops - caught me in a bad habit).  I have to clean those up.

swegele:
Could you also have this loop, go through the registered properties of the object looking for instances of IUseApplicationContext?  

Currently I see it loops through GetChildren...but I don't see that it looks through the "GetManagedProperties" (or something like that).  
For example:  A readonly object can have a property that is another readonly object or whatever but it is not a child.

rockfordlhotka:
The latest commit

1. Loops through all managed properties
2. Does a cleaner job of finding FieldManager and BusinessRules properties
3. Cleans up the code somewhat imo

swegele:
That IS cleaner, I like it.  99% there - so close I can feel it!  
Only one last scenario that I can see: 

`CriteriaBase` doesn't implement `IManageProperties`.  It only has `ReadProperty` and `LoadProperty.`  
Thus it doesn't have the method `GetManagedProperties` and thus it's managed properties don't get processed.  It does have a `FieldManager` so it could very well have that method to get it's properties.

**Option 1**
1. Add method `List<IPropertyInfo> GetManagedProperties()` to the `ManagedObjectBase`
2. Add one more condition to LocalProxy like so `if (obj is ManagedObjectBase target)` which also does that loop

**Option 2**
1.  Break current `IManageProperties` interface into two
2.  First a new interface ?`ISimpleManageProperties` - more bare bones with (Read/Load/FieldExists/GetManagedProperties)
3.  Second is the current `IManageProperties` which implements `ISimpleManageProperties` adding (Get/Set/GetChildren)
4.  Change the condition in `LocalProxy` to check `if (obj is ISimpleManageProperties target)`

You'll probably think of a better Option 3

rockfordlhotka:
@swegele good catch - I've implemented `IManageProperties` in `ManagedObjectBase` - which not only fixes CriteriaBase, but also CommandBase (and so avoids a future bug).

swegele:
`ApplicationContext` swap is now digging into all the objects!  WAY TO GO!!  

But during the loop operation it blows up when editable object with a lazy loaded property is encountered.

The LocalProxy, upon return from DataPortal, is looping through the GetManagedProperties() and it stumbles on a lazy loaded property which is not fetched yet.  This throws the following exception from BusinessBase.cs

```cs
protected virtual object ReadProperty(IPropertyInfo propertyInfo)
{
  if (((propertyInfo.RelationshipType & RelationshipTypes.LazyLoad) == RelationshipTypes.LazyLoad) && !FieldManager.FieldExists(propertyInfo))
    throw new InvalidOperationException(Resources.PropertyGetNotAllowed);
```

rockfordlhotka:
@swegele have I mentioned how much fun it is to work with someone that has your attention to detail? 👍

swegele:
Your last fix has a parenthesis in the wrong spot which caused a regression.
It should be this (just tested it):

```cs
if (typeof(IUseApplicationContext).IsAssignableFrom(item.Type) &&
  !((item.RelationshipType & RelationshipTypes.LazyLoad) == RelationshipTypes.LazyLoad && !target.FieldExists(item)))

```

_P.S.  I'm so glad to be able to give back a little in appreciation for the 18 years (since 2004) of "breadcrumb path" you've given me in my career!  And TBH it's our automated tests that detail oriented not me :-)  I'm getting too old to remember all this stuff._

rockfordlhotka:
Hopefully the last fix @swegele 😄

swegele:
Looks like that fixes it 💯 

I've got to go through the BL and fix all Command and Criteria objects not properly using managed properties.  When I've got that finished then I can run my full automated test suite.  Since that might take a day or so, I will say this issue is closed and I will open new one if I find anymore stuff.  Either way I will report back the results.

Thanks for this.

swegele (approved):
👍

swegele:
One last thought/question:  This could be further discussed on the issue #2808 if you prefer to push this PR through.

Regarding setting ApplicationContext only when using LocalProxy:

When coming OUT OF (back on the clientside) all data portals (even remote ones), wouldn't we want to reset ApplicationContext so that all objects back on the client side have the one correct reference?

PR #2837, opened 2022-03-17
Update samples for CSLA 6 prerelease 7

### LINKED ISSUE #2500: Please update the Uno Example
Please update the Uno Example.  Thanks

### LINKED ISSUE #2552: Update samples for CSLA 6
Update samples for CSLA 6

Note that ProjectTracker has its own issue due to its expansive scope: #2551

### LINKED ISSUE #2715: Error 'Could not load file or assembly 'System.Runtime.Loader, Version=4.0.0.0, ' remains for Xamarin Forms application
**Describe the bug**
Same issue as described in #2398 except for Xamarin.Forms/Xamarin.Android

**Version and Platform**
CSLA version: 5.5.2
OS: Android
Platform: Xamarin

**Code that Fails**
Executing of DataPortal.Fetch fails with System.Runtime.Loader error

**Additional context**
Possibly related to CSLA requiring Nuget spec requiring v4.3 but lack of support for binding redirects prevents me from being able to specify v4.0. 

Screenshot 2022-01-19 123644
Untitled

### LINKED ISSUE #2839: Rule context objects should expose ApplicationContext
**Describe the bug**

This relates to changes in CSLA 6 preview 7 for issue #2796

It will be helpful if the `AddObjectAuthorizationRulesContext`  can expose `ApplicationContext` directly. Since there is a public override method on `BusinessRule.AddRule` that takes a parameter of ApplicationContext, it makes sense to have the ApplicationContext available from the AddObjectAuthorizationRulesContext class.

Suggested changes in `AddObjectAuthorizationRulesContext`:
Change property: 
````
private ApplicationContext ApplicationContext { get; set; }
````
to:
````
public ApplicationContext ApplicationContext { get; private set; }
````

Suggested changes in `IAddObjectAuthorizationRulesContext`:
Add: 
````
    /// <summary>
    /// Gets the application context
    /// </summary>
    ApplicationContext ApplicationContext { get; }
````

**Version and Platform**
CSLA version: 6 prelease 7
OS: Windows, Linux, iOS, Android, etc.
Platform: WinForms, WPF, ASP.NET Core, MVC, Xamarin, etc.

### LINKED ISSUE #2850: In pure AspNet .NET 6 (razor or mvc only - no blazor)...Csla requires registration of AuthenticationStateProvider
Given an .NET 6 AspNet web site, 
using the Csla and the Csla.AspNetCore packages,
with this in program.cs `builder.Services.AddCsla(options => options.AddAspNetCore());`

When you use DI to get an ApplicationContext like so: 
`ServiceProvider.GetRequiredService<ApplicationContext>();`

You get an error from DI saying that it cannot 
resolve `Microsoft.AspNetCore.Components.Authorization.AuthenticationStateProvider`

This is because of the following in Csla.AspNetCore
`config.Services.AddScoped(typeof(IContextManager), typeof(ApplicationContextManagerBlazor));`

If you register an instance then DI works:
`builder.Services.AddScoped<AuthenticationStateProvider, ServerAuthenticationStateProvider>()`

Having Csla users remember to register this should not be required.  Several ideas:
1.  Use some configuration to only register the `ApplicationContextManagerBlazor` when BSS is being used.
or
2.  Add a default to `AddAspNetCore` of `AuthenticationStateProvider`

Thoughts?

### LINKED ISSUE #2851: Remove .NET 5 support
.NET 5 goes off support by Microsoft in May 2022.

I think we should remove .NET 5 support before CSLA 6 launches, to keep this breaking change in CSLA 6.

https://devblogs.microsoft.com/dotnet/dotnet-5-end-of-support-update/

### LINKED ISSUE #2854: CslaDataProvider needs to work with CSLA 6 dependency injection
In the past this is the sort of code used to initialize a `CslaDataProvider`:

```c#
      var dp = Resources["Order"] as Csla.Xaml.CslaDataProvider;
      using (dp.DeferRefresh())
      {
        dp.ObjectType = typeof(Test.Library.OrderEdit);
        dp.FactoryMethod = "GetObject";
        dp.FactoryParameters.Clear();
        dp.FactoryParameters.Add(123);
      }
```

It is no longer valid (or at least not recommended) to use the static `GetObject` factory method. Instead, the `CslaDataProvider` needs to have a way to accept a more modern scheme. Perhaps having an overload for `FactoryMethod` that accepts a delegate of type `Task<T>` or something along that line. More like:

```c#
      var portal = ApplicationContext.GetRequiredService<IDataPortal<Test.Library.OrderEdit>>();
      var dp = Resources["Order"] as Csla.Xaml.CslaDataProvider;
      using (dp.DeferRefresh())
      {
        dp.FactoryMethod => portal.Create(123);
      }
```

It seems like this will work with DI, and also simplify the code overall.

### LINKED ISSUE #2866: IDataPortalActivator create and resolve are obsolete
In CSLA 6 the idea of dependency injection is now directly used by CSLA, and the `IDataPortalActivator` `CreateInstance` and `ResolveType` methods are no longer invoked.

Right now we'll mark them as `Obsolete`, and will remove the methods in a future point release.

### LINKED ISSUE #2867: DataPortalServerOptions methods aren't fluent as expected
The methods in `DataPortalServerOptions` need to return `this` rather than `void` so they can be invoked in a fluent manner.

### LINKED ISSUE #2868: Some types are missing protected ApplicationContext
The bindinglist collection types have private instead of protected `ApplicationContext` properties.

### LINKED ISSUE #2869: ObjectFactory needs access to ApplicationContext
To implement the factory data access models, the factory class needs to be able to create an instance of business domain types. The only safe way to do this in CSLA 6 is via 

```c#
ApplicationContext.CreateInstanceDI<BusinessType>();
```

which obviously means that code in a subclass of `ObjectFactory` needs access to the current `ApplicationContext` instance.

This is a breaking change, because the `ApplicationContext` value will be provided via constructor injection. All business implementations of an object factory will need to add a constructor:

```c#
public class MyFactory : ObjectFactory
{
  public MyFactory(ApplicationContext applicationContext)
    : base(applicationContext) { }
}
```

### LINKED ISSUE #2877: RevalidatingInterceptor doesn't handle a root collection type
If the root object is a collection, RevalidatingInterceptor needs to run rules on all items in the list.

### LINKED ISSUE #2878: RevalidatingInterceptor doesn't handle async rules
RevalidatingInterceptor needs to use `CheckRulesAsync` and then wait until all rules are complete before proceeding.

### LINKED ISSUE #2879: RevalidatingInterceptor shouldn't trigger lazy loading of properties
If an object has a lazy loaded property that hasn't yet been loaded, RevalidatingInterceptor should not trigger loading of that property.

rockfordlhotka:
Closes #2552 
Closes #2850  
Closes #2839 
Closes #2500 
Closes #2851 
Closes #2715 
Closes #2854 
Closes #2866 
Closes #2867 
Closes #2868 
Closes #2869 
Closes #2877 
Closes #2878 
Closes #2879


swegele:
Not sure I understand why you are creating a custom AuthenticationStateProvider for BSS?  What is that for?

When using BSS, we add this in Program.cs - `builder.Services.AddServerSideBlazor();` 
This tells .NET framework to automatically register their [built in version](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.server.serverauthenticationstateprovider?view=aspnetcore-6.0) which is `ServerAuthenticationStateProvider`


[Steve Sanderson seems to indicate](https://gist.github.com/SteveSandersonMS/175a08dcdccb384a52ba760122cd2eda?permalink_comment_id=3301084#implementing-a-custom-authenticationstateprovider) that the need for a custom auth state provider is quite rare and even potentially introduces vulnerabilities.  Suggests it is common only in Blazor WASM.

Everything else looks good though.

rockfordlhotka:
And this is why a second pair of eyes is so important! Thank you.

swegele:
If it's not too much trouble, after this behemoth check-in, could you do a preview 8 if you are going to do another preview?  Would help our dev team here be able move forward with the apps that will use the BL i'm giving them.  Without them having to do funky post-build dll copies from the Csla project.

Exciting times!!

swegele:
FWIW - Our updated business layer passed all our automated integration tests (>10K).  So I'm confident there are no glaring or obvious regressions in Csla 6.

swegele:
Z   <--  Do you know what that is?  No it's not the letter "Z".  It is me on my knees begging for an updated nuget package.

I've got four developers for almost 2 weeks who all must get the latest Csla repo to their respective machines for a build.  They then manually add post-build copy directives like so:

```cs
xcopy /y /d "C:\Users\SeanW\source\repos\csla\Bin\Debug\net6.0\Csla.dll" "$(TargetDir)"
xcopy /y /d "C:\Users\SeanW\source\repos\csla\Bin\Debug\net6.0\Csla.pdb" "$(TargetDir)"
```

so that the newer Csla dlls (with all the fixes) overwrite the ones in the last preview nuget package.  It works but it is pretty cumbersome.  Please...tell me where to send the bribe money hehe.

rockfordlhotka:
@swegele I hear you! As I was saying yesterday in Discord, I am working on SimpleNTier, which is the final sample other than ProjectTracker (which I'll do separately).

Thanks to SimpleNTier I discovered a bug in one of our new features, and am working through solving it now.

I expect (hope) to resolve it yet this evening or tomorrow, and so there should be a prerelease tomorrow, or over the weekend.

Issue #2842, opened 2022-03-18
SetApplicationContext in LocalProxy doesn't process DeletedList items


swegele:
Following on the items fixed in #2830 

My tests are showing that items in `protected MobileList<C> DeletedList` are not getting processed (if I'm understanding it right).

PR #2844, opened 2022-03-19
 proposed fixes 

### LINKED ISSUE #2841: Fix for CreateInstanceDI which makes ServiceProvider parameter resolution work as expected
**EDIT:**
Whoops...fixed `if (parameters is null || parameters.Length > 0)` 
to be `if (parameters is null || parameters.Length == 0) `

@rockfordlhotka  Could I ask for a small tweak?  Depending on how you look at it, it's a bug or it's an enhancement...

In the current `ApplicationContext.CreateInstanceDI` method, when `CurrentServiceProvider` is NOT null, you call `ActivatorUtilities.CreateInstance(CurrentServiceProvider,...`  This works great when there are one or more manually passed in parameters. 

But it doesn't always work when `parameters.Length == 0` and the type/service being created has parameters that are registered with DI.   Microsoft's comment on their method is misleading making you think that DI will take care of resolution in that scenario but it does not.  It throws an error saying **"no suitable constructor found...bla bla"**.  
Now I know that is false because I put a breakpoint there and in the immediate window type `CurrentServiceProvider.GetRequiredService(objectType)` it works fine.  This means that the CreateInstance method is doing something different than the simple GetRequiredService call.

I've tested a very simple fix (IMO) that works.

**Currently it is like this:**
```cs
  if (CurrentServiceProvider != null)
    result = ActivatorUtilities.CreateInstance(CurrentServiceProvider, objectType, parameters);
  else
    result = Activator.CreateInstance(objectType, parameters);
...
```

**After adding one extra if condition:**
```cs
if (CurrentServiceProvider != null)
{
  if (parameters is null || parameters.Length == 0)
  {
    result = CurrentServiceProvider.GetRequiredService(objectType);
  }
  else
  {
    result = ActivatorUtilities.CreateInstance(CurrentServiceProvider, objectType, parameters);
  }
}
else
...
```

After this fix these three scenarios work:
1.  (works like it did before) When `CurrentServiceProvider` is null 
2.  (works like it did before) When `CurrentServiceProvider` is NOT null, and there are manually supplied parameters
3.  (now works)  When `CurrentServiceProvider` is NOT null, and there are NO supplied parameters, DI properly gets registered service and automatically resolves injected dependencies into that service type's constructor.

swegele:
#2841 and #2842

- New method `CreateInstanceViaServiceProvider` which uses **true** DI for object creation instead of CreateInstance. 

 [This answer](https://stackoverflow.com/a/45778712) from David Fowler finally helped me realize the difference between the current methods using CreateInstance versus ServiceProvider.GetRequiredService.  

The current methods create the object by CreateInstance without requiring DI registration.  During that operation the ServiceProvider (if present) merely attempts to resolve any missing parameters in its CTOR that were not manually passed in. 

Whereas, the new method actually uses DI for both _object creation AND for parameter resolution_.
I updated the comments on those existing methods to hopefully add clarity and distinctness.

- New interface `IContainsDeletedList` for the two objects which have a `DeletedList`

- Process `IContainsDeletedList` objects in LocalProxy for `SetApplicationContext`

### DIFF
--- a/Source/Csla/IContainsDeletedList.cs
+++ b/Source/Csla/IContainsDeletedList.cs
@@ -1,4 +1,12 @@
-﻿using System;
+﻿//-----------------------------------------------------------------------
+// <copyright file="IContainsDeletedList.cs" company="Marimer LLC">
+//     Copyright (c) Marimer LLC. All rights reserved.
+//     Website: https://cslanet.com
+// </copyright>
+// <summary>Interface implemented by a list which tracks its deleted items</summary>
+//-----------------------------------------------------------------------
+
+using System;
 using Csla.Core;
 using System.ComponentModel;
 using Csla.Serialization.Mobile;
--- a/Source/Csla/IContainsDeletedList.cs
+++ b/Source/Csla/IContainsDeletedList.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using Csla.Core;
+using System.ComponentModel;
+using Csla.Serialization.Mobile;
+using System.Collections.Specialized;
+using System.Collections.Generic;
+
+namespace Csla
+{
+  /// <summary>
+  /// Defines an object that holds a list of deleted items.
+  /// </summary>
+  public interface IContainsDeletedList
+  {
+    /// <summary>
+    /// List of deleted child objects
+    /// </summary>
+    IEnumerable<IEditableBusinessObject> DeletedList { get; }
+  }
+}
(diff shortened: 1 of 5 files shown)


Source/Csla/IContainsDeletedList.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
rockfordlhotka, line 1: This needs the header comment block like other files.

rockfordlhotka (changes requested):

Source/Csla/IContainsDeletedList.cs
swegele: I always struggle with wording/naming.  Hopefully that is clear.

swegele:
Still learning how this all works.  I don't know if it's appropriate for me to:
- mark a conversation as resolved or should only you do that?
- re-request review after making change you requested...or is that overkill?

rockfordlhotka (approved):

PR #2860, opened 2022-03-31
Csla 6.0 additional test fixes

TheCakeMonster:
I've taken another pass through the .NET Core version of the tests and resolved some more bugs, all of which were with the tests themselves (mostly issues to do with the upgrade of test object definitions to match CSLA 6 requirements, especially associated with the data portal switching from static to instantiated via DI)

Includes a couple of small changes to the package references to bring them up to the same version used for all other projects.

### DIFF
--- a/Source/Csla.test/Basic/Child.cs
+++ b/Source/Csla.test/Basic/Child.cs
@@ -18,7 +18,7 @@ namespace Csla.Test.Basic
     private string _data = "";
     private Guid _guid = System.Guid.NewGuid();
 
-    private GrandChildren _children = Csla.Test.Basic.GrandChildren.NewGrandChildren();
+    public static PropertyInfo<GrandChildren> GrandChildrenProperty = RegisterProperty<GrandChildren>(c => c.GrandChildren);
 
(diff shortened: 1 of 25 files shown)


rockfordlhotka:
@TheCakeMonster do all the tests pass at this point? If so, we should add the new test project to the csla.test.sln so it becomes part of the CI build.

TheCakeMonster:
> @TheCakeMonster do all the tests pass at this point? If so, we should add the new test project to the csla.test.sln so it becomes part of the CI build.

No, not yet - we're still not all that close at this point. There are 92 failing tests under .NET 6 and 269 under .NET Framework. The high number under NETFX is being caused by System.Runtime.Loader not being supported on that platform, I think - but I've been concentrating on getting close with .NET 6 first.

I think I might be starting to run out of talent, in the sense that each of the remaining failures would probably take me a long time to understand and investigate. The remaining failures don't seem to have a common theme/cause.

rockfordlhotka:
I think the loader issue will be fixed with the other pending pr.

PR #2902, opened 2022-04-22
Optimize reset of applicationcontext on return from LocalProxy

### LINKED ISSUE #2885: Business layer code (using LocalProxy) is ~4x slower after preview 8
We have a huge automated test suite in the form of "Integration Tests" that run against an instance of LocalDB (SQL server) and test around 80% of our business layer.

Preview 7 - tests take around 6 minutes to run

Preview 8 - tests take around 20 minutes to run

Given our code didn't change, something fairly basic and significant changed related to performance in the framework.  Got any ideas what the culprits might be?
In this latest preview we added a loop in the `LocalProxy.SetApplicationContext` which might be a suspect just because it's more processing.  Not sure what else might be an issue.  Thinking off the top of my head, I'm wondering if we could add some parallelism in that to increase performance.  Thinking out-loud.

Not necessarily saying it's a problem since automated tests don't really mimic real live users who are not that fast and clicking :-)  But it's just note worthy at least.

### LINKED ISSUE #2906: Remove GetDataPortal methods from base classes
The base classes (like `BusinessBase`) currently expose two ways to get a data portal instance. 

```c#
  var portal = GetDataPortal<MyType>();
  var portal = ApplicationContext.GetRequiredService<IDataPortal<MyType>>();
```

This is confusing and should be simplified.

Given that `ApplicationContext` is required for many other scenarios, it can't be removed. The `GetDataPortal` method is only used for this scenario, and is a duplicate solution, so it is the obvious candidate to be removed.

rockfordlhotka:
Closes #2885
Closes #2906

### DIFF
--- a/Source/Csla.Axml.Android/Resources/Resource.Designer.cs
+++ b/Source/Csla.Axml.Android/Resources/Resource.Designer.cs
@@ -14,7 +14,7 @@ namespace Csla.Axml
 {
 	
 	
-	[global::System.CodeDom.Compiler.GeneratedCodeAttribute("Xamarin.Android.Build.Tasks", "12.2.99.0")]
+	[global::System.CodeDom.Compiler.GeneratedCodeAttribute("Xamarin.Android.Build.Tasks", "12.3.0.26")]
 	public partial class Resource
 	{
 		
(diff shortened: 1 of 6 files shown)


rockfordlhotka:
@swegele this still does go through the object graph at the start of the call (which should only be expensive for an update), but uses a single call to change the scoped `ApplicationContext` object's state to the caller's state on the return.

So for create/fetch operations this should be _much_ faster, and should see a 50% (roughly) boost for update operations.

PR #2932, opened 2022-05-11
Fixes to more base tests in line with CSLA 6

TheCakeMonster:
Fixes to more tests today, mostly issues with the tests due to a lack of experience with some parts of the framework. However, there was one issue with the Csla.TestHelpers that it was important to resolve.

Only 51 failing now under .NET 6, and 130 under .NET Framework

I think there are some outstanding issues with CSLA 6 around how PropertyChanged is called; a number of tests are failing because the number of PropertyChanged events raised is not the same as in previous versions.

There may also be a problem with per-property authorization. We have some per-property auth tests that only fail when all of the tests are run at once; they pass if you run the auth tests on their own. I'm wondering if this is an issue caused by high thread usage or something, to do with the AsyncLocal context manager? Not sure yet.


rockfordlhotka:
fyi, I think there've always been some issues with some of the authz tests if they are run concurrently.

PR #3150, opened 2022-10-21
HttpClient no longer stored in a static field

### LINKED ISSUE #3148: HttpProxy should not store HttpClient in a static field
Oh crickey! Yeah, @saulyt, I think you've hit the nail on the head. I agree that the HttpClient being stored in a static variable is very odd. Well done, I think that might explain some of the problems we are seeing.

It's been a static variable for a very long time, but we are only now seeing it cause a problem. I think that is because we have changed the way things get disposed. If I recall, we did that as part of adding handling for the newish async disposal that has been added to .NET.

_Originally posted by @TheCakeMonster in https://github.com/MarimerLLC/csla/discussions/3128#discussioncomment-3936236_

rockfordlhotka:
Closes #3148

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,9 +2,9 @@ name: main
 
 on:
   push:
-    branches: [ main ]
+    branches: [ v6.x ]
   pull_request:
-    branches: [ main ]
+    branches: [ v6.x ]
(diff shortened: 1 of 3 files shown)


saulyt (approved):
Thanks

PR #3269, opened 2023-03-16
Show warning and info from validation on create page for MVC sample

danielmartind:
Updates MVC sample as #3261 

Tags to show error, warning and info from validation rules added to the create page in the MVC sample so it works in the same way as the edit page


rockfordlhotka (approved):

rockfordlhotka:
Thank you @danielmartind !

Issue #3286, opened 2023-04-13
Add ManageObjectLifetime to Blazor.ViewModel


russblair:
**Is your feature request related to a problem? Please describe.**
In need to support n-level undo in Blazor so that I can implement a Cancel button. Calling RefreshAsync when the Cancel button is clicked breaks all the child component bindings. I also need to implement ChildChanged and CollectionChanged events to allow changes made to descendant objects to bubble up to the editable root.

**Describe the solution you'd like**
I would like to see additional functionality added to the Blazor.ViewModel to support ManageObjectLifetime similar to the Csla.Xaml.Shared.ViewModelBase. This includes adding the ModelChildChanged event, ModelCollectionChanged event, SetProperties event, and the IsBusy event.

**Describe alternatives you've considered**
I was unable to access the ViewModel.Model BeginEdit, ApplyEdit and CancelEdit from my page view models. I created an extended ViewModel that inherits from Blazor.ViewModel which worked but felt this functionality should be part of Csla.

russblair:
To use n-level undo in your Blazor application, use the following code to extend the ViewModel.
```
using Csla;
using Csla.Blazor;

namespace MyProject.UI.Blazor.Client.Shared
{
    /// <summary>
    /// ViewModel extended to provide support for ManageObjectLifetime.
    /// </summary>
    public class ViewModelManageObjectLifetime<T> : ViewModel<T>
    {
        /// <summary>
        /// Creates an instance of the type.
        /// </summary>
        /// <param name="applicationContext"></param>
        public ViewModelManageObjectLifetime(ApplicationContext applicationContext) : base(applicationContext)
        {
            ManageObjectLifetime = true;
        }

        /// <summary>
        /// Cancels changes made to the model 
        /// if ManageObjectLifetime is true.
        /// </summary>
        public virtual void Cancel()
        {
            DoCancel();
        }
    }
}
```

PR #3289, opened 2023-04-17
Add ManageObjectLifetime to Blazor.ViewModel Issue #3286

russblair:
**Add ManageObjectLifetime to Blazor.ViewModel Issue #3286**

Using Xaml.Shared.ViewModelBase as the example, I added ModelChildChanged, ModelCollectionChanged, IsBusy, BusyChanged, ManageObjectLifetime, and DoCancel.

Making these changes led to additional questions that I have noted in the code using "// TODO:".  Rocky, please let me know how you wish to proceed.

### DIFF
--- a/Source/Csla.Blazor/ViewModel.cs
+++ b/Source/Csla.Blazor/ViewModel.cs
@@ -232,17 +232,10 @@ namespace Csla.Blazor
       try
       {
         UnhookChangedEvents(Model);
-        var savable = Model as Core.ISavable;
-        if (ManageObjectLifetime)
-        {
-          // clone the object if possible
-          if (Model is ICloneable cloneable)
-            savable = (Core.ISavable)cloneable.Clone();
 
-          //apply changes
-          if (savable is Core.ISupportUndo undoable)
+        if (ManageObjectLifetime && Model is Core.ISupportUndo undoable)
             undoable.ApplyEdit();
-        }
+
         IsBusy = true;
         Model = await DoSaveAsync();
         Saved?.Invoke();
@@ -271,13 +264,8 @@ namespace Csla.Blazor
     protected virtual async Task<T> DoSaveAsync()
     {
       if (Model is Core.ISavable savable)
-      {
-        var result = (T)await savable.SaveAsync();
-        if (Model is Core.IEditableBusinessObject editable)
-          new Core.GraphMerger(ApplicationContext).MergeGraph(editable, (Core.IEditableBusinessObject)result);
-        else
-          Model = result;
-      }
+        await savable.SaveAndMergeAsync();
+
       return Model;
     }
 
--- a/Source/Csla.Blazor/ViewModel.cs
+++ b/Source/Csla.Blazor/ViewModel.cs
@@ -51,6 +51,11 @@ namespace Csla.Blazor
     /// raises its PropertyChanged event
     /// </summary>
     public event PropertyChangedEventHandler ModelPropertyChanged;
+    /// <summary>
+    /// Event raised when the Model object
+    /// raises its ChildChanged event
+    /// </summary>
+    public event Action ChildChanged;
 
     /// <summary>
     /// Raises the ModelChanging event
@@ -60,11 +65,24 @@ namespace Csla.Blazor
     protected virtual void OnModelChanging(T oldValue, T newValue)
     {
       _info.Clear();
+      if (oldValue is IBusinessBase oldBusinessBase)
+        oldBusinessBase.ChildChanged -= (s, e) => OnChildChanged();
+      if (newValue is IBusinessBase newBusinessBase)
+        newBusinessBase.ChildChanged += (s, e) => OnChildChanged();
+
       if (oldValue is INotifyPropertyChanged oldObj)
         oldObj.PropertyChanged -= (s, e) => OnModelPropertyChanged(e.PropertyName);
       if (newValue is INotifyPropertyChanged newObj)
         newObj.PropertyChanged += (s, e) => OnModelPropertyChanged(e.PropertyName);
       ModelChanging?.Invoke(oldValue, newValue);
+    }
+
+    /// <summary>
+    /// Raises the ChildChanged event
+    /// </summary>
+    protected virtual void OnChildChanged()
+    {
+      ChildChanged?.Invoke();
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


rockfordlhotka (approved):

russblair:
Hi @rockfordlhotka, did you have any guidance in regards to the 4 questions prefaced with "// TODO:"?

Source/Csla.Blazor/ViewModel.cs
```diff
@@ -194,12 +332,31 @@ public T Model
...
     }
 
-    private readonly Dictionary<string, object> _info = new Dictionary<string, object>();
+    /// <summary>
+    /// Gets or sets a value indicating whether the
+    /// view model should manage the lifetime of
+    /// the business object, including using n-level
+    /// undo.
+    /// </summary>
+    //[Browsable(false)]    // TODO: ask if this is desired
```
rockfordlhotka, line 341: That attribute controls whether the property appears in things like the IDE and intellisense. I think it probably _should_ be browsable, which is the default.

Source/Csla.Blazor/ViewModel.cs
```diff
@@ -194,12 +332,31 @@ public T Model
...
+    public bool ManageObjectLifetime { get; set; } = false;
+
+    /// <summary>
+    /// Gets a value indicating whether this object is
+    /// executing an asynchronous process.
+    /// </summary>
+    public bool IsBusy { get; protected set; } = false;
+
+    // TODO: should we implement IsDirty, IsValid, CanSave, CanCancel, CanCreate, CanDelete, CanFetch, CanRemove, CanAddNew, SetProperties()
```
rockfordlhotka, line 350: I'm not a fan of echoing things that can be bound directly off `vm.Model`. If implementing these things in the view model add functionality or enable scenarios that aren't otherwise available that's one thing, but if they just echo what's already available then that's extra code to maintain for no value.

Source/Csla.Blazor/ViewModel.cs
```diff
@@ -296,15 +459,21 @@ protected virtual string ModelErrorText
...
+    /// the view model during refresh or save
     /// operations.
     /// </summary>
     public Exception Exception { get; private set; }
 
+    #endregion
+
+    #region ObjectLevelPermissions
+
+    // TODO: Should we implement SetPropertiesAtObjectLevel(); and local state caching like Xaml.Shared.ViewModelBase?
```
rockfordlhotka, line 471: Yes, I think so - good catch!

rockfordlhotka (changes requested):

Source/Csla.Blazor/ViewModel.cs
```diff
@@ -176,6 +283,37 @@ protected virtual async Task<T> DoSaveAsync()
...
+          finally
+          {
+            HookChangedEvents(Model);
+            //OnSetProperties();
+          }
+        }
+      }
+    }
+
+    // TODO: should we implement BeginAddNew(), DoRemove(), DoDelete()
```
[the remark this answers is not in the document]
russblair: Ok. I removed the comment.

Source/Csla.Blazor/ViewModel.cs
russblair: Ok. I removed the comment line.

Source/Csla.Blazor/ViewModel.cs
russblair: Agreed.

Source/Csla.Blazor/ViewModel.cs
russblair: Added.

rockfordlhotka (changes requested):
There are a couple places in the save operations where the object is cloned. I think that should actually be using SaveAndMergeAsync rather than cloning.

russblair:
I copied the cloning code from the Xaml.Shared.ViewModelBase. Would you like me to change the Xaml.Shared.ViewModelBase as well? )Both the SaveAndMergeAsync logic and remove the [Browsable(false)] from the ManageObjectLifetime property.)

rockfordlhotka:
I think so?

I think the DoSaveAsync should work the way it did prior to this change: https://github.com/MarimerLLC/csla/blob/3b9b37603ead6e94bba8126aa65d7833fa8e9fe8/Source/Csla.Blazor/ViewModel.cs#L148

But with the Apply/Cancel Edit functionality added.

And yes, I think the Xaml code should work this way also.

russblair:
1. I never really understood why the Clone() command was used in the SaveAsync.  It seemed to me we could just call the ApplyEdit.
2. In the DoSaveAsync, when would there ever be a model that supports ISavable but not IEditableBusinessObject? If that is not possible then the code could become
```
protected virtual async Task<T> DoSaveAsync()
{
  if (Model is Core.ISavable savable)
    await savable.SaveAndMergeAsync();

  return Model;
}
```
 Does this sound right to you?

rockfordlhotka:
Some of this is a historical artifact from before we had GraphMerger. 

In some scenarios with the local data portal and autoclone turned off, it was possible to end up with a corrupted business object graph as a result of a save operation. Cloning the original graph ensures that if the save operation fails, that the caller still has access to an instact (non-corrupted) object graph.

iirc, this is no longer required with GraphMerger being used, and so we avoid a whole bunch of complexity.

In terms of the interfaces, when using the standard CSLA base classes you are right, that can't happen. However, the interfaces exist for advanced scenarios where someone is implementing their own base class and does want something that can be saved, but that doesn't implement all the editable object behaviors.

russblair:
I see. Should we leave the virtual method simple and allow the edge case you just mentioned to override as required?

russblair:
Also, should we remove the [Browsable(false)] from the Xaml.Shared.ViewModeBase ManageObjectLifetime property?
```
[Browsable(false)]
[Display(AutoGenerateField = false)]
[ScaffoldColumn(false)]
public bool ManageObjectLifetime
{
  ...
}
```

rockfordlhotka:
Yes please

russblair:
I just reread your earlier post. Would you prefer I move the 
```
if (ManageObjectLifetime && Model is Core.ISupportUndo undoable)
            undoable.ApplyEdit();
```
from the SaveAsync() to the DoSaveAsync()?

rockfordlhotka:
Where is it in the Xaml implementation? We should follow that model, as we know it works.

Also, the original Blazor implementation uses GraphMerger directly, rather than calling SaveAndMergeAsync. I don't remember why that is, but perhaps we should stick with the existing implementation as we know it works.

russblair:
The implementation I'm using as an example is in Csla.Xaml.Shared ViewModelBase.cs.  I have now used the same pattern. Pushing changes in a minute.

russblair:
Misunderstood your last question. The ApplyEdit was in the SaveAsync in the Xaml ViewModelBase. I have left it there for both.

rockfordlhotka (approved):

PR #3546, opened 2023-11-09
Normalize LocalProxy by subclassing DataPortalProxy #2861

### LINKED ISSUE #2861: LocalProxy to subclass DataPortalProxy for consistency/extensibility
### Discussed in https://github.com/MarimerLLC/csla/discussions/2856

> Do you want to create an issue to have LocalProxy subclass DataPortalProxy?

### LINKED ISSUE #3556: Add OnServerComplete methods to data portal proxies
Enhance data portal proxies to notify a subclass that the data portal call has returned from the server.

* OnServerComplete - called for all returns from the logical server
* OnServerCompleteClient - called for a return from the logical server, only if the _current_ execution is on the logical client

Based on a conversation in #2856 and #3451

swegele:
Fixes #2861 by making LocalProxy subclass DataPortalProxy so it is like the other 3 proxies; Also adds a virtual method called when switching appcontext back to client which allows dev to inherit from LocalProxy giving a hook to watch for salient things "coming back" to client from logical server;

Closes #3556

### DIFF
--- a/Source/Csla/DataPortalClient/DataPortalProxy.cs
+++ b/Source/Csla/DataPortalClient/DataPortalProxy.cs
@@ -329,7 +329,7 @@ namespace Csla.DataPortalClient
 
     /// <summary>
     /// Method called once only after returning to the <see cref="ApplicationContext.LogicalExecutionLocations.Client"/> after a DataPortal call.
-    /// It does not get called when subsequent chained DataPortal calls get called on the server side.
+    /// It does not get called during chained DataPortal operations initiated on the server side.
     /// </summary>
     /// <param name="result">The <see cref="DataPortalResult"/> based upon <see cref="DataPortalResponse"/> that came back from server</param>
     protected virtual void ReturnedToClientSideAndPreparedResult(DataPortalResult result, Type objectType, DataPortalOperations operationType)
--- a/Source/Csla/DataPortalClient/DataPortalProxy.cs
+++ b/Source/Csla/DataPortalClient/DataPortalProxy.cs
@@ -100,6 +100,10 @@ namespace Csla.DataPortalClient
       {
         result = new DataPortalResult(ApplicationContext, null, ex);
       }
+
+      if (ApplicationContext.LogicalExecutionLocation == ApplicationContext.LogicalExecutionLocations.Client)
+        ReturnedToClientSideAndPreparedResult(result, objectType, DataPortalOperations.Create);
+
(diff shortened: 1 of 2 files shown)


swegele:
@rockfordlhotka @TheCakeMonster  - this PR is a followup attempt at the things we discussed wayyyy back in #2856.

In review we have many different well defined hooks / extensibility points to tap into on the server side of the DataPortal (logical or physical).  (example: `OnDataPortalInvoke`, `OnDataPortalInvokeComplete`, interceptors, etc.)

But on the client side we only have access to the virtual method `ConvertResponse` (by way of subclassing a proxy that inherited from `DataPortalProxy`).  Unfortunately that didn't help with `LocalProxy` which was treated differently.  That really left no easy practical consistent way to intercept client side actions.

So this PR attemtps to do several things:
- Make `LocalProxy` inherit from `DataPortalProxy` so it is consistent as Rocky suggested
- Adds a more helpful virtual method `ReturnedToClientSideAndPreparedResult` useable in all proxies returning a helpful `DataPortalResult` instead of just byte array stuff.
- Adds a very helpful virtual method specific to `LocalPortal` called `RestoringClientSideApplicationContext`

I explained this to my wife this evening like this: "Now I more easily have the ability to consistently intercept everyone returning from overseas immediately as they get off the plane at the home airport.  I can watch for important generic things.  Before I only could ask the travelers to do a self-check-in when they got all the way home."  

I hope all this makes sense. Let me know what you think.

rockfordlhotka (changes requested):
I like the extensibility point of having a virtual method when the call returns from the logical server.

I actually think there should _maybe_ be two such methods:

* `OnServerComplete`
* `OnServerCompleteClient`

The second one would really be optional, since the first one could do its own check to see if the logical execution location is now client - but I can see the convenience of having this built-in to the framework.

So what is in the current PR is basically `OnServerCompleteClient`, because it is only invoked if the logical execution location is now client. I think that's fine, but there should also be the `OnServerComplete` that is _always_ invoked when a logical server call completes (even if the code is still loigcally on the server).

rockfordlhotka (changes requested):
There is a warning about IsServerRemote being hidden - that seems wrong to me.

rockfordlhotka (changes requested):
There is a warning where LocalProxy has a method with no await in an async method. Is that a valid warning? If so, we should suppress the warning.

PR #3597, opened 2023-12-04
Added GraphMerger Tests and fix

### LINKED ISSUE #3581: Serious GraphMerger bug with 8.2.7
It looked like the `GraphMerger` problems had been resolved, but since Friday I have been trying to work this one out. I have managed to strip down an example to just two tests, one which shows the expected result using just `SaveAsync()`, and the failure with `SaveAndMergeAsync()`. The only difference between the two tests is how the save is performed:

```cs
if (merge)
    await instance.SaveAndMergeAsync();
else
    instance = await instance.SaveAsync();
```

[GraphMergerTest.zip](https://github.com/MarimerLLC/csla/files/13413737/GraphMergerTest.zip)

Open the **Test Explorer** and run `WidgetTests`. `UpdateWithMergeWidgetTest` will fail, and `UpdateWithoutMergeWidgetTest` should pass.

In the **Standard Output**, you'll see that one of the child items has been incorrectly merged in, there is a duplicate `Guid`, and one of them has been lost. It is usually the last item which is a duplicate of the second.

This is a show stopper for us currently. I could roll back to 6.2.2, but this would require a fair bit of reworking as the Blazor view model methods changed in 7.0.0. I would really appreciate someone looking into this.

russblair:
Closes #3581


russblair:
Rocky, you may want to change the namespaces for the GraphMergerTest files to relocate them into your Csla.Tests structure.

Issue #3395, opened 2023-06-29
Fix unit tests in Csla.test


russblair:
**Is your feature request related to a problem? Please describe.**
Csla unit tests in Csla.test not passing.

**Describe the solution you'd like**
Fix tests and or code.

**Describe alternatives you've considered**
None.

**Additional context**
Rocky, please assign this ticket to me.

rockfordlhotka:
@russblair I invited you to my csla fork, because I've updated the `csla.tests` project to net8.0 and added it to the `csla.test` solution (which is run by the CI build).

My plan is to go through the failing tests one-by-one to either get them running or mark them as `Ignore` or delete them (if necessary).

I'll start from the top. If you have time this week and want to start from the bottom, that's excellent - hopefully we meet in the middle.

If you don't have time right before the Christmas/New Years holidays I fully understand!

russblair:
Hi Rocky,

 

I will try to find some time to help with this however family are in town for the holidays.

 

Regards,

Russ

rockfordlhotka:
Not all tests pass, but the remainer are mostly timing issues or use of static fields in the tests themselves. Future work is required in #3684

PR #3632, opened 2023-12-24
Work on getting unit tests running

### LINKED ISSUE #3186: Move static properties of ApplicationContext to instance properties of Options objects
We have a number of static properties on the ApplicationContext class that are used to define how the framework behaves. These are settings. When we run unit tests on the framework itself, we get some slightly inconsistent results because we are testing different scenarios that require changing the static properties. When other unit tests are in flight, this can change their behaviour in unintended ways, leading to transitory failures.

**Potential Solution**
Remove these static properties and instead keep these settings in one of the Options classes that are used for holding other settings. We can then inject these settings into the consumers who use the properties to enable a much more DI-centric, consistent experience. This will make the unit tests of the framework more consistent.

Users of the framework may also benefit from this change for the same reason, but perhaps to a lesser degree. It will be less common for people who are delivering end applications to have to run tests in all of these different scenarios. However, developers of other frameworks that extend CSLA may suffer from the same problem to a greater degree.

**Additional Impacts**
There are several configuration techniques. One of them fits a bit less well with the idea of these being properties of options; The code in the Csla.Configuration.CslaConfigurationOptions type (in the Csla\Configuration\Bind folder) is most affected by this change. We need to consider why this additional method of configuration is present, and whether this remains valuable and important, and thus how we would need to approach fixing this class.

**Additional Context**
Some of the failures we see in the Csla.Test project are as a result of this problem.

### LINKED ISSUE #3557: Rework how data portal server maps requested domain type to concrete type
In CSLA 5 the `IDataPortalActivator` type was used to map the domain type requested from the data portal into a concrete type. For example, the client might ask for an `IPersonEdit` and get back a `PersonEdit` instance.

In CSLA 8 I think there should be a dictionary of types for mapping that is part of the data portal server configuration. 

Then, in `SimpleDataPortal` this dictionary can be [used to map ](https://github.com/MarimerLLC/csla/blob/384da1f8359c675aa7b40b85984e5dd651c2412f/Source/Csla/Server/SimpleDataPortal.cs#L56C60-L56C60)the requested `objectType` into a concrete type (if necessary).

To make this work:

* [ ] Add configuration types and API for this mapping dictionary
* [ ] Use the mapping dictionary in `SimpleDataPortal` for each of the data portal operations

### LINKED ISSUE #3631: Fix deserialization error with edit level added in BusinessBindingListBase

### LINKED ISSUE #3633: LocalProxy needs to be threadsafe
The current `LocalProxy` implementation is not threadsafe, and in high load where one `LocalProxy` instance is reused in a concurrent setting, this can result in an exception where an `IServiceProvider` is used after it has been disposed.

rockfordlhotka:
#3395 
Closes #3631 Fix deserialization edit level bug in `BusinessBindingListBase`
Closes #3633 Fix issue with `LocalProxy` and concurrency
Closes #3557 Rework how data portal server maps requested domain type to concrete type
Closes #3186 Move (most) static properties out of `ApplicationContext` into configuration options types


rockfordlhotka:
@russblair Here's my current PR so you can track my progress.

I'm working from the top down, and just got ChildChanged tests to run (and everything above that point). Looks like DataAnnotations are next.

PR #3636, opened 2023-12-27
Updates to CSLA

### LINKED ISSUE #3338: Change CslaModelBinder to use CheckRulesAsync instead of CheckRules
What is happening here, is that the MVC data binding creates the `RoleEdit` instance and loads the properties of the object. That runs the rules as the properties are set. 

When I say "MVC data binding", in reality this is handled by a CSLA-specific implementation in [CslaModelBinder](https://github.com/MarimerLLC/csla/blob/main/Source/Csla.Web.Mvc.Shared/CslaModelBinder.cs).

Looking at the implementation, I can see where it might be wise to rework this to call `CheckRulesAsync` instead of `CheckRules` and entirely abstract the issue into the binder. This change would probably address your issue without any special code in any individual controller.

My recommendation to you is to take a copy of `CslaModelBinder`, switch the `CheckRules` call, and configure aspnetcore to use this binder instead of the original.

_Originally posted by @rockfordlhotka in https://github.com/MarimerLLC/csla/discussions/3336#discussioncomment-5887531_

### LINKED ISSUE #3558: Remove methods/types marked as Obsolete
There are some types and methods that were marked as Obsolete in earlier versions of CSLA. Those should be removed in the CSLA 8 release.

### LINKED ISSUE #3635: IDataPortalCache interface won't work as intended (and is uncompilable in an actual async context)
**Describe the bug**
The current implementation of [`IDataPortalCache`](Source/Csla/DataPortalClient/IDataPortalCache.cs) won't work in an actual async scenario. 

The problem is:
`Task Foo(p1, p2, out string p3)` will _compile_  just fine. But as soon as you try to implement it with `async Task` the compiler _won't_ compile because out (and in or ref) parameters are illegal in an `async` declared method( c.f.  https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/async#return-types first paragraph)

Further more: Is it intended that the cache will actually sequentialize _all_ access to the server when objects are cacheable? (Given we are in one DI scope typically Blazor or WPF)

rockfordlhotka:
Closes #3558 
Closes #3635 
#3186 
Closes #3338

### DIFF
--- a/Source/Csla.Web.Mvc.Shared/CslaModelBinder.cs
+++ b/Source/Csla.Web.Mvc.Shared/CslaModelBinder.cs
@@ -30,7 +30,7 @@ namespace Csla.Web.Mvc
     /// Bind the form data to a new instance of an IBusinessBase object.
     /// </summary>
     /// <param name="bindingContext">Binding context</param>
-    public Task BindModelAsync(ModelBindingContext bindingContext)
+    public async Task BindModelAsync(ModelBindingContext bindingContext)
     {
       ApplicationContext = (ApplicationContext)bindingContext.HttpContext.RequestServices.GetService(typeof(ApplicationContext));
       if (bindingContext == null)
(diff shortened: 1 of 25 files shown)


mtavares628 (approved):
This looks good

PR #3641, opened 2023-12-27
Implement .NET 8 Blazor state management

### LINKED ISSUE #3596: Support LocalContext and ClientContext with the Blazor 8 auto-render model
The new render modes in .NET 8 Blazor are complex when it comes to managing per-user state. In CSLA this includes `LocalContext` and `ClientContext`, which are expected to remain consistent and available throughout the lifetime of the application.

The [Per-User Blazor 8 State](https://blog.lhotka.net/2023/11/28/Per-User-Blazor-8-State) blog post documents the issue and a solution.

That solution needs to be adapted and incorporated into CSLA itself to provide (hopefully) transparent support for consistent state management in CSLA Blazor apps.

rockfordlhotka:
Closes #3596

### DIFF
--- a/Source/Csla.AspNetCore/Blazor/State/SessionIdManager.cs
+++ b/Source/Csla.AspNetCore/Blazor/State/SessionIdManager.cs
@@ -6,7 +6,6 @@
 // <summary>Manages the per-user id value</summary>
 //-----------------------------------------------------------------------
 using System;
-using System.Threading.Tasks;
 using Csla.State;
 using Microsoft.AspNetCore.Http;
 
@@ -32,7 +31,7 @@ namespace Csla.Blazor.State
     /// maintained in a browser cookie for the
     /// current user.
     /// </remarks>
-    public Task<string?> GetSessionId()
+    public string? GetSessionId()
     {
       const string sessionIdName = "cslaSessionId"; // must match Csla.Blazor.WebAssembly.State value
       var httpContext = HttpContextAccessor.HttpContext;
@@ -54,7 +53,7 @@ namespace Csla.Blazor.State
       {
         throw new InvalidOperationException("HttpContext == null");
       }
-      return Task.FromResult(result);
+      return result;
     }
   }
 }
--- a/Source/Csla.AspNetCore/Blazor/State/SessionIdManager.cs
+++ b/Source/Csla.AspNetCore/Blazor/State/SessionIdManager.cs
@@ -33,7 +33,7 @@ namespace Csla.Blazor.State
     /// </remarks>
     public string? GetSessionId()
     {
-      const string sessionIdName = "cslaSessionId"; // must match Csla.Blazor.WebAssembly.State value
+      const string sessionIdName = "cslaSessionId";
       var httpContext = HttpContextAccessor.HttpContext;
       string? result;
 
--- a/Source/Csla.AspNetCore/Blazor/State/SessionIdManager.cs
+++ b/Source/Csla.AspNetCore/Blazor/State/SessionIdManager.cs
@@ -1,0 +1,61 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="SessionIdManager.cs" company="Marimer LLC">
+//     Copyright (c) Marimer LLC. All rights reserved.
+//     Website: https://cslanet.com
+// </copyright>
+// <summary>Manages the per-user id value</summary>
+//-----------------------------------------------------------------------
+using System;
+using System.Threading.Tasks;
+using Csla.State;
+using Microsoft.AspNetCore.Http;
+
+#nullable enable
+namespace Csla.Blazor.State
+{
+  /// <summary>
+  /// Manages the per-user id value
+  /// for state management using a
+  /// browser cookie and Guid value.
+  /// </summary>
+  /// <param name="httpContextAccessor"></param>
+  public class SessionIdManager(IHttpContextAccessor httpContextAccessor) : ISessionIdManager
+  {
+    private readonly IHttpContextAccessor HttpContextAccessor = httpContextAccessor;
+
+    /// <summary>
+    /// Gets the per-user id value
+    /// for the current user state.
+    /// </summary>
+    /// <remarks>
+    /// Id is a Guid value. The id value is
+    /// maintained in a browser cookie for the
+    /// current user.
+    /// </remarks>
+    public Task<string?> GetSessionId()
+    {
+      const string sessionIdName = "cslaSessionId"; // must match Csla.Blazor.WebAssembly.State value
+      var httpContext = HttpContextAccessor.HttpContext;
+      string? result;
+
+      if (httpContext != null)
+      {
+        if (httpContext.Request.Cookies.ContainsKey(sessionIdName))
+        {
+          result = httpContext.Request.Cookies[sessionIdName];
+        }
+        else
+        {
+          result = Guid.NewGuid().ToString();
+          httpContext.Response.Cookies.Append(sessionIdName, result);
+        }
+      }
+      else
+      {
+        throw new InvalidOperationException("HttpContext == null");
+      }
+      return Task.FromResult(result);
+    }
+  }
+}
+#nullable disable
--- a/Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
+++ b/Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
@@ -45,7 +45,6 @@ namespace Csla.Blazor.State
         var key = _sessionIdManager.GetSessionId();
         session.SessionId = key;
         Replace(session, _sessions[key]);
-        _sessions[key].IsCheckedOut = false;
       }
     }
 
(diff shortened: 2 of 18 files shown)


ossendorf-at-hoelscher:
Hi. Would you mind comments and questions from me to this pr changes?

rockfordlhotka:
I love feedback!

ossendorf-at-hoelscher:
> I love feedback!
> 
> @ossendorf-at-hoelscher if you'd like to be able to use the PR feedback tools in GitHub, please send me a signed contributor agreement and I'll add you.

I'll discuss this with my superior if I can get some worktime for such things. Otherwise I'll send one from my private account 😄

Source/Csla.AspNetCore/Blazor/State/SessionIdManager.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <summary>
+    /// Gets the per-user id value
+    /// for the current user state.
+    /// </summary>
+    /// <remarks>
+    /// Id is a Guid value. The id value is
+    /// maintained in a browser cookie for the
+    /// current user.
+    /// </remarks>
+    public string? GetSessionId()
```
ossendorf-at-hoelscher, line 34: Wouldn't it be better to return `Guid?` here so calling code doesn't need to convert to a Guid everytime?

Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
```diff
@@ -0,0 +1,81 @@
...
+namespace Csla.Blazor.State
+{
+  /// <summary>
+  /// Manages all user session data for a given
+  /// root DI container.
+  /// </summary>
+  /// <param name="sessionIdManager"></param>
+  public class SessionManager(ISessionIdManager sessionIdManager) : ISessionManager
+  {
+    private readonly Dictionary<string, Session> _sessions = [];
```
ossendorf-at-hoelscher, line 22: Isn't this instance always be on the server side? If so `Dictionary` is not thread-safe. Use the `ConcurrentDictionary` instead.

Source/Csla.Blazor/State/StateManager.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+      }
+    }
+
+    /// <summary>
+    /// Waits for state to be available as it is transferred
+    /// from Blazor wasm to the web server.
+    /// </summary>
+    /// <param name="timeout">Milliseconds to wait before timing out</param>
+    private async Task WaitForState(int timeout)
```
ossendorf-at-hoelscher, line 82: The description does not match the actual implemention. (IMHO I would prefer a TimeSpan over an int to make the intent more clear and no need to know in which unit of time it's actually used).
Timeout is used as a counter and the delay is always 1s.

Tbh: I would rewrite this to use a `TaskCompletionSource` and either the Timeout kicks in or the sessions `ICheckedOut` changes. IIRC the session provides a property changed event, no?
So you can do something like this:
```csharp
var  session = _sessionManager.GetSession();
var tcs = new TaskCompletionSource();
try
{
    session.PropertyChanged += Handler;
    await tcs.Task.WaitAsync(<TimeOutAsTimeSpan>); // throws TimeoutException when Timeout hit
}
finally
{
    session.PropertyChanged -= Handler;
}

void Handler(<CorrectParamsForPropertyChangedHere>)
{
    switch(e.PropertyName)
    {
        case nameof(session.IsCheckedOut):
            if(!session.IsCheckedOut)
            {
                tcs.SetResult();
            }
            break;
    }
}
```

Source/Csla.AspNetCore/Blazor/State/SessionIdManager.cs
rockfordlhotka: I'm creating the API to be neutral about how someone might implement the id value. I'm using a Guid, but someone might use a primary key from a database table or something - hard to predict what someone might do.

Source/Csla.Blazor/State/StateManager.cs
rockfordlhotka: Hmm. I tried this, but it blocks the thread, preventing the wasm page from disposing, and so results in a deadlock. The spinlock I used by waiting for `IsCheckedOut` to change doesn't block the thread, and therefore allows the wasm page to dispose - which results in the call to the controller to update the server state and unlock `Session`.

rockfordlhotka: Oh, sure, now my _original_ implementation is locking too. Damn!

rockfordlhotka:
@ossendorf-at-hoelscher I made quite a few changes - some based on your feedback.

I _had_ it working, and then did something that broke the transition from wasm back to server. And I don't know what I did to break it 😢

Some observations.

1. I can't use `ConcurrentDictionary` because that appears to do a lock that blocks wasm->server transition
2. Using a tcs also seemed to block the wasm->server transition
3. Of course, 1 and 2 are now suspect, because it is currently broken anyway!
4. I'd forgotten that the initial landing page must initialize the cookie, and so state is unavailable for the site's initial landing page. I've now added properties that can be used to determine whether the state is actually available (via `StateManager`)

I don't know how deeply you (or anyone else) wants to invest in helping to get this working, but just in case, I have pushed my test project to GitHub:

https://github.com/rockfordlhotka/CslaBlazorApp

The solution directly references the `Csla`, `Csla.AspNetCore`, `Csla.Blazor`, and `Csla.Blazor.WebAssembly` projects. To help load the solution, here's my relative directory structure:

```
s:\src\CslaBlazorApp
s:\src\rdl\csla
```

If you clone those two repos into that folder structure, the solution should just load and run.

Unfortunately, right now, navigating from the Counter page (once it reloads as wasm) back to either server-rendered page will result in a timeout because the Counter page's `Dispose` method isn't invoking asynchronously like it should. It is being blocked by the initialize process of the server pages.

I can't figure out why right now. My [Blazor8State](https://github.com/rockfordlhotka/Blazor8State) app is implemented basically the same exact way, and it works fine (as did the CslaBlazorApp at one point in here).

StefanOssendorf:
@rockfordlhotka A short analysis shows that your `session.IsCheckedOut` does not transition from `true` (used in WASM) to `false` (used on Server) when you have such a switch between pages. Is this bool really necessary?

StefanOssendorf:
I suppose I should have read your comment more carefully 😅.

The Counters `Dispose` method is invoked when navigating from Counter -> Weather. But the updated session (on server) is never(?) updated on the client side. So you run into the timeout. Furthermore it's not a deadlock. It's just that no state change comes back after updating it on the server.

On that regard: You should change the Blazor StateManager from `void SaveState` to `Task SaveState` and await the Task of `SendSession()`. Otherwise you get timing problems which will be hard to track down.

rockfordlhotka:
The goal is to support LocalContext, and the developer can change those values as they choose. So any changes made in wasm should be available on the server (and visa versa).

With my Blazor8State app I was able to make this work using the ischeckedout flag and the Dispose in the wasm page.

I _momentarily_ had it working in CSLA as well, and really wish I'd done a git commit at that point!

rockfordlhotka:
> Furthermore it's not a deadlock

Sadly it really is. The wasm page isn't disposed because the server page is blocking the transition.

If the server page doesn't block, then dispose is called, the state is updated on the server, and the while loop exits.

StefanOssendorf:
Do we have different before?

Just to make sure: when I start your app with F5 and navigate away from
counter. I get always the timeout exception. The WASM page gets disposed
after the exception happens. So the spinlock is executing and not
blocked/dead.

Btw: happy new year 😄

Rockford Lhotka ***@***.***> schrieb am Mo., 1. Jan. 2024,
22:46:

rockfordlhotka:
Now you see why I added the timeout, otherwise the browser is permanently locked.

Run the Blazor8State app to see how it works, and why I'm puzzled that the csla app fails.

Source/Csla.Blazor/State/StateManager.cs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /// <summary>
+    /// Waits for state to be available as it is transferred
+    /// from Blazor wasm to the web server.
+    /// </summary>
+    /// <param name="timeout">Time to wait before timing out</param>
+    private async Task WaitForState(TimeSpan timeout)
+    {
+      var session = _sessionManager.GetSession();
```
ossendorf-at-hoelscher, line 140: Could this be a problem? Here gets a _new_ session created and not the already existing one reused.

rockfordlhotka:
I figured it out! Finally!

I was meticulously comparing my working Blazor8State code to the not working CslaBlazorApp code. Line by line, undoing all sorts of nice improvements, etc.

Finally realized that the server-static page in the working code is streaming. That's the key, is to have the server-static page be streaming, so it starts rendering and thus releases the wasm page to dispose.

StefanOssendorf:
Great. So back to implementing the nice improvments :D

StefanOssendorf:
Since the `ConcurrentDictionary`/ `TaskCompletionSource` wasn't the problem wouldn't it be better to try them now (with a working solution)?

Issue #3639, opened 2023-12-27
Update samples for CSLA 8


rockfordlhotka:

ctwoodwa:
Eagerly waiting with bated breathe.

PR #3644, opened 2024-01-03
Bug fixes for Blazor state, rewrite BlazorExample for .NET 8

rockfordlhotka:
#3596 
#3639 - specifically BlazorExample


Source/Csla.Blazor/State/StateManager.cs
```diff
@@ -74,7 +74,7 @@ private async Task<Session> WaitForState(TimeSpan timeout)
         var tcs = new TaskCompletionSource();
         session.PropertyChanged += (s, e) =>
         {
-          if (e.PropertyName == "IsCheckedOut" && !session.IsCheckedOut)
+          if (e.PropertyName == "IsCheckedOut" && !session.IsCheckedOut && !tcs.Task.IsCompleted)
```
ossendorf-at-hoelscher, line 77: I think instead of this you should propery unregister the delegate.
A local function with a try{}finally{} is the right way. I'm not sure if this otherwise will lead to a memory leak due to the closure.

PR #3667, opened 2024-01-31
Support .NET 7 context behavior

### LINKED ISSUE #3657: Support Blazor 7 state management
The new CSLA 8 state management capability doesn't work in .NET 7, and so it is necessary to continue to provide the old Blazor 7 context manager as an option for apps still targeting Blazor 7.

This is via feedback from @mtavares628

rockfordlhotka:
Closes #3657

### DIFF
--- a/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
+++ b/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
@@ -1,0 +1,220 @@
+﻿#if NET5_0_OR_GREATER
+//-----------------------------------------------------------------------
+// <copyright file="ApplicationContextManagerInMemory.cs" company="Marimer LLC">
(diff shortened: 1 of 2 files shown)


swegele:
@rockfordlhotka  I'm a little confused as to what the new `ApplicationContextManagerInMemory` is there for (maybe the comments are confusing because it references using `HttpContextAccessor` but it doesn't actually use that unless I'm missing something).  

If a circuit exists then we shouldn't be using HttpContextAccessor because that is not reliable as per MS guidance?

rockfordlhotka:
Yeah, the comments need updating for sure!

The in-memory type I just added is actually the _old_ context manager from Blazor 6 and 7 (the comments were wrong then too). I just re-added it with a new type name because the _new_ context manager for Blazor 8 doesn't work with older versions of Blazor.

So this is really a temporary thing, until Blazor 6 and 7 go out of support, and then this can be removed (again).

PR #3677, opened 2024-02-06
Fix spin waiting for rule completion at the server.

### LINKED ISSUE #3676: RevalidatingInterceptor is a spin wait and not a short waiting
**Describe the bug**
The implementation at https://github.com/MarimerLLC/csla/blob/bd821511e7a420ee6c90a535adb7116674cbc6eb/Source/Csla/Server/Interceptors/ServerSide/RevalidatingInterceptor.cs#L90 does not await the `Task.Delay(1)` to "wait" for the completion of the async rules. That means it's spinning this while-loop at full speed generating (probably) a lot of tasks which are ignored immediately.

**Version and Platform**
CSLA version: 7.0.3
OS: Windows
Platform: WPF, ASP.NET Core,

**Code that Fails**
[Provide the failing code, ideally an isolated repro of the issue you are encountering.](https://github.com/MarimerLLC/csla/blob/bd821511e7a420ee6c90a535adb7116674cbc6eb/Source/Csla/Server/Interceptors/ServerSide/RevalidatingInterceptor.cs#L90)

**Additional context**
To fix this issue it's necessary to make the whole Interceptor chain async aware.

StefanOssendorf:
Fixes #3676 

Breaking changes:
- IInterceptDataPortal.Initialize() changed from void to Task
- InterceptorManager.Initialize() changed from void to Task
- RevalidatingInterceptor (fix IInterceptDataPortal change)

### DIFF
--- a/Source/Csla/Server/DataPortal.cs
+++ b/Source/Csla/Server/DataPortal.cs
@@ -127,7 +127,7 @@ namespace Csla.Server
 
         AuthorizeRequest(new AuthorizeRequest(objectType, criteria, DataPortalOperations.Create));
 
-        await Initialize(new InterceptArgs { ObjectType = objectType, Parameter = criteria, Operation = DataPortalOperations.Create, IsSync = isSync });
+        await InitializeAsync(new InterceptArgs { ObjectType = objectType, Parameter = criteria, Operation = DataPortalOperations.Create, IsSync = isSync });
 
         DataPortalResult result;
         DataPortalMethodInfo method;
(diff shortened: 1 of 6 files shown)


Source/Csla/Server/DataPortal.cs
```diff
@@ -127,7 +127,7 @@ private Reflection.ServiceProviderMethodCaller ServiceProviderMethodCaller
 
         AuthorizeRequest(new AuthorizeRequest(objectType, criteria, DataPortalOperations.Create));
 
-        Initialize(new InterceptArgs { ObjectType = objectType, Parameter = criteria, Operation = DataPortalOperations.Create, IsSync = isSync });
+        await Initialize(new InterceptArgs { ObjectType = objectType, Parameter = criteria, Operation = DataPortalOperations.Create, IsSync = isSync });
```
rockfordlhotka, line 130: If we're going to have a breaking change anyway, let's change this to `InitializeAsync` to be consistent with other CSLA code.

StefanOssendorf: > If we're going to have a breaking change anyway, let's change this to `InitializeAsync` to be consistent with other CSLA code.

Only for the internal method or all changed methods?

rockfordlhotka: I think the interface (public) changed to return a `Task` right? So that method name should change, and then presumably all the non-public ones too?

StefanOssendorf: The public and some internal methods needed to change, too. So I'll just change all touched methods and add the Async suffix.

StefanOssendorf: Done. Found some missed Tests which did not await the change which is now fixed, too.

PR #3732, opened 2024-03-13
Add CascadeIfDirty property to make optimization opt-in

### LINKED ISSUE #3721: Enhance rule execution behavior to minimize breaking change impact
There is a breaking change in CSLA 8 from issue #3616, which is important and necessary.

However, there may be ways to mitigate some of the impact of the change. There's a healthy thread about this: https://github.com/MarimerLLC/csla/discussions/3720

rockfordlhotka:
Fixes #3721

### DIFF
--- a/Source/Csla/Rules/BusinessRuleBase.cs
+++ b/Source/Csla/Rules/BusinessRuleBase.cs
@@ -24,6 +24,12 @@ namespace Csla.Rules
     private RuleUri _ruleUri;
 
     /// <summary>
+    /// If true, rule will only cascade if the primary property
+    /// is dirty.
+    /// </summary>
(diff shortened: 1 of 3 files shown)


rockfordlhotka:
@lwmorris067 and @hurcane - here's the PR for my proposed fix to the breaking change around running rules.

rockfordlhotka:
At this point I am strongly leaning toward putting the ideas @hurcane documented into the CSLA 9 release. I think it is a fantastic change, but I'd like to allow time to make any deep changes required, get thorough testing (because the impact may be large), etc. I'm concerned that there's too much time pressure right now to get CSLA 8 out (so we support Blazor 8).

PR #3742, opened 2024-03-17
Fixes #1376 (AllowSynchronousIO = true now obsolete)

### LINKED ISSUE #1376: MobileFormatter and its reader/writer should be async
A breaking change in ASP.NET Core 3 means that extra config is required if any dependent libraries use synchronous operations for IO or serialization. Right now `MobileFormatter` (more accurately its reader/writer pair) are synchronous.

These should be updated to be async to avoid the extra config, and to potentially improve throughput.

StefanOssendorf:
Fixes #1376 

* Avoid sync IO through asyn reading into a memory stream which will be used by the (de)serializer
* Add integration test to verify that it works as expected

I added a basic `TestServer` to simulate HTTP calls in memory. I created a factory to make it possible to re-use it for future tests.

One thing is weird: The newly added test works when executed alone but when I execute all tests via the Test Runner the test goes red. I think there's maybe something static which interferes with running all tests in succession. For now I ignored that problem so here is the PR :-)

### DIFF
--- a/Source/Csla.TestHelpers/Csla.TestHelpers.csproj
+++ b/Source/Csla.TestHelpers/Csla.TestHelpers.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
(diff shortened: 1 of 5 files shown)


StefanOssendorf:
Note: I plan on backporting it to 7.x

rockfordlhotka:
I am holding this PR until after the 8.0.0 release, so it'll go into 8.1.0

Issue #2551, opened 2021-10-21
Update ProjectTracker sample for CSLA 9


rockfordlhotka:
Narrow the scope to a Blazor UI - other UIs can be added later.

rockfordlhotka:
Includes #2813

rockfordlhotka:
The Project side of the app works, and I'm going to call it good for 8.0.0 so it can be released.

PR #3743, opened 2024-03-21
Updates to ProjectTracker

### LINKED ISSUE #3741: Project edit page stuck at "Loading Blazor wasm runtime..."
When the ProjectTracker runs for the very first time or when the page's browser storage is cleared of the wasm runtime and when going to a project details page, the editproject page is stuck at displaying the message "Loading Blazor wasm runtime...".

The page needs to be refreshed to display the page properly. 

Is the expectation that the page will dynamically display the contents when the page finishes downloading the Blazor wasm runtime in the background?

rockfordlhotka:
Closes #3741 Fix rendering issue with project edit page

#2551


ossendorf-at-hoelscher:
When you are at it 😅. Any reason the whole click-handlers are `async void` instead of `async Task`?

PR #3747, opened 2024-03-25
Remove session checked-out concept

### LINKED ISSUE #3740: Blazor StateManager doesn't work with multiple wasm-interactive tabs open in a browser
When the Project Tracker app is running, open a new browser tab with the url of the projects list page and I get this error:

System.TimeoutException: The operation has timed out.

### LINKED ISSUE #3748: Remove hard-coded use of MobileFormatter
There are a number of places where I've been lazy and directly used MobileFormatter rather than getting the formatter via configuration. Those should be fixed.

rockfordlhotka:
Closes #3740 
Closes #3748

#2551

### DIFF
--- a/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
+++ b/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
@@ -41,8 +41,6 @@ namespace Csla.AspNetCore.Blazor
       InitializeUser();
     }
 
-    private ContextDictionary LocalContext { get; set; }
-    private ContextDictionary ClientContext { get; set; }
     private IPrincipal CurrentPrincipal { get; set; }
     private readonly ClaimsPrincipal UnauthenticatedPrincipal = new();
     private bool disposedValue;
@@ -187,18 +185,20 @@ namespace Csla.AspNetCore.Blazor
     /// </summary>
     public ContextDictionary GetLocalContext()
     {
-      if (LocalContext is null)
+      ContextDictionary localContext;
+      var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
+      var session = sessionManager.GetSession();
+      session.TryGetValue("localContext", out var result);
+      if (result is ContextDictionary context)
       {
-        var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
-        var session = sessionManager.GetSession();
-        session.TryGetValue("localContext", out var result);
-        if (result is ContextDictionary context)
-          LocalContext = context;
-        else
-          LocalContext = [];
-        SetLocalContext(LocalContext);
+        localContext = context;
       }
-      return LocalContext;
+      else
+      {
+        localContext = [];
+        SetLocalContext(localContext);
+      }
+      return localContext;
     }
 
     /// <summary>
@@ -207,7 +207,6 @@ namespace Csla.AspNetCore.Blazor
     /// <param name="localContext">Local context.</param>
     public void SetLocalContext(ContextDictionary localContext)
     {
-      LocalContext = localContext;
       var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
       var session = sessionManager.GetSession();
       session["localContext"] = localContext;
@@ -219,18 +218,20 @@ namespace Csla.AspNetCore.Blazor
     /// <param name="executionLocation"></param>
     public ContextDictionary GetClientContext(ApplicationContext.ExecutionLocations executionLocation)
     {
-      if (ClientContext is null)
+      ContextDictionary clientContext;
+      var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
+      var session = sessionManager.GetSession();
+      session.TryGetValue("clientContext", out var result);
+      if (result is ContextDictionary context)
       {
-        var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
-        var session = sessionManager.GetSession();
-        session.TryGetValue("clientContext", out var result);
-        if (result is ContextDictionary context)
-          ClientContext = context;
-        else
-          ClientContext = [];
-        SetClientContext(ClientContext, ApplicationContext.ExecutionLocation);
+        clientContext = context;
       }
-      return ClientContext;
+      else
+      {
+        clientContext = [];
+        SetClientContext(clientContext, ApplicationContext.ExecutionLocation);
+      }
+      return clientContext;
     }
 
     /// <summary>
@@ -240,7 +241,6 @@ namespace Csla.AspNetCore.Blazor
     /// <param name="executionLocation"></param>
     public void SetClientContext(ContextDictionary clientContext, ApplicationContext.ExecutionLocations executionLocation)
     {
-      ClientContext = clientContext;
       var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
       var session = sessionManager.GetSession();
       session["clientContext"] = clientContext;
--- a/Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
+++ b/Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
@@ -33,7 +33,7 @@ namespace Csla.Blazor.State
       if (!_sessions.ContainsKey(key))
         _sessions.TryAdd(key, []);
       result = _sessions[key];
-      result.LastTouched = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
+      result.Touch();
       return result;
     }
 
@@ -48,7 +48,7 @@ namespace Csla.Blazor.State
         var key = _sessionIdManager.GetSessionId();
         var existingSession = _sessions[key];
         Replace(newSession, existingSession);
-        existingSession.LastTouched = newSession.LastTouched;
+        existingSession.Touch();
       }
     }
 
--- a/Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
+++ b/Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
@@ -66,26 +66,6 @@ namespace Csla.Blazor.State
     }
 
     /// <summary>
-    /// Retrieves the current user's session from
-    /// the web server to the wasm client.
-    /// </summary>
-    public Task<Session> RetrieveSession() => throw new NotSupportedException();
-
-    /// <summary>
-    /// Sends the current user's session from
-    /// the wasm client to the web server.
-    /// </summary>
-    /// <returns></returns>
-    public Task SendSession()
-    {
-      var key = _sessionIdManager.GetSessionId();
-      var existingSession = _sessions[key];
-      if (existingSession != null)
-        _sessions[key].LastTouched = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
-      return Task.CompletedTask;
-    }
-
-    /// <summary>
     /// Remove all expired session data.
     /// </summary>
     /// <param name="expiration">Expiration duration</param>
@@ -99,5 +79,9 @@ namespace Csla.Blazor.State
       foreach (var key in toRemove)
         _sessions.TryRemove(key, out var x);
     }
+
+    Task<Session> ISessionManager.RetrieveSession() => throw new NotImplementedException();
+    Session ISessionManager.GetCachedSession() => throw new NotImplementedException();
+    Task ISessionManager.SendSession() => throw new NotImplementedException();
   }
 }
--- a/Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
+++ b/Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
@@ -33,7 +33,7 @@ namespace Csla.Blazor.State
       if (!_sessions.ContainsKey(key))
         _sessions.TryAdd(key, []);
       result = _sessions[key];
-      result.LastTouched = DateTimeOffset.UtcNow;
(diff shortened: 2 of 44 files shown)


Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -187,18 +185,20 @@ private void SetHostPrincipal(Task<AuthenticationState> task)
     /// </summary>
     public ContextDictionary GetLocalContext()
     {
-      if (LocalContext is null)
+      ContextDictionary localContext;
+      var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
+      var session = sessionManager.GetSession();
+      session.TryGetValue("localContext", out var result);
```
StefanOssendorf, line 191: What about compacting this whole method to:
```csharp
if(!session.TryGetValue("localContext", out var storedContext) || storedContext is not ContextDictionary context) {
	context = [];
	SetLocalContext(context);
}

return context;
```
I don't know if a current TFM is impeding on that?

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -219,18 +218,20 @@ public void SetLocalContext(ContextDictionary localContext)
     /// <param name="executionLocation"></param>
     public ContextDictionary GetClientContext(ApplicationContext.ExecutionLocations executionLocation)
     {
-      if (ClientContext is null)
+      ContextDictionary clientContext;
+      var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
+      var session = sessionManager.GetSession();
+      session.TryGetValue("clientContext", out var result);
```
StefanOssendorf, line 224: Same suggestion like `GetLocalContext`.

Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
```diff
@@ -77,15 +76,22 @@ private static void Replace(Session newSession, Session oldSession)
     /// the wasm client to the web server.
     /// </summary>
     /// <returns></returns>
-    public Task SendSession() => throw new NotSupportedException();
+    public Task SendSession()
+    {
+      var key = _sessionIdManager.GetSessionId();
+      var existingSession = _sessions[key];
```
StefanOssendorf, line 82: What about extracting this piece of code into a method
```csharp
private void UpdateSessionLastTouched(string key) {
  var existingSession = _sessions[key];
  if (existingSession != null)
    existingSession.LastTouched = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
}
```

This would also remove the lookup of the key twice :D
And use it here and in `GetSession()`

Source/Csla.AspNetCore/Blazor/State/StateController.cs
```diff
@@ -37,23 +38,33 @@ public class StateController(ApplicationContext applicationContext, ISessionMana
...
+      else
+      {
+        var message = (SessionMessage)ApplicationContext.CreateInstanceDI(typeof(SessionMessage));
+        message.Session = session;
+        if (FlowUserIdentityToWebAssembly)
+        {
+          var principal = new CslaClaimsPrincipal(ApplicationContext.Principal);
+          message.Principal = principal;
+        }
+        var formatter = Csla.Serialization.SerializationFormatterFactory.GetFormatter(ApplicationContext);
```
StefanOssendorf, line 61: My future aspirations in work here: Instead of getting it from this static class just let it be injected into the Controller as a field.

Source/Csla.AspNetCore/Blazor/State/StateController.cs
```diff
@@ -65,25 +76,44 @@ public virtual byte[] Get()
...
-      session.IsCheckedOut = false;
       _sessionManager.UpdateSession(session);
     }
+  }
 
+  /// <summary>
+  /// Message type for communication between StateController
+  /// and the Blazor wasm client.
+  /// </summary>
+  public class StateResult
```
StefanOssendorf, line 93: Shouldn't this type be in it's own file?

Source/Csla.AspNetCore/Blazor/State/StateController.cs
```diff
@@ -65,25 +76,44 @@ public virtual byte[] Get()
...
+    /// <summary>
+    /// Gets or sets the serialized session data.
+    /// </summary>
+    public byte[] SessionData { get; set; }
+  }
+
+  /// <summary>
+  /// Result status values for StateResult.
+  /// </summary>
+  public enum ResultStatuses
```
StefanOssendorf, line 108: Shouldn't this type be in it's own file?

Source/Csla.Blazor.WebAssembly/ApplicationContextManager.cs
```diff
@@ -108,18 +105,20 @@ public virtual void SetUser(IPrincipal principal)
     /// </summary>
     public ContextDictionary GetLocalContext()
     {
-      if (LocalContext is null)
+      ContextDictionary localContext;
+      var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
+      var session = sessionManager.GetSession();
+      session.TryGetValue("localContext", out var result);
```
StefanOssendorf, line 111: Same suggestion as in SessionManager here.

Source/Csla.Blazor.WebAssembly/ApplicationContextManager.cs
```diff
@@ -140,18 +138,20 @@ public void SetLocalContext(ContextDictionary localContext)
     /// <param name="executionLocation"></param>
     public ContextDictionary GetClientContext(ApplicationContext.ExecutionLocations executionLocation)
     {
-      if (ClientContext is null)
+      ContextDictionary clientContext;
+      var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
+      var session = sessionManager.GetSession();
+      session.TryGetValue("clientContext", out var result);
```
StefanOssendorf, line 144: Same suggestion as in SessionManager here.

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -40,35 +40,44 @@ public class SessionManager(
     /// </summary>
     public Session GetSession()
     {
-      _session ??= [];
-      _session.LastTouched = DateTimeOffset.UtcNow;
+      if (_session == null)
+      {
+        _session ??= [];
```
StefanOssendorf, line 45: This should not be race condition so a simple `_session = [];` should suffice.

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -40,35 +40,44 @@ public class SessionManager(
...
     /// <summary>
     /// Updates the current user's session 
-    /// data locally and on the web server.
+    /// data locally.
     /// </summary>
     /// <param name="session">Current user session data</param>
     public void UpdateSession(Session session)
     {
-      _session.LastTouched = DateTimeOffset.UtcNow;
+      _session.LastTouched = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
```
StefanOssendorf, line 58: You are updating here the old session. Seems like a bug? Maybe you could extract the method I proposed in SessionManager with an overload to unify this kind of updating the `LastTouched` field.

Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
StefanOssendorf: Idea: Get fancy and add `internal void UpdateLastTouchedToNow()` method to the `Session` type and use that everywhere?

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -80,30 +89,30 @@ public async Task<Session> RetrieveSession()
...
     /// <summary>
     /// Sends the current user's session from
-    /// the wasm client to the web server.
+    /// the wasm client to the web server 
+    /// if SyncContextWithServer is true.
     /// </summary>
     /// <returns></returns>
     public async Task SendSession()
     {
+      _session.LastTouched = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
```
StefanOssendorf, line 112: Use the new method/approach?

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -40,35 +40,44 @@ public class SessionManager(
...
+          lastTouched = _session.LastTouched;
+        var url = $"{_options.StateControllerName}?lastTouched={lastTouched}";
+
+        var stateResult = await client.GetFromJsonAsync<StateResult>(url);
+        if (stateResult.ResultStatus == ResultStatuses.Success)
         {
-          var result = await client.GetFromJsonAsync<byte[]>(_options.StateControllerName);
-          var formatter = new MobileFormatter(ApplicationContext);
-          var buffer = new MemoryStream(result)
+          var formatter = SerializationFormatterFactory.GetFormatter(ApplicationContext);
```
StefanOssendorf, line 79: Inject the serializer directly instead of going through the static factory (which get's it from the DI anyway).

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -80,30 +89,30 @@ public async Task<Session> RetrieveSession()
...
     /// </summary>
     /// <returns></returns>
     public async Task SendSession()
     {
+      _session.LastTouched = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
       if (_options.SyncContextWithServer)
       {
-        _session.LastTouched = DateTimeOffset.UtcNow;
-        var formatter = new MobileFormatter(ApplicationContext);
+        var formatter = SerializationFormatterFactory.GetFormatter(ApplicationContext);
```
StefanOssendorf, line 115: If serializer is provided via ctor-DI  replace usage

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -117,4 +126,35 @@ public async Task SendSession()
     /// <param name="expiration">Expiration duration</param>
     public void PurgeSessions(TimeSpan expiration) => throw new NotSupportedException();
   }
+
+  /// <summary>
+  /// Message type for communication between StateController
+  /// and the Blazor wasm client.
+  /// </summary>
+  public class StateResult
```
StefanOssendorf, line 134: Should this type be in it's own file?

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -117,4 +126,35 @@ public async Task SendSession()
...
+    /// <summary>
+    /// Gets or sets the serialized session data.
+    /// </summary>
+    public byte[] SessionData { get; set; }
+  }
+
+  /// <summary>
+  /// Result status values for StateResult.
+  /// </summary>
+  public enum ResultStatuses
```
StefanOssendorf, line 149: Should this type be in it's own file?

StefanOssendorf (changes requested):
Some thoughts and unification of code :)

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
rockfordlhotka: Does this increase or decrease readability? I argue that it decreases readability.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
rockfordlhotka: Does this increase or decrease readability? I argue that it decreases readability.

Source/Csla.AspNetCore/Blazor/State/StateController.cs
rockfordlhotka: Yes

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
rockfordlhotka: That's a good idea - but a bigger change because the serializer isn't currently provided as a DI service. Maybe something for a future version.

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
rockfordlhotka: Yes

rockfordlhotka:
Good feedback @StefanOssendorf - lots of changes if you want to take a look at what I've done in response.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
StefanOssendorf: @rockfordlhotka Just to answer since this is a subjective thing here.
I think it increases readability but at a second thought I would invert the if to:
```csharp
if(session.TryGetValue("localContext", out var storedContext) && storedContext is ContextDictionary context) {
  return ctx;
}

ContextDictionary context = [];
SetLocalContext(context);
return context;
```

I personally don't like/use the "declare variables at the beginning of a method regardless where it's used". So I try to declare it as close as possible to it's first usage. This method is quite short so it's not that far but it still took me two or three looks to see where the variable was declared and initialized.

PR #3774, opened 2024-04-04
Add content-type to HttpProxy requests

### LINKED ISSUE #1700: Set content-type in HttpProxy and add extensibility
Right now the `HttpProxy` implementation does not

1. Set the content-type in each request header ([making compression impossible](https://github.com/MarimerLLC/csla/discussions/1694))
1. Have an extensibility point for setting headers on the request (which seems valuable for various scenarios including security)

This work item is to set the `context-type` [when the `HttpRequestMessage` is created](https://github.com/MarimerLLC/csla/blob/master/Source/Csla.Shared/DataPortalClient/HttpProxy.cs#L417).

And this work item is to add a `virtual` method that is invoked, passing the `HttpRequestMessage` object as a parameter, so a subclass can alter the request; for example, by adding headers.

Also do the same in the [server controller base class ](https://github.com/MarimerLLC/csla/blob/master/Source/Csla.Web.Mvc.Shared/Server/Hosts/HttpPortalController.cs).

rockfordlhotka:
Closes #1700

### DIFF
--- a/Source/Csla/DataPortalClient/HttpProxy.cs
+++ b/Source/Csla/DataPortalClient/HttpProxy.cs
@@ -117,6 +117,14 @@ namespace Csla.Channels.Http
     protected virtual void SetHttpRequestHeaders(HttpRequestMessage request)
     { }
 
+    /// <summary>
+    /// Override to set headers or other properties of the
+    /// WebClient before it is sent to the server.
+    /// </summary>
+    /// <param name=client">WebClient instance</param>
+    protected virtual void SetWebClientHeaders(WebClient client)
+    { }
+
     private async Task<byte[]> CallViaHttpClient(byte[] serialized, string operation, string routingToken)
     {
       HttpClient client = GetHttpClient();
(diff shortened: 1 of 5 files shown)


Source/Csla/DataPortalClient/HttpProxy.cs
```diff
@@ -117,13 +117,22 @@ protected override async Task<byte[]> CallDataPortalServer(byte[] serialized, st
     protected virtual void SetHttpRequestHeaders(HttpRequestMessage request)
     { }
 
+    /// <summary>
+    /// Override to set headers or other properties of the
+    /// WebClient before it is sent to the server.
+    /// </summary>
+    /// <param name=client">WebClient instance</param>
```
StefanOssendorf, line 124: Missing " before client.

StefanOssendorf (dismissed):

Source/Csla/DataPortalClient/HttpProxy.cs
```diff
@@ -117,13 +117,22 @@ protected override async Task<byte[]> CallDataPortalServer(byte[] serialized, st
     protected virtual void SetHttpRequestHeaders(HttpRequestMessage request)
     { }
 
+    /// <summary>
+    /// Override to set headers or other properties of the
+    /// WebClient before it is sent to the server.
+    /// </summary>
```
StefanOssendorf, line 123: Maybe document here that this is _only_ used for the sync-CSLA-method versions and clarify in `SetHttpRequestHeader` that it's _only_ used for the async-versions?

PR #3776, opened 2024-04-04
Use WaitForIdle() method in simple data portal to wait for rule execution

### LINKED ISSUE #3755: Data portal should call WaitForIdle on root object before returning
Right now, if an object graph has async rules running on the logical server as the data portal operation completes, an exception is thrown because the graph is busy.

The data portal should, prior to the is busy check, await the root object's `WaitForIdle` method to wait until all async rules are complete (or the timeout). This is in `SimpleDataPortal`.

The `ObjectFactory` type should also expose a `WaitForIdle` helper method to make it consistent for factory code to explicitly wait for rule completion.

As per this thread: https://github.com/MarimerLLC/csla/discussions/3751#discussioncomment-8932459

StefanOssendorf:
* `WaitForIdle(TimeSpan timeout)` added to  `IDataPortalTarget`
* `WaitForIdle` implementations consolidated into `BusyHelper` static class which depends on the `INotifyBusy` interface
* `.ConfigureAwait(false)` added

Closes #3755

### DIFF
--- a/Source/Csla/Core/BusyHelper.cs
+++ b/Source/Csla/Core/BusyHelper.cs
@@ -38,9 +38,8 @@ namespace Csla.Core {
         source.BusyChanged -= ObserverForIsBusyChange;
       }
 
-      void ObserverForIsBusyChange(object sender, BusyChangedEventArgs e) {
-        // e.PropertyName == string.Empty means CSLA itself raised the busy-changed event
-        if (!source.IsBusy && !e.Busy && (string.IsNullOrEmpty(e.PropertyName) || e.PropertyName == nameof(INotifyBusy.IsBusy))) {
+      void ObserverForIsBusyChange(object sender, BusyChangedEventArgs e) {        
+        if (!source.IsBusy && !e.Busy) {
           tcs.TrySetResult(null);
         }
       }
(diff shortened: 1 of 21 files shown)


Source/Csla/Core/BusyHelper.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        if (!source.IsBusy) {
+          return;
+        }
+
+        var timeoutTask = Task.Delay(timeout);
+        var finishedTask = await Task.WhenAny(tcs.Task, timeoutTask);
+
+        if (finishedTask == timeoutTask) {
+          throw new TimeoutException($"{source.GetType().FullName}.{methodName} after {timeout}.");
```
rockfordlhotka, line 40: If we put actual words in an exception message, that message needs to come from the resources (resx) file so it can be localized.

Lately I've been trying hard to avoid the use of any words in exceptions to minimize the localization effort in that regard.

rockfordlhotka (changes requested):

Source/Csla/Core/BusyHelper.cs
StefanOssendorf: That's a good approach to avoid that nasty thing called localization :D

StefanOssendorf: So instead of the "after" just a "-" (dash) to include the timeout time or should I remove that completly and do it like the old implementation did?

rockfordlhotka: The dash works for me.

StefanOssendorf: Done

rockfordlhotka (approved):

PR #3793, opened 2024-04-14
Add csharp_new_line_before_open_brace=all to editorconfig

StefanOssendorf:
Add `csharp_new_line_before_open_brace = all` [New-line options](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/csharp-formatting-options#new-line-options)

I personally use the style
```csharp
public void Foo() {
}
```
instead of (the now demanded by editorconfig)
```csharp
public void Foo()
{
}
```
so this change just means a bit less clean up work (for me) and other who have other personal coding styles.

### DIFF
--- a/Source/.editorconfig
+++ b/Source/.editorconfig
@@ -4,6 +4,7 @@ indent_style = space
 indent_size = 2
 
 [*.cs]
+csharp_new_line_before_open_brace = all
 csharp_new_line_before_catch = true
 csharp_new_line_before_else = true
 csharp_new_line_before_finally = true


rockfordlhotka:
This PR makes me think of one of the first big lessons I learned in my career - I even blogged about it (in a lengthy story) years ago:

https://blog.lhotka.net/2021/01/04/How-I-Got-Into-Computers-First-Job-Edition

Consistency of code is critical in my experience. Once some code has been written, it should be impossible for someone to come back to that code later and know who wrote it - because all the code is consistent in terms of style.

Personal preferences are something people can use in personal projects, or in a project they own (where I recommend they enforce their choices too).

In a code base with lots of developers that will work on it over years (or decades), consistency is king.

PR #3806, opened 2024-04-17
use collection expressions

SimonCropp:
thoughts on using collection expressions?


rockfordlhotka:
I'm still not 100% sold on the syntax. I understand the concise value, and yet worry that it makes code less readable at a glance.

This is one of those language features that feels like "original c#" vs "new c#", and people can read/use one or the other.

I'm interested in thoughts from other folks too.

SimonCropp:
@rockfordlhotka 

for me it is so similar to an array constructor that it wasnt much of a stretch to get used to. i just think of it as "smart array constructor where i dont need to be concerned about the target type"

there are also some non trivial perf implications. for example take this

```
IReadOnlyCollection<string> x = ["a"]; 
```

it actually compiles to
```
IReadOnlyCollection<string> x = new ReadOnlySingleElementList<string>("a"); 
```

where ReadOnlySingleElementList is a special type that is optimized for storage and lookup where it knows that the usage is for a IReadOnlyCollection with a single item.

```
IReadOnlyCollection<string> x = ["a", "b"]; 
```

compiles to 

```
new ReadOnlyArray<string>(new string[2] { "a", "b" });
```

and when u change the target type collection expression will automatically pick the best type for that target type and the number of elements in that item

another one is it knows when to compile to `Enumerable.Empty` or `Array.Empty`. there were several cases in this PR where Empty was not being used. and with collection expressions u dont need to reason about it

PR #3811, opened 2024-04-17
avoid incorrect as usage in PropertyStatus.popup_Loaded

SimonCropp:
assumptions are the same. but better to have a cast exception than a null ref exception


rockfordlhotka:
@SimonCropp this PR needs a conflict resolved

SimonCropp:
resolved

PR #3812, opened 2024-04-17
remove unused variables and discard unused parameters

SimonCropp:


Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -108,7 +108,7 @@ private void AuthenticationStateProvider_AuthenticationStateChanged(Task<Authent
       }
       else
       {
-        task.ContinueWith((t) =>
+        task.ContinueWith(_ =>
```
StefanOssendorf, line 111: @SimonCropp 
Is there any difference in using `task` within the `ContinueWith` instead of the `Task` passed as parameter to it? Can there be subtle differences so we should use the parameter within the continuation to check for completion/fail etc.?

SimonCropp: @StefanOssendorf oh good point. yes we should use the parameter in this case. because of the variable closure and the lambda.

i will fix

PR #3817, opened 2024-04-17
avoid some casting by using the correct value type on a dictionary

SimonCropp:


rockfordlhotka (changes requested):
I think the type should be `IPropertyInfo`, not `PropertyInfo`?

SimonCropp:
@rockfordlhotka done

rockfordlhotka (approved):

PR #3821, opened 2024-04-17
elide simple asyncs

SimonCropp:


rockfordlhotka:
@SimonCropp I am hopefully learning something here. Do these changes create zero functional differences from the previous code? Is this new code faster, or just potentially cleaner?

StefanOssendorf:
It's more efficient in these cases as far as I saw. Here a nice blog post about that https://blog.stephencleary.com/2016/12/eliding-async-await.html

SimonCropp:
in summary (and some of these are negligible but improvements all the same)

 * no state machine = smaller dll + faster compile + faster jit + less mem to load assembly + faster assembly load
 * no state machine executed = faster execution + less memory + less garbage collection
 * IMO more readable
 
there are more opportunities for eliding. but i picked some easy and non controversial ones first

PR #3831, opened 2024-04-18
remove nunit code

SimonCropp:
i might be mistaken. but it doesnt seem that nunit is being used anywhere


rockfordlhotka:
True. We haven't used nunit in many, many years. For a long time we've kept those usings in the codebase in case we ever decided to switch back, but I don't see that happening. Good to remove them.

@SimonCropp please resolve merge issues.

SimonCropp:
resolved

PR #3845, opened 2024-04-19
remove some un-used variables

SimonCropp:


rockfordlhotka:
@SimonCropp 

Just skimming the diff here in GitHub - did you run the tests after doing the change?

(sadly a bunch of tests don't current work on the CI server, so they don't run automatically 😢 )

SimonCropp:
will close this one for now. and re apply gradually

SimonCropp:
what tests dont work on the build server?

ossendorf-at-hoelscher:
@SimonCropp At the moment 0 tests are executed during the github action. According to rocky because they didn't run there.
We have currently 2 issues with the tests.
1. If you execute all from within VS some fail even though they pass when executed individually (we assume static stuff which does not play nice with this execution)
2. Executing with a github action fails (as stated)

PR #3880, opened 2024-04-25
Fixes 3857: Remove protected setter from InputPropertes

### LINKED ISSUE #3857: Remove protected setter in BusinessRuleBase from InputProperties
As in #3855 proposed.

Remove the `protected set;` from `InputProperties` in `BusinessRuleBase`.
Common usage of lists is to initialize them only once and then add/delete/clear them as necessary. There's also a NetAnalyzer which suggests that but I can't find the number right now.

StefanOssendorf:
- Remove protected set from InputProperties.
- Fix the breaking change
- Some code cleanup (namespace, nameof() instead of magic strings)

Fixes #3857

For discussion: Would it be better to change the `List` to `HashSet` since a property should only be once within the lists? Having a property multiple times in e.g. `InputProperties` will result in an exception due to the usage of `Dictionary.Add` which will throw when a duplicate key is added. And the affected properties are made to be unique with `Distinct()` which could also be discarded if changed to `HashSet`.

### DIFF
--- a/Source/Csla.test/Fakes/Server/BusyStatus/ItemWithAsynchRule.cs
+++ b/Source/Csla.test/Fakes/Server/BusyStatus/ItemWithAsynchRule.cs
@@ -78,7 +78,7 @@ namespace Csla.Testing.Business.BusyStatus
         : base(primaryProperty)
       {
         IsAsync = true;
-        InputProperties = [primaryProperty];
+        InputProperties.Add(primaryProperty);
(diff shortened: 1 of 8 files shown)


rockfordlhotka:
@StefanOssendorf 

iirc, `HashSet` is maybe faster, but consumes more memory than a `List`. I'd want to know the consequences before making such a change.

PR #3891, opened 2024-04-27
#3802 - replace null parameter with containingType

### LINKED ISSUE #3802: Bug: ContainingType parameter isn't passed along instead null is used
https://github.com/MarimerLLC/csla/blob/881c262f695cfa02b7e936f123280aadd5fb867f/Source/Csla/PropertyInfo.cs#L104-L105

This looks like a bug to me. The third parameter `null` in `this()` is actually `Type containingType`. But why isn't the parameter passed along but instead `null` used?
Is this a bug or intended?

19bibo85:
Fixes #3802  - Replace null parameter with containingType.

### DIFF
--- a/Source/Csla.test/PropertyInfo/PropertyInfoTests.cs
+++ b/Source/Csla.test/PropertyInfo/PropertyInfoTests.cs
@@ -60,7 +60,7 @@ namespace Csla.Test.PropertyInfo
       Assert.AreEqual("x", PropertyInfoRoot.NameDefaultValueProperty.DefaultValue);
       Assert.AreEqual("x", PropertyInfoRoot.NewPropertyInfoRoot(dataPortal).NameDefaultValue);
     }
-    
+
     [TestMethod]
     public void TestStringNullDefaultValue()
     {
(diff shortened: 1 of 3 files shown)


SimonCropp:
is it possible to add a test for this?

StefanOssendorf:
And did you sign the contribution agreement and send it to Rocky?

rockfordlhotka:
> And did you sign the contribution agreement and send it to Rocky?

Yes, I have the agreement.

19bibo85:
@SimonCropp Test `TestContainingType` has been added to the `PropertyInfoTests` class.

Source/Csla.test/PropertyInfo/PropertyInfoTests.cs
```diff
@@ -69,5 +69,12 @@ public void TestStringNullDefaultValue()
       Assert.AreEqual(null, PropertyInfoRoot.StringNullDefaultValueProperty.DefaultValue);
       Assert.AreEqual(null, PropertyInfoRoot.NewPropertyInfoRoot(dataPortal).StringNullDefaultValue);
     }
+
+    [TestMethod]
+    public void TestContainingType()
```
StefanOssendorf, line 74: As already written as a comment n the PR. This test does not check the fix from `PropertyInfo.cs`.
When you revert the fix the test is still green.

StefanOssendorf (changes requested):

Source/Csla.test/PropertyInfo/PropertyInfoTests.cs
19bibo85: @StefanOssendorf  Thank you for reviewing the PR. You're right, I have just amended the `ContainingTypeProperty` using the `PropertyInfo` constructor changed. I confirm that reverting the change the test now fails. :-)

StefanOssendorf (approved):

PR #3892, opened 2024-04-28
Fix ruleSet parameter omission in AddRule overloads

### LINKED ISSUE #3803: Bug: Add lambda rule handler does not use the given ruleSet and adds the rule to the default rule set
https://github.com/MarimerLLC/csla/blob/881c262f695cfa02b7e936f123280aadd5fb867f/Source/Csla/Rules/BusinessRulesExtensions.cs#L99-L102

The `string ruleSet` parameter is not passed to the next method call which results in all rules are added into the default rule set.

kamranrauf75:
Fixes #3803 
Bug: Add lambda rule handler does not use the given ruleSet and adds the rule to the default rule set


rockfordlhotka:
I have a contributor agreement from @kamranrauf75

StefanOssendorf:
Thank you for your contribution. Could you please also add a test for this bug. Thanks again!

StefanOssendorf:
@kamranrauf75 Friendly reminder to please add tests for this PR to be merged.

kamranrauf75:
@StefanOssendorf Thank you for reviewing my pull request. I understand the importance of unit tests, but the change I've made is quite minor, and tests for such a change doesn't make much sense to me. But still I'll try to add relevant tests.

PR #3898, opened 2024-05-01
Breaking change: Blazor StateManager does not use the provided timeout parameter

### LINKED ISSUE #3800: Breaking change: Blazor StateManager does not use the provided timeout parameter
https://github.com/MarimerLLC/csla/blob/881c262f695cfa02b7e936f123280aadd5fb867f/Source/Csla.Blazor/State/StateManager.cs#L46-L52

As shown the timeout parameter is not used in this method. The method is called by the two `InitializeAsync()` methods where one provides a timeout parameters as well.

I think we have two ways to correct this:
1. Remove the timeout parameter and remove the `InitializeAsync(TimeSpan)` method.
2. We extend `ISessionManager.RetrieveSession()` with an overload which also accepts a timeout and pass it along

I think option 2 would be the best so the user can decide whether he wants to wait or not.
In this regard: Should the `SaveState` also get a timeout parameter with passing along to `SendSession()`?

luizfbicalho:
Closes #3800

### DIFF
--- a/Source/Csla.test/HttpProxy/HttpProxyExtensionsTests.cs
+++ b/Source/Csla.test/HttpProxy/HttpProxyExtensionsTests.cs
@@ -1,0 +1,37 @@
+﻿using Csla.Channels.Http;
+using Csla.Configuration;
+using Csla.TestHelpers;
+using FluentAssertions;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Csla.Test.HttpProxy;
+
+[TestClass]
+public class HttpProxyExtensionsTests
+{
+  [TestMethod]
+  public void UseHttpProxy_WhenConfiguringACustomHttpClientFactoryItShouldBeUsedWhenHttpProxyIsCreated()
+  {
+    bool hasBeenCalled = false;
+    HttpClient CustomFactory(IServiceProvider sp)
+    {
+      hasBeenCalled = true;
+      return new HttpClient();
+    }
+
+    var diContext = TestDIContextFactory.CreateContext(
+      o => o.DataPortal(
+        dp => dp.ClientSideDataPortal(
+          cdp => cdp.UseHttpProxy(
+            hp => hp.WithHttpClientFactory(CustomFactory)
+            )
+          )
+        )
+      );
+
+    _ = diContext.ServiceProvider.GetRequiredService<Csla.DataPortalClient.IDataPortalProxy>();
+
+    hasBeenCalled.Should().BeTrue();
+  }
+}
--- a/Source/Csla.test/HttpProxy/HttpProxyOptionsExtensionsTests.cs
+++ b/Source/Csla.test/HttpProxy/HttpProxyOptionsExtensionsTests.cs
@@ -1,0 +1,67 @@
+﻿using Csla.Channels.Http;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Csla.Test.HttpProxy;
+
+[TestClass]
+public class HttpProxyOptionsExtensionsTests
+{
+  [TestMethod]
+  public void WithDataPortalUrl_ShouldSetTheDataPortalUrlAsExpected()
+  {
+    var options = new HttpProxyOptions();
+    options.WithDataPortalUrl("http://localhost:1337");
+
+    var expectedOptions = new HttpProxyOptions
+    {
+      DataPortalUrl = "http://localhost:1337"
+    };
+
+    options.Should().BeEquivalentTo(expectedOptions);
+  }
+
+  [TestMethod]
+  public void UseTextForSerialization_ShouldSetTheUseTextSerializationToTrue()
+  {
+    var options = new HttpProxyOptions();
+    options.UseTextForSerialization();
+
+    var expectedOptions = new HttpProxyOptions
+    {
+      UseTextSerialization = true
+    };
+
+    options.Should().BeEquivalentTo(expectedOptions);
+  }
+
+  [TestMethod]
+  public void UseDefaultForSerialization_ShouldSetTheUseTextSerializationToFalse()
+  {
+    var options = new HttpProxyOptions();
+    options.UseDefaultForSerialization();
+
+    var expectedOptions = new HttpProxyOptions
+    {
+      UseTextSerialization = false
+    };
+
+    options.Should().BeEquivalentTo(expectedOptions);
+  }
+
+  [TestMethod]
+  public void WithHttpClientFactory_ShouldSetTheFactoryToTheProvidedFactory()
+  {
+    var options = new HttpProxyOptions();
+    options.WithHttpClientFactory(httpClientFactory);
+
+    var expectedOptions = new HttpProxyOptions
+    {
+      HttpClientFactory = httpClientFactory
+    };
+
+    options.Should().BeEquivalentTo(expectedOptions);
+    
+    static HttpClient httpClientFactory(IServiceProvider sp) => new();
+  }
+}
--- a/Source/Csla/DataPortalClient/HttpProxy.cs
+++ b/Source/Csla/DataPortalClient/HttpProxy.cs
@@ -21,6 +21,8 @@ namespace Csla.Channels.Http
   /// </summary>
   public class HttpProxy : DataPortalProxy
   {
+    private HttpClient _httpClient;
+
     /// <summary>
     /// Creates an instance of the type, initializing
     /// it to use the supplied HttpClient object and options.
@@ -42,7 +44,7 @@ namespace Csla.Channels.Http
     /// Current options for the proxy.
     /// </summary>
     protected HttpProxyOptions Options { get; set; }
-    private HttpClient _httpClient;
+
     private string VersionRoutingTag { get; set; }
 
 #nullable enable
--- a/Source/Csla/DataPortalClient/HttpProxyExtensions.cs
+++ b/Source/Csla/DataPortalClient/HttpProxyExtensions.cs
@@ -6,8 +6,8 @@
 // <summary>Implement extension methods data portal channel</summary>
 //-----------------------------------------------------------------------
 
-using System.Net.Http;
 using Csla.DataPortalClient;
+using Csla.Channels.Http;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Csla.Configuration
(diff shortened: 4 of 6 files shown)


Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -53,25 +63,35 @@ public async Task<Session> RetrieveSession()
...
+            }
+          }
+          else // NoUpdates
           {
-            provider.SetPrincipal(message.Principal);
+            _session = GetSession();
           }
         }
-        else // NoUpdates
+        catch(OperationCanceledException)
```
StefanOssendorf, line 91: Why is the `GetSession()` necessary after the timeout kicked in?

Source/Csla.Blazor/State/StateManager.cs
```diff
@@ -39,12 +39,14 @@ public Task InitializeAsync(TimeSpan timeout)
     /// Get state from cache.
     /// </summary>
     /// <param name="timeout">Time to wait before timing out</param>
-    private async Task GetState(TimeSpan timeout)
+    private async Task<Session> GetState(TimeSpan timeout)
```
StefanOssendorf, line 42: This method is only used by the initialization methods which do not return the session.
So instead of adding it here to the method signature the right thing to do would be to just discard the `RetrieveSession` result like: 
`_ = await _sessionManager.RetrieveSession(timeout);`
That would simplify the method a bit.

Source/Csla.Blazor.WebAssembly.Tests/SessionManagerTests.cs
```diff
@@ -0,0 +1,205 @@
...
+    public async Task SendSession_WithCancelledCancellationToken_ShouldThrowOperationCanceledException()
+    {
+      var cts = new CancellationTokenSource();
+      cts.Cancel();
+      await Assert.ThrowsExceptionAsync<TaskCanceledException>(() => _sessionManager.SendSession(cts.Token));
+    }
+
+
+    [TestMethod]
+    public void  RetrieveCAchedSessionSession()
```
StefanOssendorf, line 199: This tests looks like it's unfinished?

Source/Csla.Blazor.WebAssembly.Tests/Csla.Blazor.WebAssembly.Tests.csproj
```diff
@@ -0,0 +1,36 @@
...
+    <ApplicationIcon />
+    <OutputType>Library</OutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="8.0.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
+    <PackageReference Include="Moq" Version="4.20.70" />
```
StefanOssendorf, line 15: @rockfordlhotka Please decide whether we want to be depent on that or not.

Source/Csla.Blazor.WebAssembly.Tests/SessionManagerTests.cs
```diff
@@ -0,0 +1,205 @@
...
+
+    [TestMethod]
+    public async Task SendSession_WithValidCancellationToken_ShouldNotThrowException()
+    {
+      var cts = new CancellationTokenSource();
+      await _sessionManager.SendSession(cts.Token);
+    }
+
+    [TestMethod]
+    public async Task SendSession_WithCancelledCancellationToken_ShouldThrowOperationCanceledException()
```
StefanOssendorf, line 190: Method name does not match test. If we were to change the exception thrown by `SendSession` from `TaskCanceledException` to `OperationCanceledException` this test would fail. `Assert.ThrowsExceptionsAsync` tests for the exact type of exception and you are testing for `TaskCanceledException` isntead of `OperationCanceledException` as the method name suggest.
Maybe you can take advantage of using [FluentAssertions](https://fluentassertions.com/exceptions/) here for easier checking?

Source/Csla.Blazor.WebAssembly.Tests/SessionManagerTests.cs
```diff
@@ -0,0 +1,205 @@
...
+      var timeout = TimeSpan.FromHours(1);
+      await _sessionManager.SendSession(timeout);
+    }
+
+    [TestMethod]
+    public async Task SendSession_WithZeroTimeout_ShouldNotThrowException()
+    {
+      var timeout = TimeSpan.Zero;
+      await Assert.ThrowsExceptionAsync<TaskCanceledException>(() => _sessionManager.SendSession(timeout));
+
```
StefanOssendorf, line 179: Empty line

Source/Csla.Blazor.WebAssembly.Tests/SessionManagerTests.cs
```diff
@@ -0,0 +1,205 @@
...
+    [TestMethod]
+    public async Task SendSession_WithTimeoutValue_ShouldNotThrowException()
+    {
+      var timeout = TimeSpan.FromHours(1);
+      await _sessionManager.SendSession(timeout);
+    }
+
+    [TestMethod]
+    public async Task SendSession_WithZeroTimeout_ShouldNotThrowException()
```
StefanOssendorf, line 175: Method name does not reflect test. You are testing here for an exception.

Source/Csla.Blazor.WebAssembly.Tests/SessionManagerTests.cs
```diff
@@ -0,0 +1,205 @@
...
+
+    [TestMethod]
+    public async Task RetrieveSession_WithValidCancellationToken_ShouldNotThrowException()
+    {
+      var cts = new CancellationTokenSource();
+      await _sessionManager.RetrieveSession(cts.Token);
+    }
+
+    [TestMethod]
+    public async Task RetrieveSession_WithCancelledCancellationToken_ShouldThrowOperationCanceledException()
```
StefanOssendorf, line 160: Same comment as on `SendSession_WithCancelledCancellationToken_ShouldThrowOperationCanceledException`.

Source/Csla.Blazor.WebAssembly.Tests/SessionManagerTests.cs
```diff
@@ -0,0 +1,205 @@
...
+      var timeout = TimeSpan.FromHours(1);
+      await _sessionManager.RetrieveSession(timeout);
+    }
+
+    [TestMethod]
+    public async Task RetrieveSession_WithZeroTimeout_ShouldNotThrowException()
+    {
+      var timeout = TimeSpan.Zero;
+      await Assert.ThrowsExceptionAsync<TaskCanceledException>(() => _sessionManager.RetrieveSession(timeout));
+
```
StefanOssendorf, line 149: Empty line

Source/Csla.Blazor.WebAssembly.Tests/SessionManagerTests.cs
```diff
@@ -0,0 +1,205 @@
...
+    [TestMethod]
+    public async Task RetrieveSession_WithTimeoutValue_ShouldNotThrowException()
+    {
+      var timeout = TimeSpan.FromHours(1);
+      await _sessionManager.RetrieveSession(timeout);
+    }
+
+    [TestMethod]
+    public async Task RetrieveSession_WithZeroTimeout_ShouldNotThrowException()
```
StefanOssendorf, line 145: Method name does not reflect test. You are testing here for an exception.

StefanOssendorf:
Thanks for your changes. I'll review them shortly.
Please make sure to have signed contributor agreement sent do rocky. Currently I don't see you in the respective team. But rocky is currently on the road so it might just take a bit of time for him to update it.

Source/Csla.Blazor/State/StateManager.cs
```diff
@@ -57,12 +56,12 @@ private async Task GetState(TimeSpan timeout)
     /// at which you know the user is navigating to another
     /// page.
     /// </remarks>
-    public void SaveState()
+    public void SaveState(TimeSpan timeout)
```
StefanOssendorf, line 59: @rockfordlhotka Is there any reason why this isn't `Task` too? Did you want to avoid `IAsyncDisposable` or just fire&forget for the user? Writing this it's awkward to have a timeout but not the choice to wait for the timeout or not 🤔

StefanOssendorf (commented):
Two minor things and a question for rocky.

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -53,25 +65,34 @@ public async Task<Session> RetrieveSession()
...
           {
-            provider.SetPrincipal(message.Principal);
+            _session = GetSession();
           }
         }
-        else // NoUpdates
+        catch(OperationCanceledException ocex)
         {
-          _session = GetSession();
+          throw new TimeoutException ($"{this.GetType().FullName}.RetrieveSession.", ocex);
```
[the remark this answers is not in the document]
luizfbicalho: Sure, will do that, 

In fact I gave it a thought and shoudn't we try/catch this at the method that receives the timeout, this way if we call the method with the cancellation token we will get the TaskcancelledException and if we call with the timeout we will get the timeout exception?

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -95,7 +132,14 @@ public async Task SendSession()
...
         formatter.Serialize(buffer, _session);
         buffer.Position = 0;
-        await client.PutAsJsonAsync<byte[]>(_options.StateControllerName, buffer.ToArray());
+        try
+        {
+          await client.PutAsJsonAsync<byte[]>(_options.StateControllerName, buffer.ToArray(), ct);
+        }
+        catch(OperationCanceledException ocex)
+        {
+          throw new TimeoutException($"{this.GetType().FullName}.SendSession.", ocex);
```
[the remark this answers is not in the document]
luizfbicalho: Sure, will do that, 

In fact I gave it a thought and shoudn't we try/catch this at the method that receives the timeout, this way if we call the method with the cancellation token we will get the TaskcancelledException and if we call with the timeout we will get the timeout exception?

Source/Csla.Blazor/State/StateManager.cs
luizfbicalho: I'm open to get another tasks and update the async, maybe start adding cancellation token in other interfaces

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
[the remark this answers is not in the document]
StefanOssendorf: That sounds like a good idea. Please do :)

Source/Csla.Blazor/State/StateManager.cs
StefanOssendorf: I don't know about adding cancellation token in other interfaces.
I don't think it's easy to add it easily in the whole framework. We should definitly consult @rockfordlhotka before anyone tackles that.

luizfbicalho: Do you want to change this to a Task now?

rockfordlhotka: It should return `Task`. 

I am fine with an overload of the method that optionally accepts a cancelation token - assuming it is possible to cancel the underlying operation.

On the server, the operation typically isn't even async, but _could_ be if someone used REDIS or something. On the wasm client it uses http, and so we may be able to pass the token to the `HttpClient` call?

rockfordlhotka (changes requested):
Please do return `Task` and (if possible) use a cancelation token via an optional overload of the method.

PR #3903, opened 2024-05-02
#3897 Fix error catching in blazor ApplicationContextManagers

### LINKED ISSUE #3890: Do not call GetAuthenticationStateAsync outside of the DI scope for a Razor component
A change in CSLA from 8.0.0 to 8.0.1 appears to have caused an issue with authentication: "Do not call GetAuthenticationStateAsync outside of the DI scope for a Razor component"

Described in #3815

### LINKED ISSUE #3897: Blazor:  ApplicationContextManagerBlazor/InMemory does not await a task which could have an exception
As discuessed in https://github.com/MarimerLLC/csla/discussions/3815#discussioncomment-9286925 there is a potential bug in `ApplicationContextManagerBlazor` and `ApplicationContextManagerInMemory`.

See the following code:
https://github.com/MarimerLLC/csla/blob/279940686a95fc0ca1892d5a4b9300326e719156/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs#L71-L75
https://github.com/MarimerLLC/csla/blob/279940686a95fc0ca1892d5a4b9300326e719156/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs#L61-L65

In both cases the task is retrieved but not awaited which means an implementation of `GetAuthenticationStateAsync()` with real async code does not raise any exception here.
The `ServerAuthenticationStateProvider` throws an exception without bein async which means the exception is immediately thrown and not when the task is awaited. (c.f. https://github.com/MarimerLLC/csla/discussions/3815#discussioncomment-9287012)

swegele:
make async and match new error text after Microsoft changed it old exception text.  See [discussion.](https://github.com/MarimerLLC/csla/discussions/3815)

Not sure about the async part if I've got that right.  But I tested it and things work fine.  
After upgrading to CSLA 8.1.0, before this change, I would hit the exception immediately on blazor project start (probably because of my HostedService).  After this fix I am able to run the blazor project without issue.  All my thousands of tests pass too.

Fixes #3897
Fixes #3890

EDIT:  I also used this to get your BlazorExample working to on the list page (according to @russblair repro steps [here](https://github.com/MarimerLLC/csla/discussions/3815#discussioncomment-9239628))

### DIFF
--- a/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
+++ b/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
@@ -70,13 +70,23 @@ namespace Csla.AspNetCore.Blazor
         Task<AuthenticationState> task;
         try
         {
-          await(task = AuthenticationStateProvider.GetAuthenticationStateAsync());
+          task = AuthenticationStateProvider.GetAuthenticationStateAsync();
+          await task;
         }
         catch (InvalidOperationException ex)
         {
           task = Task.FromResult(new AuthenticationState(UnauthenticatedPrincipal));
           string message = ex.Message;
-          if (message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync) + " outside of the DI scope for a Razor component"))
+
+          //check for pre .net 8 error text
+          bool oldGetCalledBeforeSetError = message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync))
+            && message.Contains(nameof(IHostEnvironmentAuthenticationStateProvider.SetAuthenticationState));
+
+          //check for .net 8+ error text
+          bool newGetCalledOutsideRazorScopeError = message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync)
+            + " outside of the DI scope for a Razor component", StringComparison.OrdinalIgnoreCase);
+
+          if (oldGetCalledBeforeSetError || newGetCalledOutsideRazorScopeError)
           {
             SetHostPrincipal(task);
           }
--- a/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
+++ b/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
@@ -76,17 +76,11 @@ namespace Csla.AspNetCore.Blazor
         catch (InvalidOperationException ex)
         {
           task = Task.FromResult(new AuthenticationState(UnauthenticatedPrincipal));
+
           string message = ex.Message;
-
-          //check for pre .net 8 error text
-          bool oldGetCalledBeforeSetError = message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync))
-            && message.Contains(nameof(IHostEnvironmentAuthenticationStateProvider.SetAuthenticationState));
-
-          //check for .net 8+ error text
-          bool newGetCalledOutsideRazorScopeError = message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync)
-            + " outside of the DI scope for a Razor component", StringComparison.OrdinalIgnoreCase);
-
-          if (oldGetCalledBeforeSetError || newGetCalledOutsideRazorScopeError)
+          //see ms error https://github.com/dotnet/aspnetcore/blob/87e324a61dcd15db4086b8a8ca7bd74ca1e0a513/src/Components/Server/src/Circuits/ServerAuthenticationStateProvider.cs#L16
+          //not much safe to test on except the error type and the use of this method name in message.
+          if (message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync)))
           {
             SetHostPrincipal(task);
           }
--- a/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
+++ b/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
@@ -36,7 +36,7 @@ namespace Csla.AspNetCore.Blazor
       ActiveCircuitState = activeCircuitState;
       CurrentPrincipal = UnauthenticatedPrincipal;
       AuthenticationStateProvider.AuthenticationStateChanged += AuthenticationStateProvider_AuthenticationStateChanged;
-      InitializeUser();
+      _ = InitializeUser();
     }
 
     private IPrincipal CurrentPrincipal { get; set; }
--- a/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
+++ b/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
@@ -63,21 +63,20 @@ namespace Csla.AspNetCore.Blazor
     /// </summary>
     protected HttpContext HttpContext { get; set; }
 
-    private void InitializeUser()
+    private async Task InitializeUser()
     {
       if (ActiveCircuitState.CircuitExists)
       {
         Task<AuthenticationState> task;
         try
         {
-          task = AuthenticationStateProvider.GetAuthenticationStateAsync();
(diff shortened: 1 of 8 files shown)


Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -63,21 +63,20 @@ public ApplicationContextManagerBlazor(IHttpContextAccessor hca, AuthenticationS
...
-          task = AuthenticationStateProvider.GetAuthenticationStateAsync();
+          await(task = AuthenticationStateProvider.GetAuthenticationStateAsync());
         }
         catch (InvalidOperationException ex)
         {
           task = Task.FromResult(new AuthenticationState(UnauthenticatedPrincipal));
           string message = ex.Message;
-          if (message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync))
-              && message.Contains(nameof(IHostEnvironmentAuthenticationStateProvider.SetAuthenticationState)))
+          if (message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync) + " outside of the DI scope for a Razor component"))
```
StefanOssendorf, line 79: When did the message get changed? We should `#if` or check for both versions of the exception to keep compatibility with older .Net versions.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -63,21 +63,20 @@ public ApplicationContextManagerBlazor(IHttpContextAccessor hca, AuthenticationS
...
-    private void InitializeUser()
+    private async Task InitializeUser()
     {
       if (ActiveCircuitState.CircuitExists)
       {
         Task<AuthenticationState> task;
         try
         {
-          task = AuthenticationStateProvider.GetAuthenticationStateAsync();
+          await(task = AuthenticationStateProvider.GetAuthenticationStateAsync());
```
StefanOssendorf, line 73: This looks kinda odd for me. I would prefer to split it into two lines.
First assign the variable and then await the task.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -63,21 +63,20 @@ public ApplicationContextManagerBlazor(IHttpContextAccessor hca, AuthenticationS
     /// </summary>
     protected HttpContext HttpContext { get; set; }
 
-    private void InitializeUser()
+    private async Task InitializeUser()
```
StefanOssendorf, line 66: Because of the change the ctor now produces 3 warnings. Please get rid of them with the discard syntax: `_ = InitializeUser()`.
I hate stuff like this in the ctor but I think we can't avoid it for now.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
```diff
@@ -55,19 +56,18 @@ public ApplicationContextManagerInMemory(AuthenticationStateProvider authenticat
       InitializeUser();
     }
 
-    private void InitializeUser()
+    private async Task InitializeUser()
```
StefanOssendorf, line 59: Please apply the same comments here as in the other Manager.

StefanOssendorf (changes requested):
Thanks for the changes :)

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
swegele: Good point.  Fixed for backward compatibility.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
swegele: Done...and done

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
swegele: Done...and done

StefanOssendorf:
Thanks for the changes. Do you happen to know if it's possible to add tests for this behavior?

swegele:
Good question and that would be nice just in case some "helpful" person at MS decides to change the message again :-)   Things are a bit fuzzy going back to 2022 but I "think" it was my HostedService in my Blazor project that was causing this to throw such that I noticed it.  Obviously the BlazorExample throwing on the List page shows there are more ways to trigger it though.  Anyway, my HostedService starts at project startup and runs on a timer outside of both HttpContext and a SignalR Circuit fetching a read-only list of site translation strings.  After upgrading to CSLA 8.0.1 I get that error at startup (can't even get to my login/start page).

So it may be possible to mimic that somehow in a test?  My gut says yes it should be do-able.  It would take some fooling around on a weekend to see.  Should we put in an issue for it so we don't forget?

swegele:
@StefanOssendorf - git seems to think there is still an unresolved change request from you and I can't figure it out.  Did I miss something?

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -63,21 +63,30 @@ public ApplicationContextManagerBlazor(IHttpContextAccessor hca, AuthenticationS
...
-          if (message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync))
-              && message.Contains(nameof(IHostEnvironmentAuthenticationStateProvider.SetAuthenticationState)))
+
+          //check for pre .net 8 error text
+          bool oldGetCalledBeforeSetError = message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync))
+            && message.Contains(nameof(IHostEnvironmentAuthenticationStateProvider.SetAuthenticationState));
+
+          //check for .net 8+ error text
+          bool newGetCalledOutsideRazorScopeError = message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync)
+            + " outside of the DI scope for a Razor component", StringComparison.OrdinalIgnoreCase);
```
rockfordlhotka, line 87: Having hard-coded english text seems problematic - I must think that this text will vary depending on the culture of the user and/or server?

rockfordlhotka (changes requested):

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
swegele: I wondered about that.  Any ideas?  The exception doesn't have anything else I can grab onto that I can see.

swegele: Here is the exception in aspnet.core

https://github.com/dotnet/aspnetcore/blob/87e324a61dcd15db4086b8a8ca7bd74ca1e0a513/src/Components/Server/src/Circuits/ServerAuthenticationStateProvider.cs#L16

swegele: Stepping back for a moment...I guess the risk of simply testing for `message.Contains(nameof(AuthenticationStateProvider.GetAuthenticationStateAsync)` is low because this method is only hit on CTOR so it's not like it would be firing repeatedly over and over.  What do you think about me removing everything but that on the check for the `InvalidOperationException`?

rockfordlhotka: wow, it is amazing that they accepted hard-coded text into the repo like that.

Why do we need to be aware of the specific text? Is there some way we can shield against future changes? This seems fragile to me.

swegele:
@rockfordlhotka  does that seem a little better?  That would have gotten us through the recent language change.  Combine that with the fact that the type of exception is checked for as well.

rockfordlhotka (approved):

PR #3909, opened 2024-05-03
Resolve build issues in csla.test.sln

rockfordlhotka:
* Remove generator project, as we're not releasing it at the moment
* Remove dataset generation, as it is no longer used
* Remove System.Transactions reference as it is no longer used


StefanOssendorf:
Why are the database tests removed? Do you plan on dropping system.transactions?

rockfordlhotka:
> Why are the database tests removed? Do you plan on dropping system.transactions?

I think we might need to create a new csla.localtest.sln or something, where we can run stuff that can't run on the CI server.

Or maybe we add the tests back in when we have a database that can work on the CI server.

In any case, I'm not sure we want or need the DataSet tests anymore at all.

PR #3920, opened 2024-05-07
Updates to Blazor state management and BlazorExample

### LINKED ISSUE #3919: Blazor server state manager doesn't update values
When running in a server-interactive page, the Blazor session state doesn't get updated correctly.

What appears to happen is that the session (e.g. LocalContext) value gets updated in memory, but the last touched property on the session isn't updated, so the changed value doesn't propogate correctly to other callers.

rockfordlhotka:
Closes #3919

### DIFF
--- a/Samples/BlazorExample/BlazorExample/BlazorExample.Client/Components/Pages/Counter.razor
+++ b/Samples/BlazorExample/BlazorExample/BlazorExample.Client/Components/Pages/Counter.razor
@@ -1,7 +1,7 @@
 ﻿@page "/counter"
 @rendermode @(new InteractiveAutoRenderMode(prerender: false))
-@implements IDisposable /* CSLA wasm pages must be IDisposable */
 @inject Csla.Blazor.State.StateManager StateManager
+@inject ApplicationContext applicationContext
 
 <PageTitle>Counter</PageTitle>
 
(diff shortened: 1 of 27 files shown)


rockfordlhotka:
@StefanOssendorf I'm using this code today for my talk in Antwerp. This PR should get merged before we build the 8.1.1 version, because it fixes a bug in the Blazor state management code.

Samples/BlazorExample/BlazorExample/BlazorExample.Client/Components/Pages/Counter.razor
```diff
@@ -1,35 +1,55 @@
...
+
+        if (IsInteractive)
+        {
+            if (!applicationContext.LocalContext.Contains("counter"))
+            {
+                applicationContext.LocalContext["counter"] = 0;
+                StateManager.SaveState();
+            }
+            var tmp = applicationContext.LocalContext["counter"];
+            currentCount = tmp != null ? (int)tmp : 0;
```
StefanOssendorf, line 44: Do you expect concurrency issues here? Otherwise `LocalContext["counter"]` should never return `null`. You are making sure it's at least initialized with `0` 3 lines before.

PR #3925, opened 2024-05-11
Coerce output in RuleContext.AddOutValue

### LINKED ISSUE #3251: Support parameter conversion in RuleContext.AddOutputValue
Currently, AddOutputValue in RuleContext does not support parameter conversion like SetPropertyConvert<F,P>.

`
    public void AddOutValue(Csla.Core.IPropertyInfo property, object value)
    {
      _outputPropertyValues.Value[property] = value;
    }
`

May be adding support for parameter conversion make it easier to use, especially  when the managed field is smart types (SmartDate)?

`
 public void AddOutValue(Csla.Core.IPropertyInfo property, object value)
            {
                if(value is not null)                
                    _outputPropertyValues.Value[property] = Csla.Utilities.CoerceValue(PrimaryProperty.Type, value.GetType(), null, value);                
                else
                    _outputPropertyValues.Value[property] = PrimaryProperty.DefaultValue;
            }
`

rrrooommmaaa:
Close #3251

### DIFF
--- a/Source/Csla.test/ValidationRules/OutputCoercionTest.cs
+++ b/Source/Csla.test/ValidationRules/OutputCoercionTest.cs
@@ -1,0 +1,69 @@
+﻿using Csla.Rules;
+using Csla.TestHelpers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Csla.Test.ValidationRules
+{
+  public class CalculationObject : BusinessBase<CalculationObject>
+  {
+    public static readonly PropertyInfo<int> PropertyAProperty = RegisterProperty<int>(nameof(PropertyA));
+    public int PropertyA
+    {
+      get { return GetProperty(PropertyAProperty); }
+      set { SetProperty(PropertyAProperty, value); }
+    }
+
+    public static readonly PropertyInfo<int> PropertyBProperty = RegisterProperty<int>(nameof(PropertyB));
+    public int PropertyB
+    {
+      get { return GetProperty(PropertyBProperty); }
+      set { SetProperty(PropertyBProperty, value); }
+    }
+
+    public static readonly PropertyInfo<string> SumPropertyProperty = RegisterProperty<string>(nameof(SumProperty));
+    public string SumProperty
+    {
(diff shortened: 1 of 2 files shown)


StefanOssendorf:
Thank you for your contribution.
I'm not sure if the implementation is what we want. @rockfordlhotka shoud make a decision here.
My "problem" is: With the new implemention you can't set a `string` property to `null` because the CSLA default is `string.Empty`.
This is a breaking change in current behavior and I don't know how many people rely on that.
I'm personally okay with this change.

StefanOssendorf (changes requested):
As marker to wait for response from @rockfordlhotka

rrrooommmaaa:
@StefanOssendorf I've got the same conversion from null to String.Empty when trying to write null value directly to a string property because of
https://github.com/MarimerLLC/csla/blob/a60f4da5d42538dcb24f5226a99f0f58c9bcbd61/Source/Csla/Core/BusinessBase.cs#L2451-L2452

StefanOssendorf:
> @StefanOssendorf I've got the same conversion from null to String.Empty when trying to write null value directly to a string property because of
> 
> https://github.com/MarimerLLC/csla/blob/a60f4da5d42538dcb24f5226a99f0f58c9bcbd61/Source/Csla/Core/BusinessBase.cs#L2451-L2452

That may be the case. But currently a rule can set a string to `null` with an out value. I just tested that with 8.1.0.

rockfordlhotka:
The reason that, in general, CSLA tries to ensure that a string property is empty instead of null, is that (iirc) Windows Forms data binding can't handle null in some cases.

If someone is using Windows Forms and they created a rule that set a string to null, that would probably cause data binding to fail in their app. I don't think anyone else would notice, because I don't think any other UI frameworks have this limitation.

All that said, this is a breaking change - but one that has no real meaning, since CSLA string properties generally convert null into empty already.

Someday maybe we'll change everything to support null string property values, but that day is not today.

I think this change is fine.

StefanOssendorf:
Alrighty. LGTM then :-)

StefanOssendorf (approved):

PR #3926, opened 2024-05-11
Feature/add cancellation token 

### LINKED ISSUE #3911: Add cancelation token overloads to appropriate async methods
There are a number of async methods where having an overload with a cancelation token may make sense.

This is only appropriate if the cancelation token can flow down to a system API that can actually cancel the operation. 

It is generally _not_ appropriate for the data portal, because it is not possible to flow the cancelation token across network boundaries. For example, indicating that someone could "cancel" a client-side data portal operation makes little sense, because _maybe_ we could cancel the http or gprc or rabbitmq request on the client, but there's no way to actually cancel the operation on the server.

luizfbicalho:
Add cancellation token as an alternative to timeout

closes #3911


Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
```diff
@@ -41,132 +41,134 @@ public Session GetCachedSession()
...
+    /// Retrieves the current user's session from
+    /// the web server to the wasm client
+    /// if SyncContextWithServer is true.
+    /// </summary>
+    /// <param name="timeout">The timeout duration for the operation.</param>
+    /// <returns>The retrieved session.</returns>
+    /// <exception cref="TimeoutException">Thrown when the specified timeout is exceeded.</exception>
+    public async Task<Session> RetrieveSession(TimeSpan timeout)
+    {
+        try
```
StefanOssendorf, line 54: Wrong formatting. 3 spaces instead of 2 within the methods.
Looks like this is true for the whole file.

Source/Csla.Blazor/State/StateManager.cs
```diff
@@ -10,59 +10,100 @@
...
-  {
-    private readonly ISessionManager _sessionManager = sessionManager;
-
     /// <summary>
-    /// Get state from cache.
+    /// Get and save state from Blazor pages.
     /// </summary>
-    public Task InitializeAsync()
+    /// <param name="sessionManager">The session manager to use for state management.</param>
+    public class StateManager(ISessionManager sessionManager)
```
StefanOssendorf, line 17: Wrong formatting. 3 spaces instead of 2 within the methods.
Looks like this is true for the whole file.

Source/Csla.Blazor/State/StateManager.cs
```diff
@@ -10,59 +10,100 @@
...
+            {
+                await _sessionManager.SendSession(timeout);
+            }
+        }
+
+        /// <summary>
+        /// Get state from cache.
+        /// </summary>
+        /// <param name="ct">Cancellation token</param>
+        public Task InitializeAsync(CancellationToken ct)
```
StefanOssendorf, line 73: Please reorder the code so that the method (overloads) are beneath each other from least to most parameters.

Source/Csla.Blazor/State/StateManager.cs
```diff
@@ -10,59 +10,100 @@
...
+        /// at which you know the user is navigating to another
+        /// page.
+        /// </remarks>
+        /// <param name="ct">Cancellation token</param>
+        public async Task SaveState(CancellationToken ct)
+        {
+            var isBrowser = OperatingSystem.IsBrowser();
+            if (isBrowser)
+            {
+                await _sessionManager.SendSession(ct);
```
StefanOssendorf, line 105: Even though it's C&P please remove the curly braces for the one liner if

StefanOssendorf (changes requested):
I only checked the first four files (SessionManager, StateManager and ViewModel).
It seems all touched files have formatting issues. Because of that I abort the review until you recheck all files. The correct intendation is set by the editorconfig so this should be fixable by auto formatting.

Source/Csla.Blazor/State/StateManager.cs
luizfbicalho: In my personal opinion I don't like to remove the curly braces on one lines, i think that in the future we can get any method and add a line.

Anyway, I removed in this case.

Source/Csla.Blazor/State/StateManager.cs
luizfbicalho: now I runned the dotnet format csla.test.sln and I got more than 300 files changed, I think that after this merge we should open a new PR to format them all, what do you think @StefanOssendorf ?

Source/Csla.Blazor.WebAssembly/State/SessionManager.cs
luizfbicalho: we could add this on all of the .csproj files  @StefanOssendorf 


```xml
<PropertyGroup>
  <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
</PropertyGroup>
```

luizfbicalho:
> I only checked the first four files (SessionManager, StateManager and ViewModel). It seems all touched files have formatting issues. Because of that I abort the review until you recheck all files. The correct intendation is set by the editorconfig so this should be fixable by auto formatting.

I used the dotnet format csla.test.sln to resolve this issue, but there are too many changes across the solution, more than 300 files need change.

I just updated those files changed in this PR, we can discuss later something to improve this.

Source/Csla/Rules/BusinessRules.cs
```diff
@@ -477,10 +478,23 @@ public Task<List<string>> CheckRulesAsync(IPropertyInfo property)
...
       return affectedProperties;
     }
 
+    /// <summary>
+    /// Asynchronously checks the rules for the specified property and waits for any asynchronous rules to complete.
+    /// </summary>
+    /// <param name="property">The property to check the rules for.</param>
+    /// <param name="ct">The cancellation token.</param>
+    /// <returns>A list of affected properties.</returns>
+    public async Task<List<string>> CheckRulesAsync(IPropertyInfo property, CancellationToken ct)
```
StefanOssendorf, line 491: #

rockfordlhotka:
Regarding the change of timeout in async rule execution to use a CancelationToken - does this provide value?

Right now it appears to be a different way to achieve the same result, and so the change brings on (probably minor) risk for no gain. On the other hand, if we think it is paving the way for a future where canceling async rule execution could be done via code, then perhaps there's value?

luizfbicalho:
> Regarding the change of timeout in async rule execution to use a CancelationToken - does this provide value?
> 
> Right now it appears to be a different way to achieve the same result, and so the change brings on (probably minor) risk for no gain. On the other hand, if we think it is paving the way for a future where canceling async rule execution could be done via code, then perhaps there's value?

Timeout I think that is much more limited, I have an example one project that one of my tasks didn't pass correctly the cancellation token, and this was a problem in te IIS reset, this task couldn't be cancelled in time, and used to break the IIS reset, causing 503

rockfordlhotka:
> > Regarding the change of timeout in async rule execution to use a CancelationToken - does this provide value?
> > Right now it appears to be a different way to achieve the same result, and so the change brings on (probably minor) risk for no gain. On the other hand, if we think it is paving the way for a future where canceling async rule execution could be done via code, then perhaps there's value?
> 
> Timeout I think that is much more limited, I have an example one project that one of my tasks didn't pass correctly the cancellation token, and this was a problem in te IIS reset, this task couldn't be cancelled in time, and used to break the IIS reset, causing 503

What I am getting at though, is that the code doesn't allow someone to cancel the operation because the cancelation token is created off the timeout and isn't exposed higher in the API.

I think we need to consider what someone would want. If they call `CheckRulesAsync` and cancel it, I imagine they'd want to cancel _all_ executing rules? Sync and async. Though we don't have a way to cancel sync rules, so it would be async rules only.

Inside an async rule, presumably we pass in the cancelation token (via `context`?) so a rule author could (if possible) cancel what they are doing.

Source/Csla/Core/BusyHelper.cs
```diff
@@ -8,45 +8,71 @@
...
-
-        var timeoutTask = Task.Delay(timeout);
-        var finishedTask = await Task.WhenAny(tcs.Task, timeoutTask).ConfigureAwait(false);
-
-        if (finishedTask == timeoutTask)
-        {
-          throw new TimeoutException($"{source.GetType().FullName}.{methodName} - {timeout}.");
-        }
+#if NET6_0_OR_GREATER
+                var finishedTask = await tcs.Task.WaitAsync(ct).ConfigureAwait(false);
```
StefanOssendorf, line 62: Formatting

StefanOssendorf (changes requested):

StefanOssendorf:
I agree with Rocky here. We should remove the business rules changes here and make a proper design/concept for how we want to be able to use a cancellation token in the context of rules.
As rocky said. I as a user would expect to be able to cancel the current running rules and not only to provide a waiting timeout.

StefanOssendorf (changes requested):
I forgot. Please add tests only for the new APIs at least. Thanks :)

luizfbicalho:
> I agree with Rocky here. We should remove the business rules changes here and make a proper design/concept for how we want to be able to use a cancellation token in the context of rules. As rocky said. I as a user would expect to be able to cancel the current running rules and not only to provide a waiting timeout.

OK, which one do you want to rollback?

StefanOssendorf:
> > I agree with Rocky here. We should remove the business rules changes here and make a proper design/concept for how we want to be able to use a cancellation token in the context of rules. As rocky said. I as a user would expect to be able to cancel the current running rules and not only to provide a waiting timeout.
> 
> OK, which one do you want to rollback?

If possible this whole file `Source/Csla/Rules/BusinessRules.cs`. But if that's not easy only the public method accepting the ct would be okay for me.

luizfbicalho:
> > > I agree with Rocky here. We should remove the business rules changes here and make a proper design/concept for how we want to be able to use a cancellation token in the context of rules. As rocky said. I as a user would expect to be able to cancel the current running rules and not only to provide a waiting timeout.
> > 
> > 
> > OK, which one do you want to rollback?
> 
> If possible this whole file `Source/Csla/Rules/BusinessRules.cs`. But if that's not easy only the public method accepting the ct would be okay for me.

I changed it back to the original, and already solved the format problems, the tests I'll do in the weekend

StefanOssendorf:
Resolve the confligt and we are ready to merge :)

StefanOssendorf (approved):

luizfbicalho:
> Resolve the confligt and we are ready to merge :)

Are you sure? I'm still making some more tests

StefanOssendorf:
> > Resolve the confligt and we are ready to merge :)
> 
> Are you sure? I'm still making some more tests

I won't say no to more useful tests :)

luizfbicalho:
> > > Resolve the confligt and we are ready to merge :)
> > 
> > 
> > Are you sure? I'm still making some more tests
> 
> I won't say no to more useful tests :)

I added some more. Please check if it's all ok

StefanOssendorf (changes requested):

Source/Csla.Blazor.Test/ViewModelSaveAsyncErrorTests.cs
```diff
@@ -72,6 +73,86 @@ public async Task SavingSuccess_ErrorEventIsNotInvoked()
...
+
+      public bool IsSavable { get; set; }
+
+      public bool IsChild { get; set; }
+
+      public bool IsBusy { get; set; }
+
+      public bool IsSelfBusy { get; set; }
+
+      public event BusyChangedEventHandler BusyChanged;
```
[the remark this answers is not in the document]
StefanOssendorf: That are default build warnings. Nothing special needed to get those. `dotnet build csla.test.sln` should produce those.

Source/Csla.test/Server/FakeEntity.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    public bool IsSavable => throw new NotImplementedException();
+
+    public bool IsChild => throw new NotImplementedException();
+
+    public bool IsBusy { get; set; }
+
+    public bool IsSelfBusy => throw new NotImplementedException();
+
+    public event BusyChangedEventHandler BusyChanged;
```
StefanOssendorf, line 52: Fix warnings here, too

StefanOssendorf (changes requested):

StefanOssendorf:
Please fix first the conflicts 😅

luizfbicalho:
> Please fix first the conflicts 😅

I'll fix tonight

StefanOssendorf (approved):
lgtm

Issue #3930, opened 2024-05-13
In 8.1.1 (prob in 9 too) - when ".UseInMemoryApplicationContextManager = true" you get "Unable to resolve service for type 'Csla.State.ISessionManager'"


swegele:
First noticed by @russblair mentioned on [discord channel](https://discord.com/channels/795719451148156938/795719451148156941)
Confirmed by @swegele

**Repro:**
Blazor Server-Interactive Only project (no wasm or any refs to new .NET 8 state management stuff)
Config in blazor server program.cs 

```csharp
builder.Services.AddCsla(o => o
  .AddAspNetCore()
  .AddServerSideBlazor(bsco => bsco
    .UseInMemoryApplicationContextManager = true));

...

        app.UseEndpoints(delegate (IEndpointRouteBuilder endpoint)
        {
            endpoint.MapControllers();
        });
```

Csla config correctly skips making any reference or registration to any state management stuff.  But none-the-less somehow .NET DI is looking for `ISessionManager`.  We think that ASPNet is scanning the project (and it's dependencies) for controllers and it finds the `[ApiController]` attribute here:

Csla.AspNetCore.Blazor.State.StateController
```csharp
  [ApiController]
  [Route("[controller]")]
  public class StateController(ApplicationContext applicationContext, ISessionManager sessionManager) : ControllerBase
```

**WORKAROUND**
I did a test workaround suggested by @TheCakeMonster making a no-op implementation for the `ISessionManager` interface and registered it.  With this the DI error goes away and project runs as expected.

```csharp
using Csla.State;

    public class NoOpSessionManager : ISessionManager
    {
        public Session GetCachedSession()
        { throw new NotImplementedException(); }
        public Session GetSession()
        { throw new NotImplementedException(); }
        public void PurgeSessions(TimeSpan expiration)
        { throw new NotImplementedException(); }
        public Task<Session> RetrieveSession()
        { throw new NotImplementedException(); }
        public Task SendSession()
        { throw new NotImplementedException(); }
        public void UpdateSession(Session newSession)
        { throw new NotImplementedException(); }
    }
...

builder.Services.AddSingleton<Csla.State.ISessionManager, NoOpSessionManager>();

```

PR #3931, opened 2024-05-13
#3930 - Fix for v8.x branch removing api controller attributes which …

swegele:
…stops ASPNet from including StateController (and it's CTOR services) in service registration;  implementing a no-op session manager when inmemory app context manager is used.

Reminder: Make sure to review CONTRIBUTING.MD and make sure you've sent in your signed contributor agreement.

Reminder: Add issue number for work item (e.g. #123)

### DIFF
--- a/Source/Csla.Blazor/State/NoOpSessionManager.cs
+++ b/Source/Csla.Blazor/State/NoOpSessionManager.cs
@@ -1,9 +1,9 @@
 ﻿//-----------------------------------------------------------------------
-// <copyright file="SessionServerManager.cs" company="Marimer LLC">
+// <copyright file="NoOpSessionManager.cs" company="Marimer LLC">
 //     Copyright (c) Marimer LLC. All rights reserved.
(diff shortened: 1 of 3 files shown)


Source/Csla.Blazor/State/NoOpSessionManager.cs
```diff
@@ -0,0 +1,63 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="SessionServerManager.cs" company="Marimer LLC">
```
rockfordlhotka, line 2: The filename here is incorrect

rockfordlhotka (changes requested):

rockfordlhotka (approved):

Source/Csla.Blazor/State/NoOpSessionManager.cs
swegele: whoops - good catch!

OK I will do main branch now

PR #3942, opened 2024-05-15
#2925 text separator for concatenating errors/main

### LINKED ISSUE #2925: TextSeparator = ", " for concatenating errors
The standard validation error messages such as `[Required]` are proper sentences ending with a `.` and I think this is good practice. `Csla.Blazor.PropertyInfo` assumes error messages won't end with a `.` using `", "` to concatenate errors, so you end up with:

image

Should `TextSeparator = " "` or perhaps be a `public static string TextSeparator { get; set; }` instead?

Freelancingonupwork:
By default TextSeparator will be " "  for concatenating errors. and someone can set this as well.

Fixes #2925 
TextSeparator


Freelancingonupwork:
> By default TextSeparator will be " " for concatenating errors. and someone can set this as well.

I see some issue in PR  I see commits which are already merged in main branch please don't review until I fixed it

rockfordlhotka:
@Freelancingonupwork I switched the PR to a draft. You can switch it back once you are ready.

Freelancingonupwork:
> @Freelancingonupwork I switched the PR to a draft. You can switch it back once you are ready.

HI @rockfordlhotka,  I fixed the issues by merging main into my feature branch now review the PR

Source/Csla/Csla.csproj
```diff
@@ -12,6 +12,7 @@
     <BaseOutputPath>..\..\Bin</BaseOutputPath>
     <GenerateDocumentationFile>True</GenerateDocumentationFile>
     <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
+    <Version>9.0.0.5</Version>
```
rockfordlhotka, line 15: There should be no `Version` element here - it comes from Directory.build.props

Source/Csla.Blazor/PropertyInfo.cs
```diff
@@ -16,8 +16,10 @@ namespace Csla.Blazor
   /// </summary>
   public class PropertyInfo : IPropertyInfo, INotifyPropertyChanged
   {
-    private const string TextSeparator = ", ";
-    
+    /// <summary>
+    /// Gets the value of the TextSeparator
+    /// </summary>
+    public string TextSeparator { get; }
```
rockfordlhotka, line 22: Why are we changing the default from ", " to " "? I would think we should keep the default the same?

Source/Csla.Blazor/PropertyInfo.cs
```diff
@@ -28,14 +30,16 @@ public class PropertyInfo : IPropertyInfo, INotifyPropertyChanged
     /// </summary>
     /// <param name="model">Model object</param>
     /// <param name="propertyName">Property name</param>
-    public PropertyInfo(object model, string propertyName)
+    /// <param name="textSeprator">Text Seprator</param>
+    public PropertyInfo(object model, string propertyName, string textSeprator = " ")
```
rockfordlhotka, line 34: Would this be better as a optional parameter so people who want to use the default can just ignore this new parameter?

Source/Csla.Blazor/ViewModel.cs
```diff
@@ -387,7 +387,24 @@ public IPropertyInfo GetPropertyInfo<P>(Expression<Func<P>> property)
 
       var keyName = property.GetKey();
       var identifier = Microsoft.AspNetCore.Components.Forms.FieldIdentifier.Create(property);
-      return GetPropertyInfo(keyName, identifier.Model, identifier.FieldName);
+      return GetPropertyInfo(keyName, identifier.Model, identifier.FieldName, " ");
```
rockfordlhotka, line 390: Shouldn't this use the default rather than hard-coding the separator? Same for other instances below.

rockfordlhotka (changes requested):

Source/Csla.Blazor/PropertyInfo.cs
Freelancingonupwork: it was there in description that Default will be " ".

image

Source/Csla.Blazor/ViewModel.cs
Freelancingonupwork: Yes, you are correct. i will do that.

Source/Csla/Csla.csproj
Freelancingonupwork: I will remove it

Source/Csla.Blazor/PropertyInfo.cs
rockfordlhotka: My bad 🫢

rockfordlhotka (approved):

PR #3953, opened 2024-05-17
BusyHelper for IsBusy Change

### LINKED ISSUE #3936: Use BusyHelper for IsBusy change
https://github.com/MarimerLLC/csla/blob/58e52bed3d8ec8d5078cf02d9f76e166412e5668/Source/Csla.Blazor/ViewModel.cs#L239-L248

This should be replaced by the `BusyHelper` class from Csla.Core to avoid the spin wait.

Freelancingonupwork:
[csla/Source/Csla.Blazor/ViewModel.cs](https://github.com/MarimerLLC/csla/blob/58e52bed3d8ec8d5078cf02d9f76e166412e5668/Source/Csla.Blazor/ViewModel.cs#L239-L248)
```
 if (obj.IsBusy) 
 { 
   var stopTime = DateTime.Now + BusyTimeout; 
   while (obj.IsBusy) 
   { 
     if (DateTime.Now > stopTime) 
       throw new TimeoutException("SaveAsync"); 
     await Task.Delay(1); 
   } 
 } 
```

Replaced by BusyHelper class from Csla.Core to avoid the spin wait.
Fixes #3936


Source/Csla.Blazor/ViewModel.cs
```diff
@@ -236,15 +237,11 @@ public async Task SaveAsync()
       {
         if (Model is Core.ITrackStatus obj && !obj.IsSavable)
         {
-          if (obj.IsBusy)
+          if (obj is INotifyBusy notifyBusy)
```
StefanOssendorf, line 240: This test is unnecessary since `ITrackStatus` guaranetees obj is of that type.

Source/Csla.Blazor/ViewModel.cs
```diff
@@ -236,15 +237,11 @@ public async Task SaveAsync()
...
           {
-            var stopTime = DateTime.Now + BusyTimeout;
-            while (obj.IsBusy)
-            {
-              if (DateTime.Now > stopTime)
-                throw new TimeoutException("SaveAsync");
-              await Task.Delay(1);
-            }
+            var cslaOptions = ApplicationContext.GetRequiredService<Csla.Configuration.CslaOptions>();
+            var _waitForIdleTimeout = TimeSpan.FromSeconds(cslaOptions.DefaultWaitForIdleTimeoutInSeconds);
```
StefanOssendorf, line 243: Variables do not start with _ (underscore)

Source/Csla/Core/BusyHelper.cs
```diff
@@ -12,7 +12,7 @@ namespace Csla.Core {
   /// <summary>
   /// Helper class for busy related functionality spread across different business type implementations.
   /// </summary>
-  internal static class BusyHelper 
+  public static class BusyHelper 
   {
     public static async Task WaitForIdle(INotifyBusy source, TimeSpan timeout, [CallerMemberName] string methodName = "")
```
StefanOssendorf, line 17: Please ad ann XML comment.

Source/Csla.Blazor/ViewModel.cs
```diff
@@ -236,15 +237,11 @@ public async Task SaveAsync()
...
+            await BusyHelper.WaitForIdle(notifyBusy, _waitForIdleTimeout, "SaveAsync").ConfigureAwait(false);
```
StefanOssendorf, line 244: "SaveAsync" as parameter is unnecessary. The method uses `CallerMemberName` to get that.

Source/Csla.Blazor.Test/ViewModelSaveAsyncErrorTests.cs
```diff
@@ -72,6 +72,30 @@ public async Task SavingSuccess_ErrorEventIsNotInvoked()
       Assert.IsNull(vm.ViewModelErrorText);
     }
 
+    [TestMethod]
+    public async Task CheckBusyHelper()
```
StefanOssendorf, line 76: How does this test test the new CheckyBusyHelper?
Line 95 does the `SaveAsync()` but at that point in time person is not busy at all. person is already "non busy" at line 88 because we have only sync rules which execute immediatly during property set (line 86).

To test the busy helper we need at least 2 tests. Busy will "finish" in time and the save is successful and Busy will exceed the timeout and we have to deal with a timeout exception.
And to test that busyness we need an async rule with a specifiec `Task.Dealy()` so we know how long our timeout has to be to either be longer than the rule or shorter so we get the exception.

StefanOssendorf (changes requested):

Source/Csla.Blazor/ViewModel.cs
Freelancingonupwork: ok. I will update that.

Source/Csla.Blazor/ViewModel.cs
Freelancingonupwork: noted.

Source/Csla/Core/BusyHelper.cs
Freelancingonupwork: ok

Source/Csla.Blazor/ViewModel.cs
Freelancingonupwork: ok. I will update that.

Source/Csla.Blazor.Test/ViewModelSaveAsyncErrorTests.cs
Freelancingonupwork: ok.

Source/Csla.Blazor/ViewModel.cs
```diff
@@ -236,16 +237,8 @@ public async Task SaveAsync()
...
-          {
-            var stopTime = DateTime.Now + BusyTimeout;
-            while (obj.IsBusy)
-            {
-              if (DateTime.Now > stopTime)
-                throw new TimeoutException("SaveAsync");
-              await Task.Delay(1);
-            }
-          }
+          await BusyHelper.WaitForIdle(obj, BusyTimeout).ConfigureAwait(false);
```
StefanOssendorf, line 240: Remove `ConfigureAwait(false)` here. We are in a UI context here and should keep the context.

Source/Csla.Blazor.Test/ViewModelSaveAsyncErrorTests.cs
```diff
@@ -72,6 +72,51 @@ public async Task SavingSuccess_ErrorEventIsNotInvoked()
...
+
+      var appCntxt = _testDIContext.CreateTestApplicationContext();
+      var vm = new ViewModel<Person>(appCntxt)
+      {
+        Model = person
+      };
+
+      await vm.SaveAsync();
+      Assert.IsNotNull(vm.Exception);
+      Assert.AreEqual(vm.Exception.Message, "Csla.Blazor.Test.Person.SaveAsync - 00:00:30.");
```
StefanOssendorf, line 117: This is way too long. We are in a unit test context here and a unit test should complete fast. _Really_ fast. So try to tune it down preferably under 1s and at most 2s.

StefanOssendorf (changes requested):

Source/Csla.Blazor/ViewModel.cs
Freelancingonupwork: Ok

Source/Csla.Blazor.Test/ViewModelSaveAsyncErrorTests.cs
Freelancingonupwork: ok. noted.

StefanOssendorf (approved):

StefanOssendorf:
@Freelancingonupwork Please fix the broken test then we can merge it :)

Freelancingonupwork:
> @Freelancingonupwork Please fix the broken test then we can merge it :)

Freelancingonupwork:
Hello,

By mistake I closed the PR. and when I reopened it. builds completed without error.

rockfordlhotka:
I have noticed that there is at least one test that sporadically fails. Frustrating!

Issue #2801, opened 2022-02-25
Add support for asynchronous authorization rules


rockfordlhotka:
### Discussed in https://github.com/MarimerLLC/csla/discussions/2643

<div type='discussions-op-text'>

<sup>Originally posted by **mtavares628** December  9, 2021</sup>
My question is two-fold:

1. Is there a way to make an AuthorizationRule Async like BusinessRuleAsync?
2. Is it possible to Adjust CslaPolicy so that is would support the other overloads of BusinessRules.HasPermission as well?

Here's the context for these questions. I have a project that I've been maintaining since the v3.x days where we created CanGetObject, CanAddObject, CanEditObject, etc. When 4.x was first introduced, there wasn't really an option to pass a dynamic parameter into the rules, and I needed the ability to hit the database to check the ID of the object to see if the user has individual permissions to view that object.  Basically, a user needed to have a certain role, but also have specific permission to view a client record.  Because 4.x didn't have that (or I couldn't see how to do it) I just continued to use the CanGetObject function that I was using for 3.x. So here's an example where I've used the 4.x AddObjectAuthorization with the Roles, and then call that in a CanGetObject function that I would then call specifically from my UI:

```
        protected static void AddObjectAuthorizationRules()
        {
            BusinessRules.AddRule(typeof(Client), new IsInRole(AuthorizationActions.CreateObject, "CreateClients"));
            BusinessRules.AddRule(typeof(Client), new IsInRole(AuthorizationActions.GetObject, "ViewClientRecords"));
            BusinessRules.AddRule(typeof(Client), new IsInRole(AuthorizationActions.EditObject, "EditClientRecords"));
            BusinessRules.AddRule(typeof(Client), new IsInRole(AuthorizationActions.DeleteObject, "DeleteClients"));
        }

        /// <summary>
        /// Determines whether the current user has rights to retrieve the specific client object
        /// </summary>
        public static bool CanGetObject(int clientID)
        {
            var result = default(bool);
            if (BusinessRules.HasPermission(AuthorizationActions.GetObject, typeof(Client)))
            {
                HasPermissionForClientCommand cmd = DataPortal.Create<HasPermissionForClientCommand>(clientID, AccessPermissions.ReadAccess);
                cmd = DataPortal.Execute(cmd);
                result = cmd.HasPermission;
            }

            return result;
        }
```
[...]

PR #1226, opened 2019-08-06
1196 Pass data portal criteria to auth rule via context so the rule can have more data to make a decision

ajj7060:
#1196 Added ability to get criteria used in authorization rules

PR #3956, opened 2024-05-19
Feature/asynchronous authorization rules

luizfbicalho:
I just added the basic idea so we could talk about the whole spec. 
Didn't add the tests because of that

#2801


luizfbicalho:
> @luizbicalhoagl, I don't have any immediate concerns about the changes, as long as everything still works 😄
> 
> And by "works" what I mean is that pages that use the API synchronously today keep working, as do any consumers using the API in an async manner.

Thanks for the feedback 

I did my best to create the async methods without impacting on the previous ones

If it's ok for you I can create the tests for some async authorization rules.
Do you have any scenario in mind that needs attention?

@StefanOssendorf do you have something to add?

rockfordlhotka:
The original request came from @mtavares628, so perhaps he has a good example for testing?

I think the real scenario is an authn rule that needs to do some async data portal command - like read from a database or other backend service. You can simulate this by creating a rule that does a `FetchAsync` on the data portal - perhaps on the logical server-side `Fetch` operation do an `await Task.Delay(10)` or something.

mtavares628:
@luizfbicalho, @rockfordlhotka hit the nail on the head with my use case.  I want to create an authorization rule that is able to perform a FetchAsync or ExecuteAsync to retrieve a value back from the database to determine whether the user is authorized or not. The criteria for that fetch would be passed in to the HasPermissionAsync method as the criteria object.  You could probably extend the tests from [PR #1226](https://github.com/MarimerLLC/csla/pull/1226/files) to simulate the fetch. While it just checks the type of the criteria object in those tests, you could instead use the value from the criteria to perform the fetch.

rockfordlhotka:
@luizfbicalho let me know when you have added tests for this

luizfbicalho:
> @luizfbicalho let me know when you have added tests for this

I'll finish them thiis week

luizfbicalho:
> @luizfbicalho let me know when you have added tests for this


I added the tests in a way that I'm validating that sync rules can be called async and worked the same way, and async rules can't be called sync, throwing exception

Source/Csla/Rules/IAuthorizationRuleBase.cs
```diff
@@ -36,4 +31,31 @@ public interface IAuthorizationRule
     /// </summary>
     bool CacheResult { get; }
   }
+
+  /// <summary>
+  /// Interface defining an authorization
+  /// rule implementation.
+  /// </summary>
+  public interface IAuthorizationRule : IAuthorizationRuleBase
```
rockfordlhotka, line 39: This interface should be in its own file

Source/Csla/Rules/IAuthorizationRuleBase.cs
```diff
@@ -36,4 +31,31 @@ public interface IAuthorizationRule
...
+    /// </summary>
+    /// <param name="context">Rule context object.</param>
+    void Execute(IAuthorizationContext context);
+  }
+
+  /// <summary>
+  /// Interface defining an authorization
+  /// rule implementation.
+  /// </summary>
+  public interface IAuthorizationRuleAsync : IAuthorizationRuleBase
```
rockfordlhotka, line 52: This interface also should be in its own file

rockfordlhotka (changes requested):
We need to add an `AuthorizationRuleAsync` class to provide basic implementation of the `IAuthorizationRuleAsync` interface - like the existing `AuthorizationRule` class.

rockfordlhotka (approved):

PR #3967, opened 2024-05-23
Fix for InvalidOperationException in Blazor authz

### LINKED ISSUE #3964: InvalidOperationException in Blazor 8 state manager
The application context manager for Blazor server, in Blazor 8, is unable to get the user identity from the `AuthenticationStateProvider` service, because Blazor 8 no longer allows this type to be accessed outside the context of a razor component.

My research indicates that we should try to access the HttpContext _first_ and if that isn't available, then we should fall back to the `AuthenticationStateProvider` to get the current user identity.

This affects CSLA 8 and 9.

rockfordlhotka:
Closes #3964

#3965

### DIFF
--- a/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
+++ b/Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
@@ -69,10 +69,7 @@ namespace Csla.AspNetCore.Blazor
       {
         var user = HttpContext.User;
         if (user != null)
-        {
           CurrentPrincipal = user;
-          SetHostPrincipal(Task.FromResult(new AuthenticationState(user)));
-        }
       }
       else
       {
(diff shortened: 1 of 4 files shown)


Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -65,40 +65,29 @@ public ApplicationContextManagerBlazor(IHttpContextAccessor hca, AuthenticationS
...
         Task<AuthenticationState> task;
         try
         {
           task = AuthenticationStateProvider.GetAuthenticationStateAsync();
           await task;
         }
-        catch (InvalidOperationException ex)
+        catch (InvalidOperationException)
         {
           task = Task.FromResult(new AuthenticationState(UnauthenticatedPrincipal));
```
StefanOssendorf, line 87: Isn't the `SetHostPrinciapl(task)` now missing in this case and we end up with no user?

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
```diff
@@ -40,14 +41,18 @@ public class ApplicationContextManagerInMemory : IContextManager, IDisposable
...
     /// Creates an instance of the object, initializing it
     /// with the required IServiceProvider.
     /// </summary>
+    /// <param name="httpContextAccessor"></param>
     /// <param name="authenticationStateProvider">AuthenticationStateProvider service</param>
     /// <param name="activeCircuitState"></param>
-    public ApplicationContextManagerInMemory(AuthenticationStateProvider authenticationStateProvider, ActiveCircuitState activeCircuitState)
+    public ApplicationContextManagerInMemory(IHttpContextAccessor httpContextAccessor, AuthenticationStateProvider authenticationStateProvider, ActiveCircuitState activeCircuitState)
     {
+      HttpContext = httpContextAccessor.HttpContext;
```
StefanOssendorf, line 55: Is `IHttpContextAccessor` _always_ registered within the DI? Even in a real Blazor WASM scenario? That looks dangerous.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
```diff
@@ -57,29 +62,30 @@ public ApplicationContextManagerInMemory(AuthenticationStateProvider authenticat
...
         {
-          SetHostPrincipal(task);
+          task = AuthenticationStateProvider.GetAuthenticationStateAsync();
+          await task;
         }
-        else
+        catch (InvalidOperationException)
         {
-          throw;
+          task = Task.FromResult(new AuthenticationState(UnauthenticatedPrincipal));
```
StefanOssendorf, line 85: `SetHostPrincipal` not necessary anymore?

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
```diff
@@ -57,29 +62,30 @@ public ApplicationContextManagerInMemory(AuthenticationStateProvider authenticat
 
     private async Task InitializeUser()
     {
-      Task<AuthenticationState> task = default;
-      try
+      var httpContext = HttpContext;
```
StefanOssendorf, line 65: why the variable? It can't change during life time of this objects instance.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
rockfordlhotka: I don't think it is our job to set the current principal. Blazor 8 appears to keep HttpContext.User and the AuthenticationStateProvider in sync all by itself.

I really think CSLA was overcomplicating some of this stuff earlier.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
rockfordlhotka: It is required by CSLA for other reasons, so yes, it is always registered or the app will fail on startup.

None of this code is run on the wasm client - this is all aspnetcore Blazor server stuff.

PR #3968, opened 2024-05-23
Use HttpContext to get user, then fall back to AuthenticationStateProvider

### LINKED ISSUE #3964: InvalidOperationException in Blazor 8 state manager

rockfordlhotka:
Fixes #3964


StefanOssendorf:
See my comment on #3967 for this PR as well :)

PR #3977, opened 2024-05-25
Add .ConfigureAwait(false) on (thread) context irrelevant operations

StefanOssendorf:
Add some .ConfigureAwait(false) to remove pressure on a sync context when available

Motivation for our WPF application:
We experience a lot of work and pressure on the UI thread due to the current handling of async calls to a remote data portal.
With this PR that should be solved because the low level operations are now free to use a thread pool thread to perform non UI related work.
The PR only adds `.ConfigureAwait(false);` to `await`s where the result _can't_ be known to the UI thread.

@rockfordlhotka Should we consider this a breaking change?

### DIFF
--- a/Source/Csla/DataPortalClient/DataPortalProxy.cs
+++ b/Source/Csla/DataPortalClient/DataPortalProxy.cs
@@ -76,7 +76,7 @@ namespace Csla.DataPortalClient
         request.CriteriaData = SerializationFormatterFactory.GetFormatter(ApplicationContext).Serialize(criteria);
         request = ConvertRequest(request);
         var serialized = SerializationFormatterFactory.GetFormatter(ApplicationContext).Serialize(request);
-        serialized = await CallDataPortalServer(serialized, "create", GetRoutingToken(objectType), isSync);
+        serialized = await CallDataPortalServer(serialized, "create", GetRoutingToken(objectType), isSync).ConfigureAwait(false);
         var response = (DataPortalResponse)SerializationFormatterFactory.GetFormatter(ApplicationContext).Deserialize(serialized);
         response = ConvertResponse(response);
 
@@ -137,7 +137,7 @@ namespace Csla.DataPortalClient
 
         var serialized = SerializationFormatterFactory.GetFormatter(ApplicationContext).Serialize(request);
 
-        serialized = await CallDataPortalServer(serialized, "fetch", GetRoutingToken(objectType), isSync);
+        serialized = await CallDataPortalServer(serialized, "fetch", GetRoutingToken(objectType), isSync).ConfigureAwait(false);
 
         var response = (DataPortalResponse)SerializationFormatterFactory.GetFormatter(ApplicationContext).Deserialize(serialized);
         response = ConvertResponse(response);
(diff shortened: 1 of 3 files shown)


Source/Csla/DataPortalClient/DataPortalProxy.cs
```diff
@@ -76,7 +76,7 @@ public async virtual Task<DataPortalResult> Create(Type objectType, object crite
         request.CriteriaData = SerializationFormatterFactory.GetFormatter(ApplicationContext).Serialize(criteria);
         request = ConvertRequest(request);
         var serialized = SerializationFormatterFactory.GetFormatter(ApplicationContext).Serialize(request);
-        serialized = await CallDataPortalServer(serialized, "create", GetRoutingToken(objectType), isSync);
+        serialized = await CallDataPortalServer(serialized, "create", GetRoutingToken(objectType), isSync).ConfigureAwait(false);
```
rockfordlhotka, line 79: I'm not sure this will work with Windows Forms. Maybe iffy in WPF too.

The thing is, the proxy gets called by the UI, and the results need to be marshalled onto the UI thread, because they are often directly bound to the UI upon return.

If this works in WinForms/WPF without breaking data binding, then I'm fine with it - but that needs to be explicitly tested.

StefanOssendorf: I'll test this to make sure. But imho this can't be in any data binding. We are still handling a `byte[]` here. So there is no real BO which could be data binded.

rockfordlhotka: I just remember having a lot of trouble getting the HttpProxy working reliably with WinForms and WPF and data binding.

If nothing between the async call and the UI refreshing data binding marshals the results back onto the UI thread/context, the code can remain on the wrong thread/context.

StefanOssendorf: Mhm okay. The context is captured on each await so ultimately as long as the user does not `IDataPortal.FetchAsync().ConfigureAwait(false);` it should be coming back on the correct context.
But yeah. I'll test that again.

rockfordlhotka: The issue might be that the sync APIs delegate through the async path? I don't think that's the case, but we should make sure that sync calls don't break in WinForms either.

StefanOssendorf: Do we have sample project for winforms?
I'm currently testing this change with the app from my workplace and so far no issues there.

rockfordlhotka: Yes, there's the simple one

https://github.com/MarimerLLC/csla/tree/main/Samples/WinFormsExample

And the one that is perhaps more important for this particular test:

https://github.com/MarimerLLC/csla/tree/main/Samples/RootChildGrandchildWinFormTest

rockfordlhotka: @StefanOssendorf please let me know when you have tested and ensured it doesn't cause issues.

StefanOssendorf: Yeah I'm at it. Locally these changes have no impact at all. Because the local proxy overrides these methods. And with an actual http call have I have no issues with our wpf App to navigate and work through our views

StefanOssendorf: @rockfordlhotka I tested this with the rootchildgrandchildwinforms sample and I can't spot a difference from the main version without my changes. And as described in our WPF app it's working fine.

rockfordlhotka (approved):

PR #3979, opened 2024-05-25
Make WaitForIdle() methods virtual so a user can extend it's functionality

StefanOssendorf:
* WaitForIdle() methods are now virtual
* Reorganze WaitForIdle() methods to reduce code duplication

@rockfordlhotka 
During the changes I've become unhappy with the currenty `WaitForIdle()` structure, because a user would now have to override two methods to get consistent behavior. The methods with `TimeSpan` and `CancellationToken` because the first throws a `TimeoutException` and the latter a `TaskCanceledException`. I understand _why_ both implementation use different exceptions but from a user perspective it's awkward to override two methods to have consistent behavior. Any ideas?

### DIFF
--- a/Source/Csla/Core/BusinessBase.cs
+++ b/Source/Csla/Core/BusinessBase.cs
@@ -1240,9 +1240,9 @@ namespace Csla.Core
     /// Await this method to ensure business object is not busy.
     /// </summary>
     /// <param name="timeout">Timeout duration</param>
-    public virtual Task WaitForIdle(TimeSpan timeout)
+    public Task WaitForIdle(TimeSpan timeout)
     {
-      return BusyHelper.WaitForIdle(this, timeout);
+      return BusyHelper.WaitForIdleAsTimeout(() => WaitForIdle(timeout.ToCancellationToken()), this.GetType(), nameof(WaitForIdle), timeout);
     }
 
     /// <summary>
--- a/Source/Csla/Core/BusinessBase.cs
+++ b/Source/Csla/Core/BusinessBase.cs
@@ -1228,6 +1228,33 @@ namespace Csla.Core
     #region Data Access
 
     /// <summary>
+    /// Await this method to ensure business object is not busy.
+    /// </summary>
+    public async Task WaitForIdle()
+    {
+      var cslaOptions = ApplicationContext.GetRequiredService<Configuration.CslaOptions>();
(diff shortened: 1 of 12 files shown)


Source/Csla/Core/BusinessBase.cs
```diff
@@ -1227,6 +1227,33 @@ public virtual BrokenRulesCollection BrokenRulesCollection
...
+    {
+      var cslaOptions = ApplicationContext.GetRequiredService<Configuration.CslaOptions>();
+      await WaitForIdle(TimeSpan.FromSeconds(cslaOptions.DefaultWaitForIdleTimeoutInSeconds)).ConfigureAwait(false);
+    }
+
+    /// <summary>
+    /// Await this method to ensure business object is not busy.
+    /// </summary>
+    /// <param name="timeout">Timeout duration</param>
+    public virtual Task WaitForIdle(TimeSpan timeout)
```
rockfordlhotka, line 1243: Is there a way to have just one virtual overload?

I've done this before, with two virtual overloads that "do the same thing" and it becomes a serious point of confusion and bugs for people who overload the method(s) over time.

StefanOssendorf: Yeah. I'll think about a proper solution :)

Source/Csla/Core/BusinessBase.cs
```diff
@@ -1240,9 +1240,9 @@ public async Task WaitForIdle()
     /// Await this method to ensure business object is not busy.
     /// </summary>
     /// <param name="timeout">Timeout duration</param>
-    public virtual Task WaitForIdle(TimeSpan timeout)
+    public Task WaitForIdle(TimeSpan timeout)
```
rockfordlhotka, line 1243: If this overload doesn't invoke the virtual overload, then the user can't replace this implementation? Only the cancelation token overload?

Can't both public overloads maybe invoke a `protected virtual` that someone can override?

rockfordlhotka (changes requested):

Source/Csla/Core/BusinessBase.cs
StefanOssendorf: Huh? The first Parameter is using the `WaitForIdle(CancellationToken)` method so we have only one method to override.

rockfordlhotka: Line 1245 confused me. I got it now.

rockfordlhotka (approved):

PR #3982, opened 2024-05-27
Improve/Enhance Is(Self)Busy to handle nested busy scenarios better

### LINKED ISSUE #3954: Improve/Enhance Is(Self)Busy to handle nested busy scenarios better
Currently `Is(Self)Busy` is only a `bool`. This _must_ not change with respect to the public api.
But handling that bool should be instead an integer to count busyness and non busyness.
Something like the following would today end the "busyness" early eventhogh it's not expected/desired:
```csharp
public void Foo()
{
  try
  {
    MarkAsBusy();
    // more locs
    Bar();
    // more locs
  }
  finally
 {
    MarkAsIdle();
 }
}

public void Bar()
{
  // Do stuff here
  MarkAsIdle();
}
```
With such a structure the busyness would end early. With an internal counter we could more easily handle the busy flag.
Something like this:
```csharp
private int _isBusyCounter;

MarkAsBusy()
{
  Interlocked.Increment(ref _isBusyCounter);
  // Other stuff the implementation needs
}

MarkAsIdle()
{
  Interlocked.Decrement(ref _isBusyCounter);
  // Other stuff the implementation needs
}
```
With this as long as the dev ensure he's calling MarkBusy() and MarkIdle() in pairs the busy state will always be correct without worrying about where the method is used.

Freelancingonupwork:
Busy flag handled by an internal counter.

Fixes #3954

### DIFF
--- a/Source/Csla.test/BusinessBase/BusynessTests.cs
+++ b/Source/Csla.test/BusinessBase/BusynessTests.cs
@@ -25,28 +25,22 @@ namespace cslalighttest.BusyStatus
       _testDIContext = TestDIContextFactory.CreateDefaultContext();
     }
 
-    [TestInitialize]
-    public void Initialize()
-    {
-      TestResults.Reinitialise();
-    }
-
     [TestMethod]
-    public void ObjectIsNotBusy()
+    public void IsBusyWhenHavingTheSameInvocationsOfMarkBusyAndMarkIdleItShouldBeInAnIdleStateAgain()
     {
       UnitTestContext context = GetContext();
       var root = CreateWithoutCriteria<ObjectBusy>();
-      root.One();
+      root.IdealStateSubMethod();
       Assert.IsFalse(root.IsBusy);
     }
 
     [TestMethod]
-    public void ObjectIsBusy()
+    public void IsBusyWhenHavingTheInvocationsOfMarkBusyAndMarkIdleItShouldNotBeInAnIdleStateAgain()
     {
       UnitTestContext context = GetContext();
       var root = CreateWithoutCriteria<ObjectBusy>();
 
-      root.Three();
+      root.NonIdealStateMethod();
       Assert.IsTrue(root.IsBusy);
     }
 
@@ -59,12 +53,12 @@ namespace cslalighttest.BusyStatus
 
       }
 
-      public void One()
+      public void IdealStateMethod()
       {
         try
         {
           MarkBusy();
-          Two();
+          IdealStateSubMethod();
         }
         finally
         {
@@ -72,7 +66,7 @@ namespace cslalighttest.BusyStatus
         }
       }
 
-      public void Two()
+      public void IdealStateSubMethod()
       {
         try
         {
@@ -84,7 +78,7 @@ namespace cslalighttest.BusyStatus
         }
       }
 
-      public void Three()
+      public void NonIdealStateMethod()
       {
         MarkBusy();
         MarkBusy();
--- a/Source/Csla.test/BusinessBase/BusynessTests.cs
+++ b/Source/Csla.test/BusinessBase/BusynessTests.cs
@@ -1,0 +1,102 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="BusynessTests.cs" company="Marimer LLC">
+//     Copyright (c) Marimer LLC. All rights reserved.
+//     Website: https://cslanet.com
+// </copyright>
+// <summary>no summary</summary>
+//-----------------------------------------------------------------------
+using Csla;
+using UnitDriven;
+using Csla.TestHelpers;
+using Csla.Test;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace cslalighttest.BusyStatus
+{
+  [TestClass]
+  public class BusynessTests : TestBase
+  {
+
+    private static TestDIContext _testDIContext;
+
+    [ClassInitialize]
+    public static void ClassInitialize(TestContext context)
+    {
+      _testDIContext = TestDIContextFactory.CreateDefaultContext();
+    }
+
+    [TestInitialize]
+    public void Initialize()
+    {
+      TestResults.Reinitialise();
+    }
+
+    [TestMethod]
+    public void ObjectIsNotBusy()
+    {
+      UnitTestContext context = GetContext();
+      var root = CreateWithoutCriteria<ObjectBusy>();
+      root.One();
+      Assert.IsFalse(root.IsBusy);
+    }
+
+    [TestMethod]
+    public void ObjectIsBusy()
+    {
+      UnitTestContext context = GetContext();
+      var root = CreateWithoutCriteria<ObjectBusy>();
+
+      root.Three();
+      Assert.IsTrue(root.IsBusy);
+    }
+
+    [Serializable]
+    public class ObjectBusy : BusinessBase<ObjectBusy>
+    {
+      [Create]
+      private void Create()
+      {
+
+      }
+
+      public void One()
+      {
+        try
+        {
+          MarkBusy();
+          Two();
+        }
+        finally
+        {
+          MarkIdle();
+        }
+      }
+
+      public void Two()
+      {
+        try
+        {
+          MarkBusy();
+        }
+        finally
+        {
+          MarkIdle();
+        }
+      }
+
+      public void Three()
+      {
+        MarkBusy();
+        MarkBusy();
+        MarkIdle();
+      }
+    }
+
+    private T CreateWithoutCriteria<T>()
+    {
+      IDataPortal<T> dataPortal = _testDIContext.CreateDataPortal<T>();
+
+      return dataPortal.Create();
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


Source/Csla/Core/BusinessBase.cs
```diff
@@ -3018,9 +3018,11 @@ protected void LoadPropertyAsync<R>(PropertyInfo<R> property, Task<R> factory)
...
     #region IsBusy / IsIdle
 
-    [NonSerialized]
-    [NotUndoable]
-    private bool _isBusy;
+    private int _isBusyCounter;
+    /// <summary>
+    /// Indicates object is busy or not 
+    /// </summary>
+    public bool _isBusy => _isBusyCounter > 0;
```
StefanOssendorf, line 3025: This implementation has a couple of issues:
1. It should _not_ introduce a new property but replace the old bool field with a counter.
2. The MarkBusy/Idle methods do not work in a parallel scenario because you are comparing the field and not the returned value from  `Interlocked`.
3. There are more objects which havca `Is(Self)Busy` implementation which should also be changed

StefanOssendorf (changes requested):

StefanOssendorf:
Another note: Please do not resolve the conversations. You can add a simple "Fixed/Done" as a a comment. But we decide whether this conversation is resolved/obsolete. It makes our work a lot easier. Thank you.

Source/Csla.test/BusinessBase/BusynessTests.cs
```diff
@@ -0,0 +1,102 @@
...
+    [ClassInitialize]
+    public static void ClassInitialize(TestContext context)
+    {
+      _testDIContext = TestDIContextFactory.CreateDefaultContext();
+    }
+
+    [TestInitialize]
+    public void Initialize()
+    {
+      TestResults.Reinitialise();
```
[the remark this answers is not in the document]
Freelancingonupwork: it's not necessary. I have removed it.

Source/Csla.test/BusinessBase/BusynessTests.cs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    [TestMethod]
+    public void ObjectIsNotBusy()
```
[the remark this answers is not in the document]
Freelancingonupwork: Name changed.

Source/Csla.test/BusinessBase/BusynessTests.cs
```diff
@@ -0,0 +1,102 @@
...
+    [Serializable]
+    public class ObjectBusy : BusinessBase<ObjectBusy>
+    {
+      [Create]
+      private void Create()
+      {
+
+      }
+
+      public void One()
```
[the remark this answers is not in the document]
Freelancingonupwork: I have changed the names of Methods.

Source/Csla/Core/BusinessBase.cs
Freelancingonupwork: 1. It should not introduce a new property but replace the old bool field with a counter. - **Have changed**. 

2. The MarkBusy/Idle methods do not work in a parallel scenario because you are comparing the field and not the returned value from Interlocked. - **Have changed**. 

3. There are more objects which have Is(Self)Busy implementation which should also be changed. - I have changed in `BusinessBase.cs` and `ReadOnlyBase.cs`.

I have found Is(Self)Busy fields in other class `ObservableBindingList.cs` and `ExtendedBindingList.cs` but, there 

```
 public virtual bool IsBusy
 {
   get { return false; }
 }
public virtual bool IsSelfBusy
{
  get { return IsBusy; }
}
```
------------------------------
```
public virtual bool IsBusy
{
  get { throw new NotImplementedException(); }
}
public virtual bool IsSelfBusy
{
  get { return IsBusy; }
}
```

Do I need to change Logic in these class? if yes then I think I have to create sub milestone. let me know if i am missing any classes.

StefanOssendorf (changes requested):

Source/Csla/Core/BusinessBase.cs
```diff
@@ -3043,8 +3044,12 @@ protected void MarkBusy()
     [EditorBrowsable(EditorBrowsableState.Advanced)]
     protected void MarkIdle()
     {
-      _isBusy = false;
-      OnBusyChanged(new BusyChangedEventArgs("", false));
+      int updatedValue = Interlocked.Decrement(ref _isBusy);
+
```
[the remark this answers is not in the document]
Freelancingonupwork: Fixed. Please Check.

Source/Csla/Core/BusinessBase.cs
```diff
@@ -3069,7 +3074,7 @@ public virtual bool IsBusy
     [ScaffoldColumn(false)]
     public virtual bool IsSelfBusy
     {
-      get { return _isBusy || BusinessRules.RunningAsyncRules || LoadManager.IsLoading; }
+      get { return (_isBusy > 0 ? true : false) || BusinessRules.RunningAsyncRules || LoadManager.IsLoading; }
```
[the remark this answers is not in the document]
Freelancingonupwork: Fixed. Please Check.

Source/Csla/Core/BusinessBase.cs
```diff
@@ -3043,8 +3044,16 @@ protected void MarkBusy()
     [EditorBrowsable(EditorBrowsableState.Advanced)]
     protected void MarkIdle()
     {
-      _isBusy = false;
-      OnBusyChanged(new BusyChangedEventArgs("", false));
+      int updatedValue = Interlocked.Decrement(ref _isBusy);
+      if (updatedValue < 0)
+      {
+        _ = Interlocked.CompareExchange(ref _isBusy, 0, updatedValue);
+         updatedValue = 0;
```
StefanOssendorf, line 3051: That's not correct. With this you would trigger a PropertyChanged even though the state did not change. Keep the old updated value for the rest of the method. That's fine and desired.

Source/Csla/ReadOnlyBase.cs
```diff
@@ -1520,8 +1521,16 @@ protected void MarkBusy()
     [EditorBrowsable(EditorBrowsableState.Advanced)]
     protected void MarkIdle()
     {
-      _isBusy = false;
-      OnBusyChanged(new BusyChangedEventArgs("", false));
+      int updatedValue = Interlocked.Decrement(ref _isBusy);
+      if (updatedValue < 0)
+      {
+        _ = Interlocked.CompareExchange(ref _isBusy, 0, updatedValue);
+        updatedValue = 0;
```
StefanOssendorf, line 1528: That's not correct. With this you would trigger a PropertyChanged even though the state did not change. Keep the old updated value for the rest of the method. That's fine and desired.

Source/Csla/Core/BusinessBase.cs
```diff
@@ -3017,10 +3017,10 @@ protected void LoadPropertyAsync<R>(PropertyInfo<R> property, Task<R> factory)
...
 
     #region IsBusy / IsIdle
-
-    [NonSerialized]
-    [NotUndoable]
-    private bool _isBusy;
+    /// <summary>
+    /// Indicates object is busy or not 
+    /// </summary>
+    private int _isBusy;
```
StefanOssendorf, line 3023: Sorry my bad here. Please name it `_isBusyCounter` and add the attributes back.

Source/Csla/ReadOnlyBase.cs
```diff
@@ -1497,7 +1497,7 @@ public Task WaitForIdle(CancellationToken ct)
 
     [NonSerialized]
     [NotUndoable]
-    private bool _isBusy;
+    private int _isBusy;
```
StefanOssendorf, line 1500: Sorry my bad here. Please name it `_isBusyCounter`

StefanOssendorf (changes requested):

Source/Csla/Core/BusinessBase.cs
Freelancingonupwork: Fixed

Source/Csla/ReadOnlyBase.cs
Freelancingonupwork: Fixed

Source/Csla/Core/BusinessBase.cs
Freelancingonupwork: Done

Source/Csla/Core/BusinessBase.cs
```diff
@@ -3048,10 +3048,10 @@ protected void LoadPropertyAsync<R>(PropertyInfo<R> property, Task<R> factory)
...
 
     #region IsBusy / IsIdle
-
-    [NonSerialized]
-    [NotUndoable]
-    private bool _isBusy;
+    /// <summary>
+    /// Indicates object is busy or not 
+    /// </summary>
+    private int _isBusyCounter;
```
StefanOssendorf, line 3054: The old attributes
```csharp
[NonSerialized]
[NotUndoable]
```
are still missing.

StefanOssendorf (changes requested):

Source/Csla/Core/BusinessBase.cs
Freelancingonupwork: Done

StefanOssendorf:
Looks good to me! I'll check later when I'm at my desktop PC.

StefanOssendorf (approved):

PR #3984, opened 2024-05-28
Added - GraphMerger: async versions of public methods

### LINKED ISSUE #3976: GraphMerger: Add async versions of public methods
Currently the `GraphMerger` only has sync merge versions. Which work perfectly fine when the BO only has sync rules.
To make the merger more async friendly it should get three new async versions of these public methods:
* https://github.com/MarimerLLC/csla/blob/75e17df998d2bcbc35d258dd3033f021499738fd/Source/Csla/Core/GraphMerger.cs#L29
* https://github.com/MarimerLLC/csla/blob/75e17df998d2bcbc35d258dd3033f021499738fd/Source/Csla/Core/GraphMerger.cs#L151
* https://github.com/MarimerLLC/csla/blob/75e17df998d2bcbc35d258dd3033f021499738fd/Source/Csla/Core/GraphMerger.cs#L181

The new async versions must use `CheckRulesAsync` where the current version is using the sync method. 
The current sync version _must_ work as before. 
Furthermore it should be checked where the sync merge methods are used within the framework.
If and only if the method using the merge method is already async it should be changed to the new async api.

Freelancingonupwork:
Added async versions of public methods in `GraphMerger.cs`

Fixes #3976

### DIFF
--- a/Source/Csla/Core/GraphMerger.cs
+++ b/Source/Csla/Core/GraphMerger.cs
@@ -129,16 +129,27 @@ namespace Csla.Core
     /// </summary>
     /// <param name="target">Target of merge.</param>
     /// <param name="source">Source for merge.</param>
-    private void MergeGraph(IEditableCollection target, IEditableCollection source)
+    /// <param name="isAsync">Called from sync or async method.</param>
+    private void MergeGraph(IEditableCollection target, IEditableCollection source, bool isAsync = false)
     {
       var listType = target.GetType();
       var childType = Utilities.GetChildItemType(listType);
       var genericTypeParams = new Type[] { listType, childType };
       System.Reflection.MethodInfo methodReference;
-      if (typeof(IExtendedBindingList).IsAssignableFrom(listType))
-        methodReference = GetType().GetMethod("MergeBusinessBindingListGraph");
+      if(isAsync)
+      {
+        if (typeof(IExtendedBindingList).IsAssignableFrom(listType))
+          methodReference = GetType().GetMethod("MergeBusinessBindingListGraphAsync");
+        else
+          methodReference = GetType().GetMethod("MergeBusinessListGraphAsync");
+      }
       else
-        methodReference = GetType().GetMethod("MergeBusinessListGraph");
+      {
+        if (typeof(IExtendedBindingList).IsAssignableFrom(listType))
+          methodReference = GetType().GetMethod("MergeBusinessBindingListGraph");
+        else
+          methodReference = GetType().GetMethod("MergeBusinessListGraph");
+      }
       var gr = methodReference.MakeGenericMethod(genericTypeParams);
       gr.Invoke(this, [target, source]);
     }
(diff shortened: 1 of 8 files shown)


Source/Csla/Core/GraphMerger.cs
```diff
@@ -202,5 +213,155 @@ private void MergeGraph(IEditableCollection target, IEditableCollection source)
...
+          else
+          {
+            if (sourceValue is IEditableCollection sourceList)
+            {
+              var targetFieldExists = true;
+              if (targetFieldManager != null && (item.RelationshipType & RelationshipTypes.LazyLoad) == RelationshipTypes.LazyLoad)
+                targetFieldExists = targetFieldManager.FieldExists(item);
+              if (targetFieldExists && ReadProperty(target, item) is IEditableCollection targetList)
+              {
+                MergeGraph(targetList, sourceList,true);
```
rockfordlhotka, line 269: Doesn't this defeat the purpose? This is a sync call, and the modification to `MergeGraph` presumably does a sync call even to the methods ending in `Async`?

Source/Csla/Core/GraphMerger.cs
```diff
@@ -129,16 +129,27 @@ private static void CopyField(object source, object target, string fieldName)
...
       var listType = target.GetType();
       var childType = Utilities.GetChildItemType(listType);
       var genericTypeParams = new Type[] { listType, childType };
       System.Reflection.MethodInfo methodReference;
-      if (typeof(IExtendedBindingList).IsAssignableFrom(listType))
-        methodReference = GetType().GetMethod("MergeBusinessBindingListGraph");
+      if(isAsync)
+      {
+        if (typeof(IExtendedBindingList).IsAssignableFrom(listType))
+          methodReference = GetType().GetMethod("MergeBusinessBindingListGraphAsync");
```
rockfordlhotka, line 142: I see where you find the methods ending with `Async`, but not where you are invoking them in an async manner? Are they still being invoked as sync methods?

rockfordlhotka (changes requested):

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: My bad. I have changed the code.

Source/Csla/Core/GraphMerger.cs
```diff
@@ -129,27 +129,16 @@ private static void CopyField(object source, object target, string fieldName)
     /// </summary>
     /// <param name="target">Target of merge.</param>
     /// <param name="source">Source for merge.</param>
-    /// <param name="isAsync">Called from sync or async method.</param>
-    private void MergeGraph(IEditableCollection target, IEditableCollection source, bool isAsync = false)
+    private void MergeGraph(IEditableCollection target, IEditableCollection source)
```
rockfordlhotka, line 132: Thank you for these changes.

I still think we probably need an async `MergeGraph` method that merges collections by awaiting async versions of the `MergeBusinessListXYZ` methods.

Otherwise, if I'm implementing those methods in a collection, how would I invoke async rule checking or other behaviors on my child objects?

rockfordlhotka (changes requested):

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: Ok. let me check.

Freelancingonupwork: I have done some changes and added MergeGraphAsync method.

Source/Csla/Core/GraphMerger.cs
```diff
@@ -202,5 +202,173 @@ private void MergeGraph(IEditableCollection target, IEditableCollection source)
...
+            if (targetFieldExists && ReadProperty(target, item) is IEditableBusinessObject targetChild)
+            {
+              await MergeGraphAsync(targetChild, sourceChild);
+            }
+            else
+            {
+              if ((item.RelationshipType & RelationshipTypes.PrivateField) == RelationshipTypes.PrivateField)
+                Reflection.MethodCaller.CallPropertySetter(target, item.Name, sourceChild);
+              else
+                await Task.Run(() => LoadProperty(target, item, sourceChild));
```
StefanOssendorf, line 246: Why is there a `Task.Run`? Is there any reason for it? I don't think here should be _any_ `Task.Run` be used. That even hurts performance.

Source/Csla/Core/GraphMerger.cs
```diff
@@ -202,5 +202,173 @@ private void MergeGraph(IEditableCollection target, IEditableCollection source)
...
+      }
+    }
+    /// <summary>
+    /// Merges state from source graph into target graph.
+    /// </summary>
+    /// <param name="target">Target of merge.</param>
+    /// <param name="source">Source for merge.</param>
+    private async Task MergeGraphAsync(IEditableCollection target, IEditableCollection source)
+    {
+      var listType = await Task.Run(() => target.GetType());
```
StefanOssendorf, line 301: Remove all `Task.Run`.

Source/Csla/Core/GraphMerger.cs
```diff
@@ -202,5 +202,173 @@ private void MergeGraph(IEditableCollection target, IEditableCollection source)
...
+      var childType = await Task.Run(() => Utilities.GetChildItemType(listType));
+      var genericTypeParams = new Type[] { listType, childType };
+      System.Reflection.MethodInfo methodReference;
+      if (typeof(IExtendedBindingList).IsAssignableFrom(listType))
```
StefanOssendorf, line 305: Why is here reflection necessary to call the methods defined by this type with `target` and `source`?
When reflection is necessary the method infos should be cached within the instance to improve efficency during recursive calls.

Source/Csla/Core/GraphMerger.cs
```diff
@@ -202,5 +202,173 @@ private void MergeGraph(IEditableCollection target, IEditableCollection source)
...
+        methodReference = await Task.Run(() => GetType().GetMethod("MergeBusinessBindingListGraphAsync"));
+      else
+        methodReference = await Task.Run(() => GetType().GetMethod("MergeBusinessListGraphAsync"));
+      var gr = await Task.Run(() => methodReference.MakeGenericMethod(genericTypeParams));
+      await Task.Run(() => gr.Invoke(this, [target, source]));
```
StefanOssendorf, line 310: This invocation must be proper async. The current implementation does nto wait for `gr.Invoke()` to complete.

StefanOssendorf (changes requested):
Note: Please do not mark the conversations as resovled. Only write fixed/done as a comment and during the next review we'll either resolve it or make another comment for further improvements.

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: ok

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: Reflection is being used in MergeGraph/MergeGraphAsync method to dynamically invoke either the MergeBusinessListGraph/MergeBusinessListGraphAsync or MergeBusinessBindingListGraph/MergeBusinessBindingListGraphAsync method based on the runtime type of the target collection. 

"dynamic" keyword is not supported below .Net 4.x Compatibility projects. Same with delegate.

This will be possible implementation using cache. can you please review so I can commit.

```
    private async Task MergeGraphAsync(IEditableCollection target, IEditableCollection source)
    {
      var listType = target.GetType();
      var childType = Utilities.GetChildItemType(listType);
      var isExtendedBindingList = typeof(IExtendedBindingList).IsAssignableFrom(listType);

      var cacheKey = (listType, childType, isExtendedBindingList);

      if (!_methodCache.TryGetValue(cacheKey, out System.Reflection.MethodInfo methodReference))
      {
        var methodName = isExtendedBindingList ? "MergeBusinessBindingListGraphAsync" : "MergeBusinessListGraphAsync";
        methodReference = GetType().GetMethod(methodName).MakeGenericMethod(listType, childType);
        _methodCache.TryAdd(cacheKey, methodReference);
      }

      var task = (Task)methodReference.Invoke(this, new object[] { target, source });
      await task;
    }
```

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: noted

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: noted.

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: Fixed. Please check.

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: Fixed. Please check.

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: Fixed. Please check.

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: Fixed. Please check.

Source/Csla/Core/GraphMerger.cs
```diff
@@ -202,5 +205,178 @@ private void MergeGraph(IEditableCollection target, IEditableCollection source)
...
+        }
+        await CheckRulesAsync(target);
+      }
+    }
+    /// <summary>
+    /// Merges state from source graph into target graph.
+    /// </summary>
+    /// <param name="target">Target of merge.</param>
+    /// <param name="source">Source for merge.</param>
+    private async Task MergeGraphAsync(IEditableCollection target, IEditableCollection source)
```
StefanOssendorf, line 302: What about rewriting this method as
```csharp
var cacheKey = GetCacheKey(target);

var methodReference = _methodCache.GetOrAdd(cacheKey, ((Type ListType, Type ChildType, bool IsExtendedBindingList) key) =>
{
  var methodName = key.IsExtendedBindingList ? "MergeBusinessBindingListGraphAsync" : "MergeBusinessListGraphAsync";
  return GetType().GetMethod(methodName).MakeGenericMethod(key.ListType, key.ChildType);
});

var task = (Task)methodReference.Invoke(this, [target, source]);
await task;

static (Type ListType, Type ChildType, bool IsExtendedBindingList) GetCacheKey(IEditableCollection target)
{
  var listType = target.GetType();
  var childType = Utilities.GetChildItemType(listType);
  var isExtendedBindingList = typeof(IExtendedBindingList).IsAssignableFrom(listType);
  return (listType, childType, isExtendedBindingList);
}
```

StefanOssendorf (commented):
What do you think about the method (imho) cleanup?
Otherwise I'm okay with merging :)

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: Fixed

StefanOssendorf: Nearly. Please remove the now obsolete Variables and the outcommented code.

StefanOssendorf (changes requested):

Source/Csla/Core/GraphMerger.cs
Freelancingonupwork: Done.

StefanOssendorf: Where are the variables from line 304 to 306 used? They should have unused and hence removed.

Freelancingonupwork: Done

StefanOssendorf:
I don't see anything which would hinder us now. LGTM then?
@rockfordlhotka  anything from your side?

rockfordlhotka (approved):

Issue #3712, opened 2024-02-23
Remove .NET 6 support


rockfordlhotka:
When Microsoft deprecates .NET 6, CSLA should also remove it as a build target.

rockfordlhotka:
https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

Freelancingonupwork:
Hello,
Do we need to change `#if !NETSTANDARD2_0 && !NET6_0_OR_GREATER` Condition like this to `#if !NETSTANDARD2_0 && !NET8_0_OR_GREATER` ?

luizfbicalho:
I had already done this item

PR #4017, opened 2024-06-11
Update project to .NET 8.0 and refactor code

luizfbicalho:
Updated the target framework for the projects from .NET 6.0 and .NET 5.0 to .NET 8.0. This includes changes to the `.csproj` files and the conditional compilation symbols in the code. Package references have been updated to their respective versions for .NET 8.0. Removed code blocks and package references specific to .NET 6.0 or .NET Core 3.1. Introduced the use of `AssemblyLoadContext` class for managing the execution context of an assembly. Updated the usage of `System.Runtime.Loader` and `Csla.Runtime` namespaces, type rules and method cache handling, transaction types handling, and `TransactionalTypes` enum to be compatible with .NET 8.0. Updated usage of several classes and `AssemblyLoadContextManager` based on the .NET version.

#3712

### DIFF
--- a/NuGet/Definition/Csla.AspNetCore.NuSpec
+++ b/NuGet/Definition/Csla.AspNetCore.NuSpec
@@ -16,11 +16,6 @@
     <description>UI helpers for using CSLA .NET business types with AspNetCore.</description>
     <language>en-US</language>
     <dependencies>
-      <group targetFramework="net6.0">
-        <dependency id="Csla" version="[4.6.3-Beta10]" />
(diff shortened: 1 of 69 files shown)


rockfordlhotka (changes requested):
This is great @luizfbicalho !

One step remains, and that is to also update the nuspec files in the [NuGet/Definition](https://github.com/MarimerLLC/csla/tree/main/NuGet/Definition) directory.

rockfordlhotka (approved):

Issue #4001, opened 2024-06-01
Remove references to BinaryFormatter and NetDataContractSerializer


rockfordlhotka:
The codebase still has some references to BinaryFormatter and NetDataContractSerializer (NDCS).

These should be removed from the entire codebase.

PR #4018, opened 2024-06-11
Remove binary formatter net data contract serializer

luizfbicalho:
#4001


Source/Csla.test/Serialization/SerializationTests.cs
```diff
@@ -204,6 +204,7 @@ public void SerializableEvents()
       context.Assert.Success();
     }
 
+    [Ignore]
```
rockfordlhotka, line 207: I think this test should be removed.

Source/Csla.test/Serialization/SerializationTests.cs
```diff
@@ -215,19 +216,19 @@ public void TestSerializableEventsActionFails()
...
+        b.Serialize(new MemoryStream(), root);
+        Assert.Fail("Serialization should have thrown an exception");
+      }
+      catch (System.Runtime.Serialization.SerializationException)
+      {
+        // serialization failed as expected
+      }
     }
 
+    [Ignore]
```
rockfordlhotka, line 231: Remove this test

rockfordlhotka (changes requested):

rockfordlhotka (approved):

Issue #4015, opened 2024-06-10
Switch out HybridDictionary for ConcurrentDictionary for Local/Client context


rockfordlhotka:
As per the conversation below, if we switch Local/Client context from the `HybridDictionary` base to `ConcurrentDictionary`, the collections themselves should be usable from multiple concurrent threads with some precautions.

To do this, the concrete type currently used for the `LocalContext` and `ClientContext` properties should be changed to an interface so we can preserve backward compatibility with existing code that uses those types.

---

Ah now I get what you mean. The used baseclass of `ContextDictionary` `HybridDictionary` is not thread safe. That's correct according to the documentation. We could probably move to a more thread safe implementation like you suggested.
Since that would be a big breaking change we have to carefully consider that. 
@rockfordlhotka Your thoughts?
My first thoughts on this are to create a new `IContextDictionary` interface which provides _all_ methods provided by our current dictionary. We have to change the type used in the interfaces but at least we keep the api itself stable and could then make the actual implementation configurable while the current is the default.

_Originally posted by @StefanOssendorf in https://github.com/MarimerLLC/csla/discussions/4012#discussioncomment-9712171_

Bowman74:
@rockfordlhotka Just as a note, I could not create a unified IContextDictionary type, instead there is a different IClientContext and ILocalContext types. They look the same and ContextDictionary implements both of them. This was done for two reasons:

- So they could be overridden independently (nice to have)
- So when we ask the DI container for an instance in the scope, the client context and local context return different dictionary instances

rockfordlhotka:
You are now thinking to inject these into all the different application context manager types?

Until now, each of the different context managers has been responsible for creating and storing the dictionaries, but I suppose they could be considered to be a service - maybe?

There's some interesting stuff in `LocalProxy` when the proxy creates (sometimes) a new DI scope and copies (iirc) the Local/Client context values into the new DI scope. I think that's done with some sort of clone operation, but would have to look to be sure.

In that case, the new application context manager created for the temporary DI scope might _not_ want to get new dictionaries, as it will need the clones.

StefanOssendorf:
How is the creation of the actual context implementation now handled? We could resolve it from the DI. With .Net 8 (IIRC) we could even use the Keyed-DI-Registration feature but would have problems down the road? Or we only support that on .Net 8+?
[...]

PR #4023, opened 2024-06-12
#4015 change context to interfaces

Bowman74:
Changes for implementing #4015 to add a IContextDictionary interface and switch to use the ConcurrentDictionary.


Bowman74:
This PR should be squashed...

Source/Csla/Core/ContextDictionary.cs
```diff
@@ -72,6 +73,101 @@ void IMobileObject.SetChildren(SerializationInfo info, MobileFormatter formatter
...
+    /// <param name="index">The zero-based index in array at which copying begins.</param>
+    /// <exception cref="System.ArgumentNullException">array is null.</exception>
+    /// <exception cref="System.ArgumentOutOfRangeException">index is less than zero.</exception>
+    /// <exception cref="System.ArgumentException">
+    /// array is multidimensional. -or- The number of elements in the source System.Collections.ICollection
+    /// is greater than the available space from index to the end of the destination
+    /// array. -or- The type of the source System.Collections.ICollection cannot be cast
+    /// automatically to the type of the destination array.
+    /// </exception>
+    public void CopyTo(Array array, int index)
```
StefanOssendorf, line 166: I think some of these documentions can just be inherited. Like this one. Instead you could use `/// <inheritdoc cref="System.Collections.ICollection.CopyTo(Array, int)"/>` which uses the documentation of the provided cref.

rockfordlhotka: Oh, that's nice - I didn't know that was possible.

Bowman74: I didn't know that was possible either. That would have saved some time! :)

Bowman74: And yes, it does indeed work...

PR #4024, opened 2024-06-13
Add custom serializers to MobileFormatter

### LINKED ISSUE #1877: Consider creating a workaround for serialization of netfx ClaimsPrincipal
As per this thread: https://github.com/MarimerLLC/csla/discussions/1868#discussioncomment-113577

There appears to be a bug in the netfx implementation of `ClaimsPrincipal` where it can't deserialize from a byte stream created by the same type. The behavior works as expected in netstandard, but not netfx.

Currently `MobileFormatter` uses the `CslaClaimsPrincipal` type to "wrap" any `ClaimsPrincipal` found in an object graph so serialization of the principal/identity is possible. Similarly, `MobileFormatter` has special code to deserialize the `ClaimsPrincipal` byte array to rehydrate the principal/identity during deserialization.

That behavior relies on `ClaimsPrincipal` properly serializing/deserializing itself, and so fails in netfx.

## Possible solution

A possible solution would be to have a netfx-specific bit of code in `MobileFormatter` and/or `CslaClaimsPrincipal` to manually pull out all _settable_ property values from the principal, and all the claims from the identity, and to put them into a serializable temp object graph, which would then be serialized into a byte array. On deserialization, that temp object graph would be rehydrated, and could be used to create a new principal/identity object pair.

Whether this would actually work is unknown. Research is required.

### LINKED ISSUE #2148: Enable serialization of POCO types via MobileFormatter

### LINKED ISSUE #2531: MobileFormatter should allow custom serializers for specific types
What @TheCakeMonster and I have been discussing is a modification to `MobileFormatter` where you can provide a serializer for specific types. You'd register these serializers during app startup (via DI), and when `MobileFomatter` encounters one of these types, it would delegate the value to your serializer to convert it to/from a string (or byte array).

I think that if we have `MobileFormatter` rely on a dictionary of `<targettype, serializertype>` that can be defined during app startup, and the dictionary injected to `MobileFormatter` (or something) that we'd have a good/flexible solution.

Then, if we want to use an attribute-based model, the attribute can be used to help load the dictionary, but also, types that can't be decorated with the attribute can be directly added to the dictionary during app startup.

Inside `MobileFormatter`, if a type isn't normally serializable, the formatter will check the dictionary to see if a custom serializer is registered for the target type. If so, the formatter will ask DI for an instance of the custom serializer, and let it serialize/deserialize the target object.

```c#
public interface ICustomSerializer
{
  byte[] Serialize(object target);
  object Deserialize(byte[] serializedData);
}
```

I'm not 100% sure if this should use `byte[]` or `string` - need to review what will be simplest within `MobileFormatter` at the point where a custom serializer would be implemented.

rockfordlhotka:
Closes #2531
Closes #1877
Closes #2148


Source/Csla/Serialization/Mobile/CustomSerializers/ClaimsPrincipalSerializer.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Security.Claims;
+
+namespace Csla.Serialization.Mobile.CustomSerializers;
+
+/// <summary>
+/// A formatter that can serialize and deserialize a ClaimsPrincipal object.
+/// </summary>
+public class ClaimsPrincipalSerializer : IMobileSerializer
+{
+  /// <inheritdoc cref="IMobileSerializer.Deserialize(SerializationInfo)"/>
```
StefanOssendorf, line 18: In this case you don't need the `cref` stuff because it's the method from the interface.

Source/Csla/ApplicationContext.cs
```diff
@@ -187,7 +187,7 @@ public enum ExecutionLocations
     /// for all explicit object serialization (such as cloning,
     /// n-level undo, etc).
     /// </summary>
-    public static Type SerializationFormatter { get; internal set; } = typeof(Serialization.Mobile.MobileFormatter);
+    public static Type SerializationFormatter { get; internal set; } = null;
```
StefanOssendorf, line 190: Is there a reason why this is not put properly into the DI now with the rework?

Source/Csla/Serialization/Mobile/CustomSerializers/ClaimsPrincipalSerializer.cs
rockfordlhotka: So just the inherit tag by itself?

Source/Csla/ApplicationContext.cs
rockfordlhotka: I'll look. The formatter factory is, iirc, a static, so the question will be whether every place in the framework that uses the factory will have access to DI.

Source/Csla/Serialization/Mobile/CustomSerializers/ClaimsPrincipalSerializer.cs
StefanOssendorf: Yes. ```// <inheritdoc />``` should be sufficient.

Source/Csla/ApplicationContext.cs
rockfordlhotka: Yes, doable. Then the only static property left is one that can probably be removed, as it is to allow fallback from expression trees to reflection for method calling, and that capability hasn't been necessary since Blazor first came out without expression trees.

Source/Csla/Serialization/Mobile/MobileFormatter.cs
```diff
@@ -117,6 +111,15 @@ private void ConvertEnumsToIntegers()
       }
     }
 
+    private MobileFormatterOptions GetOptions()
+    {
+      var cslaOptions = applicationContext.GetRequiredService<CslaOptions>();
+      var options = (MobileFormatterOptions)cslaOptions.SerializationOptions.FormatterOptions;
```
StefanOssendorf, line 117: Wouldn't be `cslaOptions.SerializationOptions.FormatterOptions as MobileFormatterOptions` be better?
With that you could react to an unexpected type instead of causing a cast exception here.

Source/Csla/Serialization/Mobile/MobileFormatter.cs
```diff
@@ -117,6 +111,15 @@ private void ConvertEnumsToIntegers()
...
+      if (options is null)
+        throw new ArgumentNullException("MobileFormatterOptions");
```
StefanOssendorf, line 119: imho should be `InvalidOperationException` because it's not an argument anywhere.

Source/Csla/Serialization/Mobile/MobileFormatter.cs
```diff
@@ -256,33 +270,43 @@ public object DeserializeAsDTO(List<SerializationInfo> deserialized)
...
-            var mobile = new Security.CslaClaimsPrincipal(reader);
-            _deserializationReferences.Add(info.ReferenceId, mobile);
+            if (applicationContext.CreateInstanceDI(serializerType) is IMobileSerializer serializer)
+            {
+              object mobile = serializer.Deserialize(info);
+              _deserializationReferences.Add(info.ReferenceId, mobile);
+            }
+            else
+            {
+              throw new InvalidOperationException($"{serializerType?.Name ?? "null"} != IMobileSerializer");
```
StefanOssendorf, line 294: `serializerType` is here `not null` due to the if guaranteeing that.

Source/Csla/Serialization/Mobile/MobileFormatter.cs
```diff
@@ -256,33 +270,43 @@ public object DeserializeAsDTO(List<SerializationInfo> deserialized)
...
       }
 
       foreach (SerializationInfo info in deserialized)
       {
-        IMobileObject mobile = _deserializationReferences[info.ReferenceId];
-
-        mobile?.SetChildren(info, this);
+        var obj = _deserializationReferences[info.ReferenceId];
+        if (obj is IMobileObject mobile)
+          mobile?.SetChildren(info, this);
```
StefanOssendorf, line 309: `mobile` is `not null` here because `null is <AnyType>` is always `false`.

Source/Csla/Serialization/Mobile/MobileFormatter.cs
```diff
@@ -334,7 +358,7 @@ public IMobileObject GetObject(int referenceId)
     public byte[] SerializeToByteArray(object obj)
     {
       using var buffer = new MemoryStream();
-      var formatter = new MobileFormatter(_applicationContext);
+      var formatter = new MobileFormatter(applicationContext);
```
StefanOssendorf, line 361: Note: This looks weird. We have to create a new instance of the type we are currently using to serialize the provided object. Is that necessary due to the `fields` or legacy?

Source/Csla/ApplicationContext.cs
StefanOssendorf: Sounds awesome. Should make some tests also easier due to less shared statics :)

Source/Csla/Serialization/Mobile/MobileFormatter.cs
rockfordlhotka: Hmm. I might be that this used to be a static method and now it is an instance method? I'll research.

Source/Csla/ApplicationContext.cs
rockfordlhotka: Here's a very simple custom serializer so you can see they work:

https://github.com/MarimerLLC/csla/blob/88d55a96b039df15f9867ab6542c0ee446df5609/Source/Csla.test/Serialization/CustomSerializerTests.cs#L109

You do have me rethinking how a type is mapped to a serializer though. Right now I'm doing it with a direct type comparison, not counting inheritance or interfaces. I suppose I could broaden it so a serializer gets registered for types that can be derived from the type in the map?

Right now:

* MyType => MyTypeSerializer

But maybe:

* IMyType => MyTypeSerializer
* MyTYpe : IMyType=> MyTypeSerializer
* SomeotherType : IMyType => MyTypeSerializer

rockfordlhotka:
Here's a very simple custom serializer so you can see they work:

https://github.com/MarimerLLC/csla/blob/88d55a96b039df15f9867ab6542c0ee446df5609/Source/Csla.test/Serialization/CustomSerializerTests.cs#L109

You do have me rethinking how a type is mapped to a serializer though. Right now I'm doing it with a direct type comparison, not counting inheritance or interfaces. I suppose I could broaden it so a serializer gets registered for types that can be derived from the type in the map?

Right now:

* MyType => MyTypeSerializer

But maybe:

* IMyType => MyTypeSerializer
* MyTYpe : IMyType=> MyTypeSerializer
* SomeotherType : IMyType => MyTypeSerializer

Source/Csla/ApplicationContext.cs
rockfordlhotka: There should be no shared statics anymore, unless there's still a test for that old reflection code - but even if that exists it would be highly specialized and isolated.

rockfordlhotka:
@StefanOssendorf I think this is ready for review if you'd like to take a look.

Source/Csla/Serialization/Mobile/Configuration/MobileFormatterOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+// <summary>Options for MobileFormatter configuration</summary>
+//-----------------------------------------------------------------------
+using Csla.Serialization.Mobile;
+
+namespace Csla.Configuration;
+
+/// <summary>
+/// Options for MobileFormatter configuration
+/// </summary>
+public class MobileFormatterOptions
```
StefanOssendorf, line 15: Somehow I don't think we (should) need this. We should start retrieving the stuff we need from the DI. That makes it a lot easier for user to replace stuff as they deem fit without us to need to think of such scenarios in preparations.
With that said I would suggest to replace the configuration from the current usage (see the tests for the `AddMobileFormatter()` extensions) with something like this:
```csharp
services.AddCsla(o => o.Serialization(
  s => s.AddTypeSerializer<Custom1>().AddTypeSerializer<Custom2>()
            .UseReader<CustomReader>().UseWriter<CustomWriter>()
            .UseSerializer<MyCustomSerializer>()
           [.DontAddDefaultTypeSerializers()?]
)
```
And instead of carrying the implementation type within some options object we should just add everything directly into the DI.
For example: `AddCsla()` adds the `MobileFormatter` as itself _and_ `ISerializationFormatter` implementation. Same goes for the type serializers. We can just add them _all_ into the DI. `MobileFormatter` can retrieve them via `.GetRequiredService<IEnumerable<ITypeMap>>();`.
That would _enable_  user to reuse type map implementations added by us and provide their own.

For the `MobileFormatter` within we _need_ the type. Regardless what the user added to the DI because we have a lot of methods which are dependent on the type instead of it's interface.

Another thought for `AddTypeSerializer` we can have the following signatures:
* `AddTypeSerializer<T>() where T: TypeMap<T, S>`
* `AddTypeSerializer<T, S>() where S: IMobileSerializer`
* ...

Source/Csla/Serialization/Mobile/CustomSerializers/PocoSerializer.cs
```diff
@@ -0,0 +1,40 @@
...
+/// </summary>
+public class PocoSerializer<T> : IMobileSerializer
+{
+  /// <inheritdoc />
+  public static bool CanSerialize(Type type) => type == typeof(T);
+
+  /// <inheritdoc />
+  public object Deserialize(SerializationInfo info)
+  {
+    var json = info.GetValue<string>("s");
```
StefanOssendorf, line 25: I'm not that deep in the (de-)serialization process but the same key is used by the `ClaimsPrincipalSerializer`. Is that correct?

Source/Csla/Serialization/Mobile/CustomSerializers/PocoSerializer.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System.Text.Json;
+
+namespace Csla.Serialization.Mobile.CustomSerializers;
+
+/// <summary>
+/// A formatter that can serialize and deserialize a POCO
+/// by using System.Text.Json.
+/// </summary>
+public class PocoSerializer<T> : IMobileSerializer
```
StefanOssendorf, line 17: As far as I know STJ isn't that good with circular references. But that knowledge is a bit outdated so get that fixed?

Source/Csla/Serialization/Mobile/ITypeMap.cs
```diff
@@ -0,0 +1,29 @@
...
+public interface ITypeMap
+{
+  /// <summary>
+  /// Gets or sets the original type.
+  /// </summary>
+  Type OriginalType { get; }
+  /// <summary>
+  /// Gets or sets the serializer type.
+  /// </summary>
+  Type SerializerType { get; }
```
StefanOssendorf, line 23: Only gets no set here :)

Source/Csla/Serialization/Mobile/CustomSerializers/ClaimsPrincipalSerializer.cs
```diff
@@ -0,0 +1,120 @@
...
+
+using System.Security.Claims;
+
+namespace Csla.Serialization.Mobile.CustomSerializers;
+
+#if NET8_0_OR_GREATER || NETSTANDARD2_0_OR_GREATER
+/// <summary>
+/// A formatter that can serialize and deserialize a ClaimsPrincipal object.
+/// </summary>
+public class ClaimsPrincipalSerializer : IMobileSerializer
```
StefanOssendorf, line 17: Any good reason to make them `public` instead of `internal`?
`internal` would save us from beeing unable to change them because that would be a breaking change. Furthermore if someone _really_ wants that serializer they have two options.
Ask us and we could provide a new method which retrieves the serializer for a `ClaimsPrincipal` or (my preferred variant) the user can just retrieve _all_ known serializers and use the one who says "I can process `ClaimsPrincipal`".

Source/Csla/Serialization/Mobile/TypeMap.cs
```diff
@@ -0,0 +1,38 @@
...
+/// <inheritdoc />
+public class TypeMap<T, S> : ITypeMap
+  where S : IMobileSerializer
+{
+  /// <summary>
+  /// Creates an instance of the class.
+  /// </summary>
+  public TypeMap()
+  {
+    CanSerialize = (t) => t == OriginalType;
```
StefanOssendorf, line 20: This ctor should chain to the other one.

Source/Csla/Serialization/Mobile/TypeMap.cs
```diff
@@ -0,0 +1,38 @@
...
+  {
+    CanSerialize = canSerialize;
+  }
+
+  /// <inheritdoc />
+  public Type OriginalType => typeof(T);
+  /// <inheritdoc />
+  public Type SerializerType => typeof(S);
+  /// <inheritdoc />
+  public Func<Type, bool> CanSerialize { get; set; }
```
StefanOssendorf, line 37: This should be `get` only or at max `init` but not `set`.

Source/Csla/Serialization/SerializationFormatterFactory.cs
```diff
@@ -19,7 +19,9 @@ public static class SerializationFormatterFactory
     /// </summary>
     public static ISerializationFormatter GetFormatter(ApplicationContext applicationContext)
     {
-      return (ISerializationFormatter)applicationContext.CreateInstanceDI(ApplicationContext.SerializationFormatter);
+      var options = applicationContext.GetRequiredService<Configuration.CslaOptions>();
```
StefanOssendorf, line 22: This whole method/type can be removed if we move to a pure DI "based" config.

StefanOssendorf (changes requested):

Source/Csla/Serialization/Mobile/ITypeMap.cs
```diff
@@ -0,0 +1,29 @@
...
+
+/// <summary>
+/// Maps a type to a serializer type.
+/// </summary>
+public interface ITypeMap
+{
+  /// <summary>
+  /// Gets or sets the original type.
+  /// </summary>
+  Type OriginalType { get; }
```
StefanOssendorf, line 19: Do we really need this property? Isn't it useless?
We only need `CanSerialize` and `SerializerType` or am I wrong?
Even our `TypeMap<T,S>` implementation does not need `OriginalType`. The default ctor could just use ´typeof(T)` instead of the property.

Source/Csla/Serialization/Mobile/CustomSerializers/ClaimsPrincipalSerializer.cs
rockfordlhotka: Good point, I'll make this internal

Source/Csla/Serialization/Mobile/TypeMap.cs
rockfordlhotka: I made it `internal set`

Source/Csla.test/Serialization/CustomSerializerTests.cs
```diff
@@ -0,0 +1,122 @@
...
+      var provider = services.BuildServiceProvider();
+      var applicationContext = provider.GetRequiredService<ApplicationContext>();
+
+      var nonSerializable = new NonSerializableType { Name = "test" };
+      var cloner = new Core.ObjectCloner(applicationContext);
+      var clone = (NonSerializableType)cloner.Clone(nonSerializable);
+    }
+
+    [TestMethod]
+    [ExpectedException(typeof(NotSupportedException))]
```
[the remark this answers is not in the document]
rockfordlhotka: Switched to `MobileFormatterException`

Source/Csla/Serialization/SerializationFormatterFactory.cs
rockfordlhotka: Removed formatter factory

Source/Csla/Serialization/Mobile/ITypeMap.cs
rockfordlhotka: Removed property

Source/Csla/Serialization/Mobile/TypeMap.cs
rockfordlhotka: yes

Source/Csla/Serialization/Mobile/MobileFormatter.cs
```diff
@@ -117,6 +111,14 @@ private void ConvertEnumsToIntegers()
       }
     }
 
+    private MobileFormatterOptions GetOptions()
+    {
+      var cslaOptions = applicationContext.GetRequiredService<CslaOptions>();
+      if (cslaOptions.SerializationOptions.FormatterOptions is not MobileFormatterOptions options)
+        throw new NotSupportedException("MobileFormatterOptions == null");
```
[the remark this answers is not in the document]
rockfordlhotka: obsolete

Source/Csla/Serialization/Mobile/IMobileSerializer.cs
```diff
@@ -0,0 +1,39 @@
...
+  /// </summary>
+  public interface IMobileSerializer
+  {
+#if NET8_0_OR_GREATER
+    /// <summary>
+    /// Gets a value indicating whether this serializer can
+    /// serialize the specified type.
+    /// </summary>
+    /// <param name="type">Type to check</param>
+    static bool CanSerialize(Type type) => false;
```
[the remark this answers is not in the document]
rockfordlhotka: Yes, because I don't want to have to create an instance of the serializer if we're not going to use it. This allows detection of whether the serializer can handle the type _before_ we create the instance.

Source/Csla/Serialization/Mobile/Configuration/MobileFormatterOptions.cs
rockfordlhotka: It is gone

Source/Csla/Serialization/Mobile/CustomSerializers/PocoSerializer.cs
rockfordlhotka: Yes, but the key name is per-instance of a serialized thing. There's no need for keys to be unique across different instances of serialized objects.

Source/Csla/Serialization/Mobile/CustomSerializers/PocoSerializer.cs
rockfordlhotka: It is not. If people want full fidelity serialization they'll need to use MobileFormatter, Newtonsoft, or (maybe) protobuf.

This will be a limitation of the default POCO serialization support, as I sure don't want to write and maintain what amounts to a whole _other_ MobileFormatter that can serialize POCO graphs with full fidelity.

Source/Csla/Configuration/Fluent/CslaOptions.cs
```diff
@@ -80,7 +81,7 @@ public CslaOptions UseReflectionFallback(bool value)
     /// <summary>
     /// Gets the SerializationOptions instance.
     /// </summary>
-    internal SerializationOptions SerializationOptions { get; } = new SerializationOptions();
+    public SerializationOptions SerializationOptions { get; private set; }
```
StefanOssendorf, line 84: Any reason for the `set`? Should be removed.

Source/Csla/Configuration/Fluent/SerializationOptions.cs
```diff
@@ -7,44 +7,41 @@
...
     /// <summary>
     /// Sets the serialization formatter type used by CSLA .NET
     /// for all explicit object serialization (such as cloning,
     /// n-level undo, etc). Default is MobileFormatter.
     /// </summary>
-    public SerializationOptions SerializationFormatter<T>() where T: ISerializationFormatter
+    public SerializationOptions UseSerializationFormatter<T>() where T : ISerializationFormatter
     {
-      ApplicationContext.SerializationFormatter = typeof(T);
+      CslaOptions.Services.TryAddTransient(typeof(ISerializationFormatter), typeof(T));
```
StefanOssendorf, line 32: This should just add the formatter specified. `CslaOptions.Services.AddTransient<ISerializationFormatter, T>()`. It's up to the user to use the configuration correctly. It would be very awkward why my call did not register the type.

Source/Csla/Configuration/Fluent/SerializationOptions.cs
```diff
@@ -7,44 +7,41 @@
...
-    /// Sets type of the writer that is used to write data to 
-    /// serialization stream in SerializationFormatterFactory.GetFormatter().
+    /// Gets the serialization formatter type used by CSLA .NET.
     /// </summary>
-    public SerializationOptions MobileWriter<T>() where T : ICslaWriter
-    {
-      CslaReaderWriterFactory.WriterType = typeof(T);
-      return this;
-    }
+    public Type SerializationFormatterType
```
StefanOssendorf, line 39: For me this is still awkward. I would just add `MobileFormatter` as a default into the DI with `AddTransient<ISerializationFormatter, MobileFormatter>`  and the user config adds his implementation after that one. Default behvior of MS DI (and all others as far as I know) the last registered implementation is the default one. If you want _all_ you have to request an `IEnumerable` of the interface.

Source/Csla/Serialization/Mobile/TypeMap.cs
```diff
@@ -0,0 +1,30 @@
...
+  /// Creates an instance of the class.
+  /// </summary>
+  public TypeMap()
+    : this(type => type == typeof(T))
+  { }
+
+  /// <inheritdoc />
+  public Type SerializerType => typeof(S);
+  /// <inheritdoc />
+  public Func<Type, bool> CanSerialize { get; private set; } = canSerialize;
```
StefanOssendorf, line 29: The `private set` necessary here? Otherwise remove it.

StefanOssendorf (changes requested):

PR #4039, opened 2024-06-16
Features/incremental source generator

### LINKED ISSUE #3068: Upgrade source generators to the more efficient incremental generator style
The SerializationPartialsGenerator type was created to automatically implement the IMobileObject interface on POCO types and does so by implementing the ISourceGenerator interface. In .NET 6, Microsoft added a newer generator API that offers better performance on larger code bases - incremental source generators. Incremental source generators can reduce load on the build infrastructure during build and development, reducing the risk that the generator will slow developer activities.

Upgrading from implementing ISourceGenerator to instead implement IIncrementalGenerator could be beneficial for users of CSLA who develop extremely large codebases.

This would affect the SerializationPartialsGenerator class and associated types.

The difference for a single generator is probably quite small, so this is a low priority task. This task would become more relevant if we added more source generators.

luizfbicalho:
Closes #3068 

@rockfordlhotka  I added nullable reference types on the generator, please check if the changes to enable nullable that I added are ok for you


rockfordlhotka:
@luizfbicalho I've asked @TheCakeMonster for input here. I am not sure this code is current or valid anymore. It was never released into production.

Source/Csla.Generators/cs/Csla.Generators.CSharp/AutoSerializable/IncrementalSerializationPartialsGenerator.cs
```diff
@@ -0,0 +1,50 @@
...
+                if (ctx.TargetNode is TypeDeclarationSyntax typeDeclarationSyntax)
+                {
+                  return TypeDefinitionExtractor.ExtractTypeDefinition(new DefinitionExtractionContext(ctx.SemanticModel), typeDeclarationSyntax);
+                }
+                return null;
+              })
+          .Where(static m => m is not null);
+
+      // Combine the collected syntax nodes with the semantic model
+      var compilationAndClasses = context.CompilationProvider.Combine(classDeclarations.Collect());
```
StefanOssendorf, line 31: This is highly inefficient because _every_ keypress forces the source generator to rerun even though it's probably not affected.
I can recommend this and the whole series around that post: https://andrewlock.net/creating-a-source-generator-part-9-avoiding-performance-pitfalls-in-incremental-generators/#5-be-careful-using-compilationprovider

luizfbicalho: Thanks 
I changed it to only select the nullable from the compilation.

While @TheCakeMonster  validate the generated source I'll take a look in the others pitfalls and see if there are more

TheCakeMonster:
@rockfordlhotka You are probably right that it isn't entirely current, but only in the sense that I have not checked it recently. I think it's a useful facility, so it's worth getting it released, unless there is a more modern replacement? I thought I saw an email of a merged PR for being able to serialize POCO types go past in the whizz of emails there have been in the past few days, but I can't keep up with them all at the moment.

The idea of switching to incremental source generation is a good one. It's quite a lot more efficient, I think, so it improves the build and development experience. @JasonBock is, of course, much more knowledgeable in this area than me; I simply did the grunt work, including changing it after his review.

There were a couple of reasons why it didn't get released:

1. I couldn't remember how to change the nuget build to get it included. This is not something I know anything about, as I've never been in a position to need nuspec. This is the main reason.
2. I started to wonder if this suffered the same security problem as BinaryFormatter, in that it might be capable of deserializing unintended types. I don't _think_ it suffers quite the same problem, but that's very different from being able to say that I know that it doesn't.

Source/Csla.Generators/cs/Csla.Generators.CSharp/ExceptionExtensions.cs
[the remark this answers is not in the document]
luizfbicalho: I changed this because of that error

RS1035	The symbol has been marked as banned for use in analyzers, and an alternate should be used instead.	

Now I got the two replaces

luizfbicalho:
Is there anything else I can do about this PR?

rockfordlhotka:
Not right now. I'll review it when I get a chance.

rockfordlhotka (changes requested):
@luizfbicalho 

I don't see any issues with the code in the PR.

I am left wondering how it is actually used? I (like probably most CSLA devs) aren't familiar with this technology, and I'm not sure how I'd use it in my code?

For example, can I use this all by itself and MobileFormatter will "just work"? Or should it integrate into MobileFormatter using the new ability to add custom serializers? Should I use this instead of the built-in PocoSerializer that is reflection-based? If so, how?

I guess what I'm looking for is a readme.md in the `Source/`Csla.Generators` directory that tells someone how to get started.

luizfbicalho:
> I guess what I'm looking for is a readme.md in the `Source/`Csla.Generators` directory that tells someone how to get started.

I just added it,let me know what do you think of it

Source/Csla.Generators/readme.md
```diff
@@ -0,0 +1,97 @@
...
+    public string Postcode { get; set; } = string.Empty;
+
+  }
+}
+```
+
+## Understanding [AutoSerializable]
+
+- **Source Generators:** Introduced in .NET 5, source generators allow for compile-time code generation based on the existing codebase. They are a powerful tool for reducing boilerplate code and automating repetitive coding tasks.
+- **[AutoSerializable] Attribute:** This custom attribute is used to mark a class for automatic serialization code generation. When the source generator sees this attribute, it generates the necessary serialization and deserialization methods for the class, tailored to its properties.
```
rockfordlhotka, line 42: I think the `AutoSerializable` class is in the `Csla` project, but really that class should be in the generators project. Same namespace, but the file/type should only be available if the user has chosen to use this particular serializer by referencing that NuGet package.

rockfordlhotka (changes requested):
I like the readme, thank you!

Just the change about the location of the `AutoSerializable` type and building the NuGet package and we should be good.

Source/Csla.Generators/readme.md
```diff
@@ -0,0 +1,97 @@
+# Using the Source Generator with AddressPOCO in CSLA.NET
+
+This guide provides a step-by-step approach to using the source generator in the CSLA.NET project, specifically with the `AddressPOCO.cs` class as an example. The source generator is a powerful tool that can automatically generate additional source code for your projects, enhancing productivity and maintainability.
+
+## Prerequisites
+
+- Visual Studio 2019 or later with the .NET Core development workload installed.
+- The CSLA.NET NuGet package added to your project.
+- Ths CSLA.NET Source Generator NuGet package added to your project.
```
rockfordlhotka, line 9: Is this being built now?

There's a PR I hope to close today that changes the nuget package creation to be based on the csproj files, so this will need to work within that model.

rockfordlhotka (changes requested):

Source/Csla.Generators/readme.md
luizfbicalho: I created a base PR of the nuget packages, After this merge I can add to the project do generate this project as nuget package also

Source/Csla.Generators/readme.md
luizfbicalho: I didn't understand what I need to change, Do you want the autoSerializable to be moved to the Generator Project?

rockfordlhotka: Yes, exactly. It is in the csla project now, but should be in the generator project.

luizfbicalho: Moved to the generator. there are other 2 unused attributes, we could talk about them later

rockfordlhotka (approved):

PR #4043, opened 2024-06-17
#4015 Fix recursion issue and add unit tests

Bowman74:
#4105 Changes so fix infinite recursion issues and add unit tests


Source/Csla/Core/ContextDictionary.cs
```diff
@@ -76,25 +76,34 @@ void IMobileObject.SetChildren(SerializationInfo info, MobileFormatter formatter
...
+        throw new ArgumentException("An item with the same key has already been added.");
+      }
     }
 
     /// <inheritdoc cref="System.Collections.IDictionary.Remove(object)"/>
     public void Remove(object key)
     {
-      ((IDictionary)this).Remove(key);
+      object dummy;
+      var removed = TryRemove(key, out dummy);
```
StefanOssendorf, line 102: `out var _` is a bit more effcient because the compiler can discard the variable completly.

StefanOssendorf (changes requested):

Source/Csla/Core/ContextDictionary.cs
Bowman74: Changed

rockfordlhotka: @Bowman74 don't mark conversations as resolved - the reviewer should do that

Source/Csla/Core/ContextDictionary.cs
```diff
@@ -76,25 +76,34 @@ void IMobileObject.SetChildren(SerializationInfo info, MobileFormatter formatter
...
     }
 
     /// <inheritdoc cref="System.Collections.IDictionary.Add(object, object?)"/>
     public void Add(object key, object value)
     {
-      ((IDictionary)this).Add(key, value);
+      bool added = TryAdd(key, value);
+      if (!added)
+      {
+        throw new ArgumentException("An item with the same key has already been added.");
```
rockfordlhotka, line 94: 1. Is this the right exception to throw? It seems maybe wrong?
2. English text can't be hard-coded. Either the text needs to come from the Resources file or avoid using text at all.

Source/Csla/Core/ContextDictionary.cs
```diff
@@ -76,25 +76,34 @@ void IMobileObject.SetChildren(SerializationInfo info, MobileFormatter formatter
...
 
     /// <inheritdoc cref="System.Collections.IDictionary.Remove(object)"/>
     public void Remove(object key)
     {
-      ((IDictionary)this).Remove(key);
+      object _;
+      var removed = TryRemove(key, out _);
+      if (!removed)
+      {
+        throw new NotSupportedException("The key does not exist in the dictionary.");
```
rockfordlhotka, line 105: English text can't be hard-coded. Either the text needs to come from the Resources file or avoid using text at all.

Source/Csla/Core/ContextDictionary.cs
```diff
@@ -104,19 +113,14 @@ public void Remove(object key)
     /// <inheritdoc cref="System.Collections.ICollection.SyncRoot"/>
     public object SyncRoot
     {
-      get => ((IDictionary)this).SyncRoot;
+      get => throw new NotSupportedException("SyncRoot NotSupported");
```
rockfordlhotka, line 116: English text can't be hard-coded. Either the text needs to come from the Resources file or avoid using text at all.

In this case just `throw new NotSupportedException(nameof(SyncRoot));`

rockfordlhotka (changes requested):

Source/Csla/Core/ContextDictionary.cs
Bowman74: That is the exception the `IDictionary` Add method raises for an already existing key if you peek at the documentation.

rockfordlhotka: OK, then it must be the right exception type 😄 

We still need the text to be in the resx file and come from the Resources class.

Bowman74: Changed to use resource

Source/Csla/Core/ContextDictionary.cs
Bowman74: Changed to use resource

Source/Csla/Core/ContextDictionary.cs
Bowman74: Changed to use resource

Source/Csla/Core/ContextDictionary.cs
```diff
@@ -98,8 +98,8 @@ public void Add(object key, object value)
     /// <inheritdoc cref="System.Collections.IDictionary.Remove(object)"/>
     public void Remove(object key)
     {
-      object dummy;
-      var removed = TryRemove(key, out dummy);
+      object _;
```
rockfordlhotka, line 101: I'm not sure you actually need to declare this?

Or maybe you do and it can be

`var removed = TryRemove(key, out var _);`

?

rockfordlhotka (changes requested):

Source/Csla/Core/ContextDictionary.cs
Bowman74: Unit tests indicated that syntax is fine as we never use the out value. Changed.

rockfordlhotka (approved):

Issue #3207, opened 2022-11-14
Use CentralPackageVersioning in samples to simplify maintenance


TheCakeMonster:
Central package versioning is now a first class citizen of .NET, and make management of things like sample applications easier. It probably makes sense to change over the samples to make use of this scheme so that they are easier to keep upgraded to the newer versions of CSLA.

https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management

I realised this whilst working on the compilation error we ended up with in the Blazor sample apps. I ended up having to change version numbers in multiple places, which is no longer necessary.

**Suggested Solution**
Upgrade each sample application separately, so that we can upgrade them one at a time. In the longer term we might decide that refactoring the samples folder makes sense to enable even simpler upgrades. However, being too aggressive may make upgrades larger, and hinder progress rather than enable it.

**Further Improvement**
The two Blazor samples BlazorExample and BlazorServerExample share backend packages, but are separate apps. I would suggest that moving both into a Blazor subfolder, so that the Directory.Packages.props file can apply to both, would be even better; this ensures that they are more likely to both compile together. See #3204 for an example of the current problem - upgrading just one of the two sample apps breaks the other.

sshushliapin:
@TheCakeMonster, @rockfordlhotka, do you have any opinion if it makes any sense to introduce CPM in _single-project_ sample solutions such as [DataPortalCacheExample](https://github.com/MarimerLLC/csla/tree/main/Samples/DataPortalCacheExample)? I ignored them for now, since for me it looks redundant... At the same time, if consistency is more important, I can come up with a single PR that introduces CPM for all the simple .net8 based solutions (for the .net8 projects there is much less work needed so a single PR should be ok).

rockfordlhotka:
I don't see any value in that case, just added complexity.

PR #4044, opened 2024-06-18
#3207 Use Central Package Management (CPM) in CustomErrorHandling solution

sshushliapin:
Use Central Package Management (CPM) in CustomErrorHandling solution (#3207).


rockfordlhotka (approved):

rockfordlhotka:
@sshushliapin I _love_ how much cleaner these solutions are with your changes!

PR #4049, opened 2024-06-19
Fix selection of custom ContextManager through CslaOptions

### LINKED ISSUE #4047: Cannot change ContextManager through CslaOptions.ContextManagerType
**Describe the bug**
I'm not able to change ContextManager using CslaOptions.ContextManagerType in AddCsla() extension method, but I have to explicitly add it to the ServiceCollection of AppBuilder before AddCsla() method is called.

This is because of this order of execution in AddCsla method. At first it registers default ContextManagers into a ServiceCollection and after that it registers the one given in CslaOptions.ContextManagerType:
https://github.com/MarimerLLC/csla/blob/d6069d465041fdc687bc7a46602540754d56f8c8/Source/Csla/Configuration/ConfigurationExtensions.cs#L48-L50
Consequently ApplicationContextAccessor gets both of them, but selects the first valid ContextManager, which is the default one and not the one that was specified in CslaOptions.ContextManagerType.
https://github.com/MarimerLLC/csla/blob/d6069d465041fdc687bc7a46602540754d56f8c8/Source/Csla/Core/ApplicationContextAccessor.cs#L34-L41

**Version and Platform**
CSLA version: 8.2.3
OS: Windows
Platform: .NET 8.0

**Code that Fails**
```csharp
var builder = Host.CreateApplicationBuilder();

builder.Services.AddCsla((c) => c.ContextManagerType = typeof(ApplicationContextManagerTls));
var host = builder.Build();

var appContextAccessor = serviceProvider.GetRequiredService<ApplicationContextAccessor>();
var contextManager = appContextAccessor.GetContextManager();
// this is false, should be true
var isContextManagerTls = contextManager.GetType() == typeof(ApplicationContextManagerTls);
// this is true, should be false
var isContextManagerDefault = contextManager.GetType() == typeof(ApplicationContextManager);
```

If I register the ContextManager this way, it will work:
```csharp
builder.Services.AddScoped<Csla.Core.IContextManager, ApplicationContextManagerTls>();
builder.Services.AddCsla();
```

**Additional context**

rockfordlhotka:
Fixes #4047

This pull request addresses the issue of `ApplicationContextAccessor` not selecting the `ContextManager` specified in `CslaOptions.ContextManagerType` over the default one. The changes ensure that custom `ContextManager` types are prioritized and registered before the default types, allowing for the correct `ContextManager` to be used according to the configuration.

- Modifies the `RegisterContextManager` method in `ConfigurationExtensions.cs` to use `TryAddScoped` instead of `AddScoped` for registering default `ContextManager` types. This change prevents overwriting custom registrations if a custom `ContextManagerType` is provided in `CslaOptions`.
- Ensures that if a custom `ContextManagerType` is specified, it is registered before attempting to register any default `ContextManager` types, allowing the custom type to take precedence.
- Applies `TryAddScoped` in the `LoadContextManager` method for loading specific `ContextManager` types based on the environment, further ensuring that custom types are not overwritten by default or environment-specific types.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/MarimerLLC/csla/issues/4047?shareId=2674ada7-7ecc-4f5a-9ea6-4afd9258d088).

### DIFF
--- a/Source/Csla.test/AppContext/AppContextTests.cs
+++ b/Source/Csla.test/AppContext/AppContextTests.cs
@@ -10,6 +10,7 @@ using Csla.Configuration;
 using Csla.Core;
 using Csla.TestHelpers;
 using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Csla.Test.AppContext
@@ -31,7 +32,16 @@ namespace Csla.Test.AppContext
       TestResults.Reinitialise();
     }
 
-    #region Simple Test
+    [TestMethod]
+    public void UseCustomApplicationContext()
+    {
+      var services = new ServiceCollection();
+      services.AddCsla(o => o.UseContextManager<ApplicationContextManagerTls>());
+      var serviceProvider = services.BuildServiceProvider();
+
+      var applicationContext = serviceProvider.GetRequiredService<ApplicationContext>();
+      Assert.IsInstanceOfType(applicationContext.ContextManager, typeof(ApplicationContextManagerTls));
+    }
 
     [TestMethod]
     public void SimpleTest()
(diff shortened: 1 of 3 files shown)


rockfordlhotka:
This isn't complete. It is my first attempt at using the new GH Copilot workspace preview feature. Not entirely sure the AI did the right thing, and in any case, it didn't generate (or allow me to add) any unit tests.

PR #4051, opened 2024-06-19
Add analyzer which check how ObjectAuthorizationRules applied 

### LINKED ISSUE #1560: Create analyzer to ensure ObjectAuthorizationRules method is public
The Blazor wasm linker totally removes `private static` methods that aren't directly invoked in your code.

In the past, the per-type rules have been added using a `private static` method, and that method is now entirely removed before the code executes on the client, meaning that this method must now be `public static`.

For example:

```c#
        [ObjectAuthorizationRules]
        public static void AddPerTypeRules()
```

The alternative is that everyone creating a business library would need to create a linker xml file and explicitly list each of these methods to prevent the linker from removing them. That seems pretty impractical.

It might also be good for the analyzer to add

```c#
        [EditorBrowsable(EditorBrowsableState.Never)]
```

to the method. Though in my case I have intellisense configured to always show all methods, so this has no effect, but maybe for some people it would be useful?

### LINKED ISSUE #1561: Create analyzer for AddObjectAuthorizationRules methods
In the past people would use the "well-known" `AddObjectAuthorizationRules` name for the `static` method that established per-type rules.

Now people should be using the `ObjectAuthorizationRules` attribute to identify the `static` method that establishes per-type rules. They can name the method as they choose.

It would be ideal to have an analyzer that identifies the old-style approach, and offers a fix to add the new attribute.

kant2002:
make sure that methods is `public`, `static` and has `[ObjectAuthorizationRules]` attribute attached.

Fixes #1561
Fixes #1560


Source/Csla.Analyzers/Csla.Analyzers.Tests/Extensions/DataPortalOperationQualificationTests.cs
```diff
@@ -9,76 +9,76 @@ public sealed class DataPortalOperationQualificationTests
     [TestMethod]
     public void Create()
     {
-      var qualification = new DataPortalOperationQualification();
+      var qualification = new CslaOperationQualification();
```
kant2002, line 12: Forget to mention this rename. I decide do not create new class for this, but rather change name of the existing one. If that's not desired. let me know, I will create new class of the same shape.

rockfordlhotka: Hmm, I wonder if we shouldn't duplicate the type actually.

Data portal operations are a thing unto themselves - totally separate from rules.

I am worried that mixing the two, even in tests, might lead to confusion in the future as we maintain the code over time.

rockfordlhotka (changes requested):

Source/Csla.Analyzers/Csla.Analyzers.Tests/Extensions/DataPortalOperationQualificationTests.cs
kant2002: You are right. I fix it.

rockfordlhotka (approved):

Issue #1233, opened 2019-08-12
Use nullable reference types throughout CSLA codebase


JasonBock:
**Is your feature request related to a problem? Please describe.**
C#8 adds nullable reference types. Not only will CSLA need to support this, but it should also annotate its members with not just `?`, but with the new nullable attributes.

**Describe the solution you'd like**
NRT support throughout CSLA.

**Describe alternatives you've considered**
Nothing :)

**Additional context**
Use [this reference](https://devblogs.microsoft.com/dotnet/try-out-nullable-reference-types/) for details/information.

StefanOssendorf:
@rockfordlhotka
I'd love to tackle this issue. I don't like `null` :-)
Btw the nullable annotation context is not a breaking change. The annotation is only a compiler feature during development and not something the runtime checks.

rockfordlhotka:
I'd love to have you do this!

It'll be CSLA 9 thing, as I'm sure there'll be breaking changes.

StefanOssendorf:
@rockfordlhotka I'll link you here from now on for questions regarding the nullable context :-)
First question:
https://github.com/MarimerLLC/csla/blob/579c84dd14398549c1b48be3892d543a1a956260/Source/Csla/Core/FieldManager/FieldData.cs#L28-L42

Is the empty ctor only there for the `MobileFormatter`? It's at least not used within CSLA itself. I'd love to mark it `Obsolete(true)` because it would remove a lot of "maybe nullable" things in the class itself. For example the name can than be guaranteed.

rockfordlhotka:
MobileFormatter does require an empty default ctor. Whether that is explicit or implicit, it is required for all types that implement `IMobileObject`.

StefanOssendorf:
In that case I'll mark it as Obsolete. MobileFormatter can still use it but it's not usable any more at compile time :-)

StefanOssendorf:
And should I _really_ add now everywhere a null argument check where it's currently just assumed to be not null? 😅🤔

rockfordlhotka:
That seems wise doesn't it?

JasonBock:
I concur with @rockfordlhotka. If a parameter is a NRT, someone could still pass in `null`, so an `ArgumentNullException.ThrowIfNull()` call is warranted.

rockfordlhotka:
Hey - this is now part of the permanent record!! @JasonBock agrees with me!! 😁

JasonBock:
[...]

PR #4064, opened 2024-06-25
Add nullable annotation to Mvc5 and AspNetCore projects

StefanOssendorf:
I think this is the first PR which will actually hit real world scenarios. So whoever reviews this please make this thoroughly!

IMHO we should try to get rid of as most as possible maybe nulls like `ClientContext`, `LocalContext` and `ApplicationContext` but that's a discussion we should hold in #1233


StefanOssendorf:
I'll modify this post unitl the PR is approved and merged with mentionable changes:
* `ISessionIdManager` now disallows `null` keys because our underlying implementation can't even handle a `null` as a key.

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
```diff
@@ -113,44 +118,52 @@ public System.Security.Principal.IPrincipal GetUser()
...
     /// <summary>
     /// Sets the client context.
     /// </summary>
     /// <param name="clientContext">Client context.</param>
     /// <param name="executionLocation"></param>
-    public void SetClientContext(IContextDictionary clientContext, ApplicationContext.ExecutionLocations executionLocation)
+    /// <exception cref="InvalidOperationException">The underlying <see cref="HttpContext"/> is <see langword="null"/>.</exception>
+    public void SetClientContext(IContextDictionary? clientContext, ApplicationContext.ExecutionLocations executionLocation)
     {
+      ThrowIfHttpContextIsNull();
```
rockfordlhotka, line 166: It should be valid to reset the value to null.

This class implements what amounts to a private API for use only by `ApplicationContext`, which provides the public API to the end developer.

From the context of a _private_ API, null should be a valid value, as it enjoys an optimization in serialization within the data portal.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -51,7 +55,7 @@ public ApplicationContextManagerBlazor(IHttpContextAccessor hca, AuthenticationS
     /// <summary>
     /// Gets or sets a reference to the current ApplicationContext.
     /// </summary>
-    public ApplicationContext ApplicationContext { get; set; }
+    public ApplicationContext? ApplicationContext { get; set; }
```
rockfordlhotka, line 58: Shouldn't the setter throw if a null is provided like in the other context manager?

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -160,8 +166,10 @@ public IContextDictionary GetLocalContext()
     /// Sets the local context.
     /// </summary>
     /// <param name="localContext">Local context.</param>
+    /// <exception cref="InvalidOperationException"><see cref="ApplicationContext"/> is <see langword="null"/>.</exception>
     public void SetLocalContext(IContextDictionary localContext)
     {
+      ThrowIfApplicationContextIsNull();
```
rockfordlhotka, line 172: As before, null should be a valid value here.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -194,8 +204,10 @@ public IContextDictionary GetClientContext(ApplicationContext.ExecutionLocations
     /// </summary>
     /// <param name="clientContext">Client context.</param>
     /// <param name="executionLocation"></param>
+    /// <exception cref="InvalidOperationException"><see cref="ApplicationContext"/> is <see langword="null"/>.</exception>
     public void SetClientContext(IContextDictionary clientContext, ApplicationContext.ExecutionLocations executionLocation)
     {
+      ThrowIfApplicationContextIsNull();
```
rockfordlhotka, line 210: null is valid here

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
```diff
@@ -168,18 +170,16 @@ private void SetHostPrincipal(Task<AuthenticationState> task)
     /// <summary>
     /// Gets the local context.
     /// </summary>
-    public IContextDictionary GetLocalContext()
+    public IContextDictionary? GetLocalContext()
```
rockfordlhotka, line 173: The getter should create an instance on demand so the end developer never gets a null, even if the private value was null behind the scenes.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
```diff
@@ -188,10 +188,8 @@ public void SetLocalContext(IContextDictionary localContext)
     /// Gets the client context.
     /// </summary>
     /// <param name="executionLocation"></param>
-    public IContextDictionary GetClientContext(ApplicationContext.ExecutionLocations executionLocation)
+    public IContextDictionary? GetClientContext(ApplicationContext.ExecutionLocations executionLocation)
     {
```
rockfordlhotka, line 192: The getter should create an instance on demand so the end developer never gets a null, even if the private value was null behind the scenes.

Source/Csla.AspNetCore/ConfigurationExtensions.cs
```diff
@@ -35,8 +35,11 @@ public static CslaOptions AddAspNetCore(this CslaOptions config)
     /// </summary>
     /// <param name="config">CslaOptions instance</param>
     /// <param name="options">Options object</param>
-    public static CslaOptions AddAspNetCore(this CslaOptions config, Action<AspNetCoreConfigurationOptions> options)
+    /// <exception cref="ArgumentNullException"><paramref name="config"/> is <see langword="null"/>.</exception>
+    public static CslaOptions AddAspNetCore(this CslaOptions config, Action<AspNetCoreConfigurationOptions>? options)
     {
+      ArgumentNullException.ThrowIfNull(config);
```
rockfordlhotka, line 41: Shouldn't this throw on `options` instead of `config`?

Source/Csla.AspNetCore/RazorPages/PageModel.cs
```diff
@@ -45,9 +46,10 @@ public async Task<bool> SaveAsync(bool forceUpdate = false)
       try
       {
         if (Item is BusinessBase bb && !bb.IsValid)
-          AddBrokenRuleInfo(Item, null);
+          AddBrokenRuleInfo(Item, string.Empty);
```
rockfordlhotka, line 49: I wonder if `string.Empty` and `null` render differently in html? I've never tested that but I know when a null is returned that the bound html element doesn't appear (or take up space in the UI). 

We should confirm that an empty string has the same behavior when rendered.

Source/Csla.Web.Mvc.Shared/Controller.cs
```diff
@@ -72,7 +72,7 @@ protected virtual async Task<bool> SaveObjectAsync<T>(T item, Action<T> updateMo
         updateModel?.Invoke(item);
         if (item is BusinessBase bb && !bb.IsValid)
         {
-          AddBrokenRuleInfo(item, null);
+          AddBrokenRuleInfo(item, string.Empty);
```
rockfordlhotka, line 75: Same concern as above as to how html renders null vs ""

Source/Csla/Core/IContextManager.cs
```diff
@@ -41,7 +42,7 @@ public interface IContextManager
     /// <summary>
     /// Gets the local context.
     /// </summary>
-    IContextDictionary GetLocalContext();
+    IContextDictionary? GetLocalContext();
```
rockfordlhotka, line 45: add `#nullable enable` to avoid build warnings for everyone else while this is all being worked on. We can remove that later when the conversion is done.

rockfordlhotka (changes requested):

Source/Csla.AspNetCore/ApplicationContextManagerHttpContext.cs
StefanOssendorf: The `clientContext` itself can be `null` here. But I changed from a `NullReferenceException` to an `InvalidOperationException` because we _can't_ be sure `HttpContext` is not null here.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
StefanOssendorf: We don't have any context manager which guards against null.
I'd love to make sure the `ApplicationContext` is always set. But I think that's not possible due to possible circular references with `ApplicationContextAccessor`?
Or we make the `set` internal and say "As long as you use the framework as it's intended we guarantee it's not null"? 🤔

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
StefanOssendorf: It's value. But `ApplicationContext` can be `null` and instead of an `NullReferenceException` we are now throwing an `InvalidOperationException`. Imho that's the more correct exception here.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
StefanOssendorf: Exception comes from `ApplicationContext` could be `null`.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
StefanOssendorf: Good catch. I first tried to built it without `null` but reverted that. Seems like not fully.

Source/Csla.AspNetCore/Blazor/ApplicationContextManagerInMemory.cs
StefanOssendorf: Same as with the LocalContext. Missed reverting that.

Source/Csla.AspNetCore/ConfigurationExtensions.cs
StefanOssendorf: No. Config is the instance we are working with and the `options` delegte is optional which is expressed with `null`. See line 44 and 30.

Source/Csla.AspNetCore/RazorPages/PageModel.cs
StefanOssendorf: Don't know about the html. And in this case it's irrelevant because the `AddBrokenRuleInfo` method won't use it. _But_ if it's used with `null` that will throw an `ArgumentNullException` because the `ModelState.AddModelError()` method does not accept `null` as an error message.
https://github.com/dotnet/aspnetcore/blob/ee4a0e9801c4e1e673d0373e1c51a6790de174cc/src/Mvc/Mvc.Abstractions/src/ModelBinding/ModelStateDictionary.cs#L324-L330

Source/Csla.Web.Mvc.Shared/Controller.cs
StefanOssendorf: Same answer as above with regard to `null` handling in the underlying framework methods.

Source/Csla/Core/IContextManager.cs
StefanOssendorf: Yep good idea.

rockfordlhotka (approved):

Source/Csla.AspNetCore/Csla.AspNetCore.csproj
luizfbicalho, line 1: I think that a better way to get nullable as a serious part is to add 
`  <WarningsAsErrors>Nullable</WarningsAsErrors>
`

StefanOssendorf: That's a great idea! I didn't know you can specify specific warnings to be treated as errors. I thought this is a true/false setting. I'll change that in my next PR.

PR #4068, opened 2024-06-27
Fix RabbitMq data portal channel

### LINKED ISSUE #3928: RabbitMqPortal needs to be revisited because it's not functional in the current state
The `RabbitMqPortal` implementation has two constructors.
https://github.com/MarimerLLC/csla/blob/a60f4da5d42538dcb24f5226a99f0f58c9bcbd61/Source/Csla.Channels.RabbitMq/RabbitMqPortal.cs#L33
https://github.com/MarimerLLC/csla/blob/a60f4da5d42538dcb24f5226a99f0f58c9bcbd61/Source/Csla.Channels.RabbitMq/RabbitMqPortal.cs#L58

As you can see we can either set the necessary services _or_ the data portal uri but _NOT_ both. Hence I dare to say this implementation doesn't work as inteded.
So we should revisit this implementation to make it right.

**Version and Platform**
CSLA version: Current

rockfordlhotka:
Closes #3928


Source/Csla.Channels.RabbitMq/RabbitMqChannelExtensions.cs
```diff
@@ -39,5 +41,25 @@ public static DataPortalClientOptions UseRabbitMqProxy(this DataPortalClientOpti
...
+    public static DataPortalServerOptions UseRabbitMqPortal(this DataPortalServerOptions config, Action<Channels.RabbitMq.RabbitMqPortalOptions>? options)
+    {
+      if (config is null)
+        throw new ArgumentNullException(nameof(config));
+
+      var portalOptions = new Channels.RabbitMq.RabbitMqPortalOptions();
+      options?.Invoke(portalOptions);
+
+      config.Services.AddScoped(_ => portalOptions);
+      config.Services.AddTransient<RabbitMqPortal>();
```
StefanOssendorf, line 61: We are sure this is only used in an asp.net core pipeline? Transient classes which implement `IDisposable` are a source of memory leak because the DI container tracks them until the holding "scope" gets disposed. In case of the root scope at the end of the application lifetime. Some information on that here https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection-guidelines#idisposable-guidance-for-transient-and-shared-instances

Source/Csla.Channels.RabbitMq/RabbitMqPortal.cs
```diff
@@ -78,21 +48,15 @@ private void InitializeRabbitMQ()
...
-        if (string.IsNullOrWhiteSpace(DataPortalQueueName))
-          throw new UriFormatException("DataPortalQueueName");
-
-        var factory = new ConnectionFactory { HostName = url.Host };
-        if (url.Port < 0)
-          factory.Port = url.Port;
-        var userInfo = url.UserInfo.Split(':');
+        Console.WriteLine($"Initializing connection to {DataPortalUri}");
+        if (DataPortalUri == null)
+          throw new NotSupportedException($"{nameof(DataPortalUri)} == null");
```
StefanOssendorf, line 53: That should be an `InvalidOperationException`. As far as I know `NotSupportedException` is only used in place where it's truly _not_ supported. Regardless what you are trying to do.

Source/Csla.Channels.RabbitMq/RabbitMqPortalOptions.cs
```diff
@@ -0,0 +1,45 @@
...
+        throw new UriFormatException("DataPortalQueueName");
+      _dataPortalUri = value;
+    }
+  }
+  
+  /// <summary>
+  /// Gets or sets the timeout for network
+  /// operations in seconds (default is 30 seconds).
+  /// </summary>
+  public int Timeout { get; set; } = 30;
```
StefanOssendorf, line 44: Is a `TimeSpan` here possible?

Source/Csla.Channels.RabbitMq/RabbitMqPortalOptions.cs
```diff
@@ -0,0 +1,45 @@
...
+/// <summary>
+/// Options for RabbitMqPortal
+/// </summary>
+public class RabbitMqPortalOptions
+{
+  private Uri? _dataPortalUri;
+  /// <summary>
+  /// Gets or sets the data portal server endpoint URL
+  /// </summary>
+  public Uri? DataPortalUri
```
StefanOssendorf, line 20: Document the exceptions thrown here. Probably the easierst way is to remove the current XML doc and let VS autocomplete it again with `///`. It should add the exceptions in the setter. Should save a bit of typing :)

Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
```diff
@@ -33,13 +33,14 @@ public RabbitMqProxy(ApplicationContext applicationContext, RabbitMqProxyOptions
...
       DataPortalUrl = options.DataPortalUrl;
+      Timeout = options.Timeout;
     }
 
     /// <summary>
     /// Gets or sets the timeout for network
     /// operations in seconds (default is 30 seconds).
     /// </summary>
-    public override int Timeout { get; set; } = 30;
+    public override int Timeout { get; set; }
```
StefanOssendorf, line 43: We can remove this override entirely now, can't we?

Source/Csla.Channels.RabbitMq/RabbitMqProxyOptions.cs
```diff
@@ -17,5 +17,11 @@ public class RabbitMqProxyOptions
     /// Data portal server endpoint URL
     /// </summary>
     public string DataPortalUrl { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Gets or sets the timeout for network
+    /// operations in seconds (default is 30 seconds).
+    /// </summary>
+    public int Timeout { get; set; } = 30;
```
StefanOssendorf, line 25: Is a `TimeSpan` here possible?

StefanOssendorf (changes requested):

Source/Csla.Channels.RabbitMq/RabbitMqChannelExtensions.cs
rockfordlhotka: This is actually intended for use in a console app, container, daemon, or Windows Service, not aspnetcore.

Source/Csla.Channels.RabbitMq/ProxyListener.cs
```diff
@@ -131,6 +131,7 @@ public void StartListening()
         Console.WriteLine($"Received reply for {ea.BasicProperties.CorrelationId}");
         if (Wip.WorkInProgress.TryRemove(ea.BasicProperties.CorrelationId, out WipItem? item))
         {
+          Console.WriteLine($"Processing reply for {ea.BasicProperties.CorrelationId}");
```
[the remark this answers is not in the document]
rockfordlhotka: No, thank you, fixed.

Source/Csla.Channels.RabbitMq/ProxyListener.cs
```diff
@@ -160,7 +161,7 @@ public void StartListening()
 
     public void Dispose()
     {
-      Connection?.Close();
+      //Connection?.Close();
```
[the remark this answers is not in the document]
rockfordlhotka: I thought that was gone, but it is now.

Source/Csla.Channels.RabbitMq/RabbitMqChannelExtensions.cs
StefanOssendorf: So the transient service won't be disposed until
1. The user uses explicit scopes to resolve the dependency
2. The application shuts down
Can't we promote the lifetime to scoped/singleton and mention that a user _should_ use explicit scopes to avoid the memory leak?
Otherwise we could add a factory (as suggested by my link) which returns it and the user is now responsible for correct disposal.

rockfordlhotka: I switched to a factory model

Source/Csla.Channels.RabbitMq/RabbitMqPortal.cs
```diff
@@ -21,54 +21,25 @@ namespace Csla.Channels.RabbitMq
   /// <summary>
   /// Exposes server-side DataPortal functionality through RabbitMQ
   /// </summary>
+  /// <remarks>
+  /// Creates an instance of the type
```
StefanOssendorf, line 25: This seems wrong? This is the type itself and does not create it.

Source/Csla.Channels.RabbitMq/RabbitMqPortal.cs
```diff
@@ -21,54 +21,25 @@ namespace Csla.Channels.RabbitMq
...
     private IModel? Channel;
     private string? DataPortalQueueName;
 
-    /// <summary>
-    /// Gets or sets the timeout for network
-    /// operations in seconds (default is 30 seconds).
-    /// </summary>
-    public int Timeout { get; set; } = 30;
-
     private Uri? DataPortalUri { get; set; }
```
StefanOssendorf, line 43: Should we still keep this nullable and not catch that during configuration/instantiation?
We should change from runtime error to app startup error to catch that earlier.

Source/Csla.Channels.RabbitMq/RabbitMqPortalFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+/// </summary>
+/// <param name="applicationContext"></param>
+/// <param name="dataPortal"></param>
+/// <param name="rabbitMqPortalOptions"></param>
+public class RabbitMqPortalFactory(ApplicationContext applicationContext, IDataPortalServer dataPortal, RabbitMqPortalOptions rabbitMqPortalOptions)
+{
+  /// <summary>
+  /// Gets an instance of RabbitMqPortal
+  /// </summary>
+  public RabbitMqPortal GetRabbitMqPortal()
```
StefanOssendorf, line 24: Shouldn't we name the method `CreateRabbitMqPortal`?
What about introducing an inteface `IRabbitMqPortalFactory` with the method here used. And this implementation could be made internal for easier changes in the future.

StefanOssendorf (changes requested):
Only the nullable comment "blocks" this PR. The other two are just remarks/opinions but not necessary (even though I think they would be worth it)

Source/Csla.Channels.RabbitMq/RabbitMqChannelExtensions.cs
```diff
@@ -58,7 +58,7 @@ public static DataPortalServerOptions UseRabbitMqPortal(this DataPortalServerOpt
       options?.Invoke(portalOptions);
 
       config.Services.AddScoped(_ => portalOptions);
-      config.Services.AddTransient<RabbitMqPortalFactory>();
+      config.Services.AddTransient<IRabbitMqPortalFactory, RabbitMqPortalFactory>();
```
StefanOssendorf, line 61: You now have to use the new Type property for registration so the configuration is respected.

StefanOssendorf (changes requested):

Source/Csla.Channels.RabbitMq/RabbitMqChannelExtensions.cs
rockfordlhotka: Doh!

StefanOssendorf (approved):
lgtm

PR #4072, opened 2024-07-01
Mark assembly as supported in the browser

### LINKED ISSUE #1894: In Blazor wasm project, use SupportedPlatform/UnsupportedOSPlatform feature
image

kant2002:
- Mark unsupported API in the browser
- Enforce future marking of unsupported in the browser API via editorconfig

Closes #1894


Source/Csla/Server/DataPortal.cs
```diff
@@ -156,10 +156,21 @@ private Reflection.ServiceProviderMethodCaller ServiceProviderMethodCaller
             break;
 #endif
           case TransactionalTypes.TransactionScope:
+#if NET8_0_OR_GREATER
+            if (OperatingSystem.IsBrowser())
+            {
+              throw new PlatformNotSupportedException();
+            }
+            else
```
StefanOssendorf, line 164: The else is not necessary here.

Source/Csla/Server/DataPortal.cs
```diff
@@ -257,9 +268,20 @@ public async Task<DataPortalResult> Fetch(Type objectType, object criteria, Data
...
           case TransactionalTypes.TransactionScope:
-            var broker = _applicationContext.CreateInstanceDI<DataPortalBroker>();
-            portal = new TransactionalDataPortal(broker, method.TransactionalAttribute);
-            result = await portal.Fetch(objectType, criteria, context, isSync).ConfigureAwait(false);
+#if NET8_0_OR_GREATER
+            if (OperatingSystem.IsBrowser())
+            {
+              throw new PlatformNotSupportedException();
+            }
+            else
```
StefanOssendorf, line 276: Else is not necessary here

Source/Csla/Server/DataPortal.cs
```diff
@@ -351,9 +373,20 @@ private async Task<DataPortalResult> Execute(Type objectType, object criteria, D
...
```
StefanOssendorf, line 381: Else not necessary here.

Source/Csla/Server/DataPortal.cs
```diff
@@ -484,9 +517,20 @@ public async Task<DataPortalResult> Update(object obj, DataPortalContext context
...
           case TransactionalTypes.TransactionScope:
-            var broker = _applicationContext.CreateInstanceDI<DataPortalBroker>();
-            portal = new TransactionalDataPortal(broker, method.TransactionalAttribute);
-            result = await portal.Update(obj, context, isSync).ConfigureAwait(false);
+#if NET8_0_OR_GREATER
+            if (OperatingSystem.IsBrowser())
+            {
+              throw new PlatformNotSupportedException();
+            }
+            else
```
StefanOssendorf, line 525: else not necessary here

Source/Csla/Server/DataPortal.cs
```diff
@@ -587,9 +631,20 @@ public async Task<DataPortalResult> Delete(Type objectType, object criteria, Dat
...
           case TransactionalTypes.TransactionScope:
-            var broker = _applicationContext.CreateInstanceDI<DataPortalBroker>();
-            portal = new TransactionalDataPortal(broker, method.TransactionalAttribute);
-            result = await portal.Delete(objectType, criteria, context, isSync).ConfigureAwait(false);
+#if NET8_0_OR_GREATER
+            if (OperatingSystem.IsBrowser())
+            {
+              throw new PlatformNotSupportedException();
+            }
+            else
```
StefanOssendorf, line 639: else not necessary here

Source/Csla/DataPortalClient/HttpProxy.cs
```diff
@@ -82,7 +85,15 @@ HttpClientHandler CreateDefaultHandler()
         var handler = new HttpClientHandler();
         if (!Options.UseTextSerialization)
         {
-          handler.AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate;
+#if NET8_0_OR_GREATER
```
StefanOssendorf, line 88: Instead of this double `#if` wouldn't it be easier to do something like this?
```csharp
bool browserOS = false;
#if NET8_0_OR_GREATER
browserOS = OperatingSystem.IsBrowser();
#endif
if(!browserOS && !Options.UseTextSerialization)
{
 // ...
}
```

Source/Csla/DataPortalClient/HttpProxy.cs
```diff
@@ -173,6 +187,12 @@ private async Task<byte[]> CallViaHttpClient(byte[] serialized, string operation
 
     private byte[] CallViaWebClient(byte[] serialized, string operation, string routingToken)
```
StefanOssendorf, line 188: Shouldn't we mark the method also with unspportedOS?

StefanOssendorf (changes requested):

Source/Csla/DataPortalClient/HttpProxy.cs
kant2002: That would bubble problem to callsite, which is select between sync/async methods. Currently compiler analyze that if we on platform "browser" then we throw PNSE and that's valid way to "support" something.

Given that if I mark this method as unsupported, I have to perform check when `sync == true` and I still have to throw PNSE there. OR! If you think that always use async versions in the browser is valid idea, that maybe reasonable.

Source/Csla/DataPortalClient/HttpProxy.cs
kant2002: Honestly I try to avoid any variable creation (maybe that's overkill), but can you take a look at other variant which I come up with.

StefanOssendorf: I'm not against that. I personally prefer multiple return statements in a method to shortcut the method.
But csla tries to avoid multiple statements as far as I know. Maybe @rockfordlhotka  can shed some light/preference here.

Source/Csla/DataPortalClient/HttpProxy.cs
StefanOssendorf: The question is: Do `sync == true` csla operations currently work with blazor? As far as I understand these changes here it should _not_ work. So we can _assume_ async is the _only_ way in blazor/browser.
@rockfordlhotka You have experience with CSLA + Blazor. Do you use sync methods anywhere?

StefanOssendorf:
Wouldn't it make sense to create a helper class which encapsulates
```csharp
#if NET8_0_OR_GREATER
      if (OperatingSystem.IsBrowser())
      {
        throw new PlatformNotSupportedException();
      }
#endif
```
and is used everywhere this code is now?
Something like
```csharp
internal static class PlatformHelper
{
  public void ThrowNotSupportedIfBrowser()
  {
#if NET8_0_OR_GREATER
      if (OperatingSystem.IsBrowser())
      {
        throw new PlatformNotSupportedException();
      }
#endif
  }
}
```

Source/Csla/DataPortalClient/HttpProxy.cs
rockfordlhotka: I do prefer to have a single `return` statement in a method. However, if there's a compelling reason (code clarity, etc.) for multiple statements I'm ok with it as an exception.

Source/Csla/DataPortalClient/HttpProxy.cs
rockfordlhotka: Via `HttpProxy`? Yes, it would only be async in Blazor.

I could see where _LocalProxy_ might be used from a sync method, but never `HttpProxy`.

rockfordlhotka:
> Wouldn't it make sense to create a helper class

There is already a `Csla.Runtime` namespace/directory in the `Csla` project. It has an `IRuntimeInfo` service and implementation.

Perhaps that would be the place for any such helper code?

StefanOssendorf:
> > Wouldn't it make sense to create a helper class
> 
> There is already a `Csla.Runtime` namespace/directory in the `Csla` project. It has an `IRuntimeInfo` service and implementation.
> 
> Perhaps that would be the place for any such helper code?

All the better.

kant2002:
Dataflow analysis work on method body. if we put this If inside helper method, then we have to annotate everything with SupportedPlatform/UnsupportedPlatform 

For example
```
PlatformHelper.ThrowNotSupportedIfBrowser()
// Data flow does not know that all code below unreacheable on browser.
// This will produce warning that WebClient is not supported on the browser platform.
var client = new WebClient();
```

```
if (OperatingSystem.IsBrowser())
{
  throw new PlatformNotSupportedException();
}

// Data flow analysis is happy and say hello to you. He knows that this code 
// never reachable on browser.
var client = new WebClient();
```

Source/Csla/DataPortalClient/HttpProxy.cs
kant2002: I preserve current behaviour, since WebClient does not supported on the browser and blow up in runtime.

StefanOssendorf: Okay. Maybe we should create an analyzer to detect sync calls in a browser context? Is that even possible? 🤔

kant2002: I’m not sure that it would be interesting. Maybe if we have additional methods for sync calls. Otherwise we are in the data-flow business and I’m not confident that this is worth it. 

Isn’t this just blowup during regular dev flow? I think I can add proper messages to indicate that this is limitations Blazor/browser and maybe point to some wiki page to make this actionable to developers

StefanOssendorf: You are probably right. We should have a simple message in the PNSE like `Sync data portal methods are not supported in a browser enviroment. Please use the async variant instead.`

StefanOssendorf (commented):
For me this looks good. @rockfordlhotka your final thoughts?

rockfordlhotka (approved):

PR #4073, opened 2024-07-01
Remove redundant imports

kant2002:
They produce following errors (for all projects in the solution)
```
C:\d\github\csla\Samples\BlazorExample\BlazorExample\DataAccess\DataAccess.csproj(7,3): warning MSB4011: "C:\d\github\csla\Samples\BlazorExample\Directory.Packages.props" cannot be imported again. It was alre
ady imported at "C:\Program Files\dotnet\sdk\8.0.300\NuGet.props (35,3)". This is most likely a build authoring error. This subsequent import will be ignored.
```

Related to #3207

### DIFF
--- a/Samples/BlazorExample/BlazorExample/BlazorExample.Client/BlazorExample.Client.csproj
+++ b/Samples/BlazorExample/BlazorExample/BlazorExample.Client/BlazorExample.Client.csproj
@@ -8,8 +8,6 @@
     <StaticWebAssetProjectMode>Default</StaticWebAssetProjectMode>
   </PropertyGroup>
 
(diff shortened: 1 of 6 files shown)


kant2002:
Notice that in BlazorExample which I use a lot.

StefanOssendorf (approved):
LGTM

StefanOssendorf:
Why is the CI action not running?

rockfordlhotka:
> Why is the CI action not running?

It only triggers when code changes in `csla/source`.

I have it set to required - maybe I can't do that anymore, because when people change docs or samples or whatever, obviously the CI build doesn't run.

Or maybe there's a way to have it "run" but do nothing in those cases?

kant2002:
Maybe separate CI builds for each of the sample, or one additional workflow to build them all?

PR #4091, opened 2024-07-09
Features/analyzer static nameof

### LINKED ISSUE #1747: Analyzer: Update RegisterProperty parameters

Older code often has a `RegisterProperty` like this:

```c#
    public static readonly PropertyInfo<int> CustomerIdProperty = RegisterProperty<int>(c => c.CustomerId);
```

The modern equivalent is this (using `nameof`):

```c#
    public static readonly PropertyInfo<int> CustomerIdProperty = RegisterProperty<int>(nameof(CustomerId));
```

It would be nice to have an analyzer warn I'm using the old style, and even better, offer a fix across my codebase.

luizfbicalho:
Closes #1747 

To explain here 

I created 2 fixes

```csharp
private static readonly PropertyInfo<string> TestProperty = RegisterProperty<string>(c => c.TestName, "Test Name");

```
change to 
```csharp
private static readonly PropertyInfo<string> TestProperty = RegisterProperty<string>(nameof(TestName), "Test Name");

```

and 
```csharp

          public static string TestPropertyName = "TypeName";
          private static readonly PropertyInfo<string> TestProperty = RegisterProperty<string>(TestPropertyName, "Test Name");
```
to 

```csharp

          public static string TestPropertyName = "TypeName";
          private static readonly PropertyInfo<string> TestProperty = RegisterProperty<string>(nameof(TestPropertyName), "Test Name");
```
I didn't create for string constant because it's already caught by CA1507

CA1507	Using nameof helps keep your code valid when refactoring.


StefanOssendorf:
Isn't the second quickfix wrong?
From
`private static readonly PropertyInfo<string> TestProperty = RegisterProperty<string>(TestPropertyName, "Test Name");`
to
`private static readonly PropertyInfo<string> TestProperty = RegisterProperty<string>(nameof(TestPropertyName), "Test Name");`
changes the semantic. 
In the first one the value of `TestPropertyName` (=`"TypeName"`) will be used but in the second one `"TestPropertyName"` is the new value used instead of `"TypeName"`.
Or am I confusing things here?

luizfbicalho:
> Or am I confusing things here?

I added all possilities that I imagined, even if didn't find it right , so we could discuss all the possibilities,
If I put the string value "TypeName" I open the case to more problems than solutions in my opinion.

@StefanOssendorf do you have any ideas?

StefanOssendorf:
After reading the issue for me the first code fix is totally sufficient. We don't have to support everything a user can come up with.
So I would just add the analyzer+codefix mentioned in the issue.
Or do you see other constellation which are worth considering/discussing here?

luizfbicalho:
> After reading the issue for me the first code fix is totally sufficient. We don't have to support everything a user can come up with. So I would just add the analyzer+codefix mentioned in the issue. Or do you see other constellation which are worth considering/discussing here?

I'll remove the second part

rockfordlhotka (approved):

PR #4093, opened 2024-07-10
Create option for disabling strong name check during serialization

### LINKED ISSUE #3036: Consider adding an option for MobileFormatter to ignore version/signing on deserialization
Although very risky and unsafe for a user, it may be possible for `MobileFormatter` to have an option to allow deserialization into a type of the same name/assembly where the version and signing information doesn't match.

The referenced discussion thread contains info about why this is risky and may lead to people having hard-to-find and hard-to-solve issues in their app.

---

Number 4: Apparently we are not using the SerializationInfo correctly. We add all of an object's properties using info.AddValue function regardless of if the value is a custom type or IMobileObject type or System type. 

1. Add an option to turn on assembly/type matching somewhere in the CSLA options (default setting is off)
2. Change MobileFormatter to look at the matching options

Should just be those two.

_Originally posted by @crazyfox55 in https://github.com/MarimerLLC/csla/issues/3021#issuecomment-1196007880_

kant2002:
Closes #3036


Source/Csla/Serialization/Mobile/Configuration/MobileFormatterOptions.cs
```diff
@@ -18,6 +18,11 @@ public class MobileFormatterOptions
   /// Gets the list of custom serializers.
   /// </summary>
   public List<ITypeMap> CustomSerializers { get; } = [];
+  
+  /// <summary>
+  /// Gets a value indicating whether to use strong name checks during serialization/deserialization process.
+  /// </summary>
+  public bool UseStrongNamesCheck { get; internal set; } = true;
```
rockfordlhotka, line 25: Shouldn't this be a private set?

Or just make the setter public and let the developer set the property during configuration (skipping the two methods)?

rockfordlhotka (changes requested):
There needs to be some tests for this.

At the very least, use reflection to invoke the method that returns the type name to ensure it does and does not include the version info.

Source/Csla/Serialization/Mobile/Configuration/MobileFormatterOptions.cs
kant2002: I have impression that you want fluent interface for the options. At least this is seems to be the pattern. Maybe it make sense to have both imperative and fluent API?

StefanOssendorf:
I prefer the fluent interface. And I have a task (in my personal list) to
make all options fluents when I have the time xD

Andrii Kurdiumov ***@***.***> schrieb am Do., 11. Juli 2024,
07:26:

Source/Csla/Serialization/Mobile/Configuration/MobileFormatterOptions.cs
rockfordlhotka: True, fluent is easier for the end developer. The setter should still be private right?

Also, the fluent method could accept a parameter:

```c#
  public MobileFormatterOptions StrongNamesCheck(bool check)
    => UseStrongNamesCheck = check;
```

kant2002: I prefer to have it both ways, that's most inclusive in my opinion. that way library do not force preferences on developers, and I think this is not super complicated to maintain in the long run.

kant2002: I have mixed signals. Error! Error! Operator intervention required.

rockfordlhotka: Let's go with Stefan's model, I see the value in the argument.

StefanOssendorf (approved):
Looks good to me. lgtm @rockfordlhotka

rockfordlhotka (approved):

rockfordlhotka (approved):

PR #4100, opened 2024-07-15
Localize code fixes and analyzers

### LINKED ISSUE #1352: Text in analyzers should be localizable
Right now the analyzers have hard-coded US English text in them as error messages that are displayed to developers.

That text should be moved to a resource file so it can be localized. At least that is what we _think_ should happen. Some research is required to figure out if that's the right approach to localize analyzers that are running inside of Visual Studio.

kant2002:
Closes #1352


StefanOssendorf:
If it's a draft please make it a real draft. Otherwise rocky may just merge it ;)

rockfordlhotka:
My first scan through the code changes is quite positive, I like what you've done so far.

Can you provide more detail about the three decision points, and what you recommend?

kant2002:
@StefanOssendorf unfortunately I press buttons first and think second. So I miss that opportunity 😄 

@rockfordlhotka one by one.
- previously in tests was constants which is used in analyzers. I directly use string literals now. Anyway I cannot leave test code without changes. Maybe i can get to read from prebuilt LocalizationString but for me that’s tying too much to prod code. 
- just voice your opinion on how I name resource strings
- Just inform that I remove not used code. Maybe it indeed used by some obscure ways.

kant2002:
@rockfordlhotka can you take a look?

rockfordlhotka:
I'm out of town this week, and will look when I get back.

rockfordlhotka (approved):

PR #4124, opened 2024-07-28
Features/source generator autoimplement

### LINKED ISSUE #4098: Create source generator for business class properties
My thought in creating this issue is to first have a discussion about how such a generator should work, and the scope it should cover.

> @rockfordlhotka you gave me an amazing idea
> 
> after this issue I could create a source generator just like this
> 
> ```cs
> public partial class Test
> {	
>     [AutoPropertyInfo]
>     private string _city;
> 
> }
> ```
> 
> and generate this code
> 
> ```cs
> public partial class Test
> {
> 	public Test()
> 	{
> 		_city = CityProperty.DefaultValue;
> 	}
> 	
> 	public static readonly PropertyInfo<string> CityProperty = RegisterProperty<string>(nameof(City), RelationshipTypes.PrivateField);
> 	public string City
>     {
>         get => GetProperty(CityProperty, _city);
>         set => SetProperty(CityProperty, ref _city, value);
>     }
> }
> ```
> 
> this solves the point that I was looking for a way to remove all this extra code and also control the result class If you like the idea you can create the issue and assign to me.

Yep. Was also looking how to get rid of that boilerplate. I'll probably fiddle with that and write one (in my repo).
I do hope the source gens will be released independent of the main csla stuff. I think they need to have a quicker release cycle when we _really_ start supporting and using source generators.

_Originally posted by @StefanOssendorf in https://github.com/MarimerLLC/csla/issues/2334#issuecomment-2225601699_

luizfbicalho:
Closes #4098


StefanOssendorf:
I have only skimmed over some files. But there seem to be a lot of xml-comment c&p errors. Especially on classes which are used by users. Please double check that.

luizfbicalho:
> I have only skimmed over some files. But there seem to be a lot of xml-comment c&p errors. Especially on classes which are used by users. Please double check that.

Sorry about that , I had it on my list and checked it as done wrongly, I'll do that later

luizfbicalho:
> I have only skimmed over some files. But there seem to be a lot of xml-comment c&p errors. Especially on classes which are used by users. Please double check that.

I changed the comments, do you think that we should change the namespace of the attribute? or some other namespaces?

Source/Csla.Generators/cs/AutoImplementProperties/readme.md
rockfordlhotka, line 1: Is this file accurate for the latest AutoImplementProperties behaviors?

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/Internals/HashCode.cs
rockfordlhotka, line 1: We can't use the built-in hashcode capabilities of .NET?

Also, just to confirm, this is original code, and not something brought in from another source?

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/Internals/HashCode.cs
```diff
@@ -0,0 +1,371 @@
+﻿using System.ComponentModel;
```
rockfordlhotka, line 1: Missing standard file header/copyright

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/AutoImplement/Discovery/DefinitionExtractionContext.cs
rockfordlhotka, line 1: A general question I have with all the support classes that abstract Roslyn: should these be part of this specific generator, or are they useful for _any_ generator and should be their own project/dll that can be reused in the future?

rockfordlhotka (changes requested):

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/Internals/HashCode.cs
luizfbicalho: This is a gift from @StefanOssendorf

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/AutoImplement/Discovery/DefinitionExtractionContext.cs
luizfbicalho: I think that this is very specific, I copied from the other generator but it's like a linq query over a database, it's the view model that I need here.
surely they are similar but they should evolve not in the same direction

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/Internals/HashCode.cs
rockfordlhotka: I'm fine with it as long as the IP is clearly available for use in CSLA. Can you confirm that @StefanOssendorf ?

Source/Csla.Generators/cs/AutoImplementProperties/readme.md
luizfbicalho: This I forgot, sending now

luizfbicalho:
> > > I have only skimmed over some files. But there seem to be a lot of xml-comment c&p errors. Especially on classes which are used by users. Please double check that.
> > 
> > 
> > I changed the comments, do you think that we should change the namespace of the attribute? or some other namespaces?
> 
> I think the attributes should probably be in the `Csla` namespace so they are available by default if someone has added the generator package to their project.

Changed to Csla

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/Internals/HashCode.cs
ossendorf-at-hoelscher: I don't think it's my IP I guess?
I took it from here (there's also a comment in my gen that it's taken from there) https://github.com/andrewlock/StronglyTypedId/blob/master/src/StronglyTypedIds/HashCode.cs based/recommended from the author here https://andrewlock.net/creating-a-source-generator-part-9-avoiding-performance-pitfalls-in-incremental-generators/#4-watch-out-for-collection-types

ossendorf-at-hoelscher: In that regard maybe we should consider https://www.nuget.org/packages/Microsoft.Bcl.HashCode/ as a substitute?

rockfordlhotka: I think that is best, please make that change @luizfbicalho.

luizfbicalho:
Updated

rockfordlhotka (approved):

Issue #4123, opened 2024-07-26
ReadWriteTimeout in httpproxy.cs


amselem:
**Describe the bug**

-When using HttpProxy I'm experiencing consistent timeouts in long running dataportal operations, specifically when > 5 minutes. I always get a TimeOut exactly 5:00 minutes after starting the call.

-Setting the HttpProxy.TimeOut property to a value > 5 minutes didn't help.

-I found that there's a default 5 minute timeout value in the ReadWriteTimeOut property of the HttpWebRequest class (https://learn.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest.readwritetimeout?view=net-8.0). 
The ReadWriteTimeOut  is different from the  TimeOut (which is correctly set by the httpProxy timeout parameter), so please note that we have two timeouts in play here.

**Version and Platform**
CSLA version: 5.5.3

** To repro
Simply launch a long running dataportal call using the httpproxy, and wait 5:00.  The httpproxy TimeOut value won't help beyond 5:00

**Additional context**

I could subclass HttpProxy to fix it but since HttpProxy is allowing to manage the TimeOut value I thought that it would be interesting in adding this change to the main code

Suggested code change:

HttpProxy.cs , class DefaultWebClient  , change from :

```
protected override WebRequest GetWebRequest(Uri address)
      {
        var req = base.GetWebRequest(address) as HttpWebRequest;
        if (Timeout > 0)
          req.Timeout = Timeout;
        return req;
      }
```

to this  (For simplicity I used the same TimeOut property value for setting both TimeOut and ReadWriteTimeOut , but maybe a new separated property should be created in HttpProxy) :

```
protected override WebRequest GetWebRequest(Uri address)
      {
        var req = base.GetWebRequest(address) as HttpWebRequest;
        if (Timeout > 0)
        {
          req.Timeout = Timeout;
          req.ReadWriteTimeout = Timeout;  // <-- this
         }
         return req;
      }

```

amselem:
Yes I'm in v5.5.3  under  .Net 8.0 but I think it is relevant in the current/future versions since the dependency with webclient is still there (look at the current httproxy.cs)

rockfordlhotka:
I think we'll address this in v9. I am not even sure I can build v5 anymore, that's very old now 😄

luizfbicalho:
Can it be the same timeout or we should have two properties to configure that?

Can you assign that to me?

rockfordlhotka:
[...]

PR #4167, opened 2024-08-15
Features/add readwrite timeout

luizfbicalho:
Closes: #4123

### DIFF
--- a/Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
+++ b/Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
@@ -23,7 +23,7 @@ namespace Csla.Channels.RabbitMq
     /// Creates an instance of the object, initializing
     /// it to use the supplied URL.
     /// </summary>
-    /// <param name="applicationContext"></param>
+    /// <param name="applicationContext">The application context.</param>
     /// <param name="options">Proxy options</param>
     /// <exception cref="ArgumentNullException"><paramref name="applicationContext"/> or <paramref name="options"/> is <see langword="null"/>.</exception>
     public RabbitMqProxy(ApplicationContext applicationContext, RabbitMqProxyOptions options)
@@ -33,14 +33,8 @@ namespace Csla.Channels.RabbitMq
         throw new ArgumentNullException(nameof(options));
 
       DataPortalUrl = options.DataPortalUrl;
-      _timeout = options.Timeout;
+      Options = options;
     }
-
-    private readonly TimeSpan _timeout;
-    /// <summary>
-    /// Gets or sets the timeout value in seconds for the RabbitMQ request.
-    /// </summary>
-    protected override TimeSpan Timeout => _timeout;
 
     /// <summary>
     /// Gets or sets the connection to the RabbitMQ service.
@@ -65,12 +59,17 @@ namespace Csla.Channels.RabbitMq
     private ProxyListener? QueueListener { get; set; }
 
     /// <summary>
+    /// Gets the proxy options.
+    /// </summary>
+    public RabbitMqProxyOptions Options { get; }
+
+    /// <summary>
     /// Method responsible for creating the Connection,
     /// Channel, ReplyQueue, and DataPortalQueueName values
     /// used for bi-directional communication.
     /// </summary>
 #if NET8_0_OR_GREATER
-    [MemberNotNull(nameof(Connection), nameof(Channel), nameof(QueueListener))]
+      [MemberNotNull(nameof(Connection), nameof(Channel), nameof(QueueListener))]
 #endif
     protected virtual void InitializeRabbitMQ()
     {
@@ -229,7 +228,7 @@ namespace Csla.Channels.RabbitMq
 
       SendMessage(QueueListener!.ReplyQueue!.QueueName, correlationId, operation, serialized);
 
-      var timeout = Task.Delay(Timeout);
+      var timeout = Task.Delay(Options.Timeout);
       if (await Task.WhenAny(wip.ResetEvent.WaitAsync(), timeout) == timeout)
         throw new TimeoutException();
 
--- a/Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
+++ b/Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
@@ -61,7 +61,7 @@ namespace Csla.Channels.RabbitMq
     /// <summary>
     /// Gets the proxy options.
     /// </summary>
-    public RabbitMqProxyOptions Options { get; }
+    private RabbitMqProxyOptions Options { get; }
 
     /// <summary>
     /// Method responsible for creating the Connection,
@@ -69,7 +69,7 @@ namespace Csla.Channels.RabbitMq
     /// used for bi-directional communication.
     /// </summary>
 #if NET8_0_OR_GREATER
-      [MemberNotNull(nameof(Connection), nameof(Channel), nameof(QueueListener))]
+    [MemberNotNull(nameof(Connection), nameof(Channel), nameof(QueueListener))]
 #endif
     protected virtual void InitializeRabbitMQ()
     {
--- a/Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
+++ b/Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
@@ -33,8 +33,14 @@ namespace Csla.Channels.RabbitMq
         throw new ArgumentNullException(nameof(options));
 
       DataPortalUrl = options.DataPortalUrl;
-      Timeout = (int)options.Timeout.TotalSeconds;
+      _timeout = (int)options.Timeout.TotalSeconds;
     }
+
+    private readonly int _timeout;
+    /// <summary>
+    /// Gets or sets the timeout value in seconds for the RabbitMQ request.
+    /// </summary>
+    public override int Timeout => _timeout;
 
     /// <summary>
     /// Gets or sets the connection to the RabbitMQ service.
--- a/Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
+++ b/Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
@@ -33,14 +33,14 @@ namespace Csla.Channels.RabbitMq
         throw new ArgumentNullException(nameof(options));
 
       DataPortalUrl = options.DataPortalUrl;
-      _timeout = (int)options.Timeout.TotalSeconds;
+      _timeout = options.Timeout;
     }
 
-    private readonly int _timeout;
+    private readonly TimeSpan _timeout;
     /// <summary>
     /// Gets or sets the timeout value in seconds for the RabbitMQ request.
     /// </summary>
-    public override int Timeout => _timeout;
+    protected override TimeSpan Timeout => _timeout;
 
     /// <summary>
     /// Gets or sets the connection to the RabbitMQ service.
@@ -229,7 +229,7 @@ namespace Csla.Channels.RabbitMq
 
       SendMessage(QueueListener!.ReplyQueue!.QueueName, correlationId, operation, serialized);
 
-      var timeout = Task.Delay(Timeout * 1000);
+      var timeout = Task.Delay(Timeout);
       if (await Task.WhenAny(wip.ResetEvent.WaitAsync(), timeout) == timeout)
         throw new TimeoutException();
 
--- a/Source/Csla/DataPortalClient/DataPortalProxy.cs
+++ b/Source/Csla/DataPortalClient/DataPortalProxy.cs
@@ -40,11 +40,6 @@ namespace Csla.DataPortalClient
     /// </summary>
     public virtual bool IsServerRemote => true;
 
-    /// <summary>
-    /// Gets or sets the Client timeout
-    /// (0 uses default timeout).
-    /// </summary>
-    protected virtual TimeSpan Timeout { get; }
 
     /// <summary>
     /// Gets the URL address for the data portal server
--- a/Source/Csla/DataPortalClient/DataPortalProxy.cs
+++ b/Source/Csla/DataPortalClient/DataPortalProxy.cs
@@ -6,11 +6,11 @@
 // <summary>Abstract data portal proxy with common data portal proxy behaviour. Implements IDataPortalProxy</summary>
 //-----------------------------------------------------------------------
 
-using Csla.Serialization.Mobile;
+using Csla.Configuration;
 using Csla.Serialization;
+using Csla.Serialization.Mobile;
 using Csla.Server;
 using Csla.Server.Hosts.DataPortalChannel;
-using Csla.Configuration;
 
 namespace Csla.DataPortalClient
 {
@@ -44,7 +44,7 @@ namespace Csla.DataPortalClient
     /// Gets or sets the Client timeout
     /// in milliseconds (0 uses default timeout).
     /// </summary>
-    public virtual int Timeout { get; set; }
+    public virtual int Timeout { get; }
 
     /// <summary>
     /// Gets the URL address for the data portal server
--- a/Source/Csla/DataPortalClient/DataPortalProxy.cs
+++ b/Source/Csla/DataPortalClient/DataPortalProxy.cs
@@ -42,9 +42,9 @@ namespace Csla.DataPortalClient
 
     /// <summary>
     /// Gets or sets the Client timeout
-    /// in milliseconds (0 uses default timeout).
+    /// (0 uses default timeout).
     /// </summary>
-    public virtual int Timeout { get; }
+    protected virtual TimeSpan Timeout { get; }
 
     /// <summary>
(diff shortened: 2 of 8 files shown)


Source/Csla/DataPortalClient/DataPortalProxy.cs
```diff
@@ -44,7 +44,7 @@ public DataPortalProxy(ApplicationContext applicationContext)
     /// Gets or sets the Client timeout
     /// in milliseconds (0 uses default timeout).
     /// </summary>
-    public virtual int Timeout { get; set; }
+    public virtual int Timeout { get; }
```
rockfordlhotka, line 47: Could this have a protected set instead of delegating the field down to subclasses?

Source/Csla/DataPortalClient/HttpProxyOptions.cs
```diff
@@ -36,6 +36,16 @@ public class HttpProxyOptions
     /// </summary>
     public Func<IServiceProvider, HttpClient> HttpClientFactory { get; set; } = DefaultHttpFactory;
 
+    /// <summary>
+    /// Gets or sets the timeout value in milliseconds for the HTTP request.
+    /// </summary>
+    public int Timeout { get; set; }
```
rockfordlhotka, line 42: As long as we're doing this for CSLA 9, let's make this into a TimeSpan - breaking change, but better code.

Source/Csla/DataPortalClient/HttpProxyOptions.cs
```diff
@@ -36,6 +36,16 @@ public class HttpProxyOptions
...
+
+    /// <summary>
+    /// Gets or sets the read-write timeout value in milliseconds for the HTTP request.
+    /// </summary>
+    public int ReadWriteTimeout { get; set; }
```
rockfordlhotka, line 47: Let's make this a TimeSpan

Source/Csla/DataPortalClient/DataPortalProxy.cs
rockfordlhotka: Also, let's make this a TimeSpan

Source/Csla/DataPortalClient/HttpProxy.cs
```diff
@@ -255,17 +259,21 @@ private string CreateOperationTag(string operation, string versionToken, string
 #if NET8_0_OR_GREATER
     [UnsupportedOSPlatform("browser")]
 #endif
-    private class DefaultWebClient : WebClient
+    private class DefaultWebClient(int timeout, int readWriteTimeout) : WebClient
```
rockfordlhotka, line 262: Let's make these into TimeSpan values

rockfordlhotka (changes requested):

Source/Csla/DataPortalClient/DataPortalProxy.cs
luizfbicalho: Or we could remove this property and let just the options in the http proxy, even because that this timeout means different things in http and rabbitmq

rockfordlhotka: Excellent - let's make that happen.

Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
```diff
@@ -33,9 +33,15 @@ public RabbitMqProxy(ApplicationContext applicationContext, RabbitMqProxyOptions
         throw new ArgumentNullException(nameof(options));
 
       DataPortalUrl = options.DataPortalUrl;
-      Timeout = (int)options.Timeout.TotalSeconds;
+      _timeout = options.Timeout;
     }
 
+    private readonly TimeSpan _timeout;
+    /// <summary>
+    /// Gets or sets the timeout value in seconds for the RabbitMQ request.
```
StefanOssendorf, line 41: The value is not necessarily in seconds. It's a time span and how that's used by rabbit mq should here be irrelevant.

Source/Csla/DataPortalClient/HttpCompressionProxy.cs
```diff
@@ -62,27 +62,32 @@ protected override void SetHttpRequestHeaders(HttpRequestMessage request)
     /// </summary>
     protected override WebClient GetWebClient()
     {
-      var client = new CompressionWebClient(Timeout);
```
StefanOssendorf, line 65: Why did your remove the `!`? The framework guarantees it's not null. Furthermore with this change a lot of null checks are now missing.

StefanOssendorf: In this case we because we derive from our own class here.

Source/Csla/DataPortalClient/HttpCompressionProxy.cs
```diff
@@ -62,27 +62,32 @@ protected override void SetHttpRequestHeaders(HttpRequestMessage request)
...
       return client;
     }
 
 #pragma warning disable SYSLIB0014
-    private class CompressionWebClient(int timeout) : WebClient
+    private class CompressionWebClient(TimeSpan timeout, TimeSpan readWriteTimeout) : WebClient
     {
       protected override WebRequest GetWebRequest(Uri address)
       {
-        var webRequest = base.GetWebRequest(address)!;
```
StefanOssendorf, line 76: Same here

Source/Csla/DataPortalClient/HttpProxyOptions.cs
```diff
@@ -36,6 +36,16 @@ public class HttpProxyOptions
...
 
+    /// <summary>
+    /// Gets or sets the timeout value for the HTTP request.
+    /// </summary>
+    public TimeSpan Timeout { get; set; }
+
+    /// <summary>
+    /// Gets or sets the read-write timeout value for the HTTP request.
+    /// </summary>
+    public TimeSpan ReadWriteTimeout { get; set; }
```
StefanOssendorf, line 47: Shouldn't we state that this only applies to sync http calls and not async ones with HttpClient?

luizfbicalho: I didn't get this part of the async in the docs, [https://learn.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest.readwritetimeout?view=net-8.0&redirectedfrom=MSDN#System_Net_HttpWebRequest_ReadWriteTimeout](https://learn.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest.readwritetimeout?view=net-8.0&redirectedfrom=MSDN#System_Net_HttpWebRequest_ReadWriteTimeout)

Source/Csla/DataPortalClient/HttpCompressionProxy.cs
luizfbicalho: In one class there was the` ! `and in the other wasn't, I just made them similar, I can put it back.

In fact I don't like the `! ` because we are forcing something that the compiler isn't sure about, I like better to enable nullable and mark all nullable as errors.

and evn after that I just use ! with a good justification

StefanOssendorf: Looks like I missed the other part 😅. That project should already be NRT aware and yes I'll make a PR to enable NRT as errors for already migrated projects. So please revert that change and I'll revisit this part in the near future.

Source/Csla/DataPortalClient/HttpProxyOptions.cs
StefanOssendorf: We use `WebClient` _only_ for sync calls. For async calls we use `HttpClient`. So this timeout only applies to sync calls in CSLA and I think we should state that to avoid misunderstanding. Or am I misunderstand your comment here? 😅

Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
```diff
@@ -58,13 +58,18 @@ public RabbitMqProxy(ApplicationContext applicationContext, RabbitMqProxyOptions
     /// </summary>
     private ProxyListener? QueueListener { get; set; }
 
+    /// <summary>
+    /// Gets the proxy options.
+    /// </summary>
+    public RabbitMqProxyOptions Options { get; }
```
StefanOssendorf, line 64: Shouldn't this be private? And it could be a field?

Source/Csla.Channels.RabbitMq/RabbitMqProxy.cs
```diff
@@ -58,13 +58,18 @@ public RabbitMqProxy(ApplicationContext applicationContext, RabbitMqProxyOptions
...
+
     /// <summary>
     /// Method responsible for creating the Connection,
     /// Channel, ReplyQueue, and DataPortalQueueName values
     /// used for bi-directional communication.
     /// </summary>
 #if NET8_0_OR_GREATER
-    [MemberNotNull(nameof(Connection), nameof(Channel), nameof(QueueListener))]
+      [MemberNotNull(nameof(Connection), nameof(Channel), nameof(QueueListener))]
```
StefanOssendorf, line 72: Why adding the spaces here? Shouldn't it be on the same level as the method and xml-doc?

Source/Csla/DataPortalClient/HttpCompressionProxy.cs
StefanOssendorf: #4171 changes the warnings to be errors

Source/Csla/DataPortalClient/HttpProxy.cs
```diff
@@ -70,14 +70,14 @@ protected virtual HttpClient GetHttpClient()
       if (_httpClient == null)
       {
         var handler = GetHttpClientHandler() ?? CreateDefaultHandler();
-        
+
         _httpClient = new HttpClient(handler);
-        if (Timeout > 0)
+        if (Options.Timeout.Milliseconds > 0)
```
StefanOssendorf, line 75: You could also do: `Options.Timeout > TimeSpan.Zero`

Source/Csla/DataPortalClient/HttpCompressionProxy.cs
```diff
@@ -78,11 +78,15 @@ protected override WebRequest GetWebRequest(Uri address)
         if (webRequest is HttpWebRequest httpWebRequest)
         {
           httpWebRequest.AutomaticDecompression = DecompressionMethods.Deflate | DecompressionMethods.GZip;
+          if (readWriteTimeout.Milliseconds > 0)
```
StefanOssendorf, line 81: You could use: `readWriteTimeout > TimeSpan.Zero`

Source/Csla/DataPortalClient/HttpCompressionProxy.cs
```diff
@@ -78,11 +78,15 @@ protected override WebRequest GetWebRequest(Uri address)
...
+          {
+            httpWebRequest.ReadWriteTimeout = readWriteTimeout.Milliseconds;
+          }
         }
 
-        if (timeout > 0)
+        if (timeout.Milliseconds > 0)
```
StefanOssendorf, line 87: You could use: `timeout > TimeSpan.Zero`

Source/Csla/DataPortalClient/HttpProxy.cs
```diff
@@ -255,17 +255,23 @@ private string CreateOperationTag(string operation, string versionToken, string
...
 
       protected override WebRequest GetWebRequest(Uri address)
       {
-        var req = base.GetWebRequest(address) as HttpWebRequest;
-        if (Timeout > 0)
-          req.Timeout = Timeout;
+        var req = base.GetWebRequest(address)!;
+        if (req is HttpWebRequest httpWebRequest)
+        {
+          if (readWriteTimeout.Milliseconds > 0)
```
StefanOssendorf, line 266: You could use: `readWriteTimeout > TimeSpan.Zero`

Source/Csla/DataPortalClient/HttpProxy.cs
```diff
@@ -255,17 +255,23 @@ private string CreateOperationTag(string operation, string versionToken, string
...
+          {
+            httpWebRequest.ReadWriteTimeout = readWriteTimeout.Milliseconds;
+          }
+        }
+        if (timeout.Milliseconds > 0)
```
StefanOssendorf, line 271: You could use: `timeout > TimeSpan.Zero`

Source/Csla/DataPortalClient/HttpProxyOptions.cs
luizfbicalho: I thought that in .net framework this timeout wasn't used in async calls, and I didn't find any comment about that

StefanOssendorf:
For the future: Please don't mark comments as resolved. It's the reviewers task to decide whether the comment was resolved or not. Thanks :)

Source/Csla/DataPortalClient/HttpProxyOptions.cs
```diff
@@ -36,6 +36,16 @@ public class HttpProxyOptions
     /// </summary>
     public Func<IServiceProvider, HttpClient> HttpClientFactory { get; set; } = DefaultHttpFactory;
 
+    /// <summary>
+    /// Gets or sets the timeout value for the HTTP request (only sync calls).
+    /// </summary>
+    public TimeSpan Timeout { get; set; }
```
StefanOssendorf, line 42: This timeout is used for sync and async calls. So please remove the ("only sync calls") here.

StefanOssendorf (changes requested):
Please make a breaking change entry in the https://github.com/MarimerLLC/csla/blob/main/docs/Upgrading%20to%20CSLA%209.md file that the Timeout now has to be set via the options.

luizfbicalho:
> For the future: Please don't mark comments as resolved. It's the reviewers task to decide whether the comment was resolved or not. Thanks :)

Sorry about that, I had never used this and in the first time I imagined that I needed to resolve to show you that I chaned this part, and you never said anything

StefanOssendorf:
All good. I think the easiest way to request a review again is to just make a name comment with "comment fixes pushed" or the like. So I'll get an e-mail when you have pushed your changes.

luizfbicalho:
Ok @StefanOssendorf  I made the changes, take a look id the documentation makes sense to you

docs/Upgrading to CSLA 9.md
```diff
@@ -174,3 +174,66 @@ CSLA 9 supports the use of nullable reference types in your code. This means tha
...
+  {
+    services.AddSingleton(); 
+    services.AddSingleton(new HttpProxyOptions { Timeout = TimeSpan.FromSeconds(30), ReadWriteTimeout = TimeSpan.FromSeconds(30) }); 
+    }
+  }
+}
+
+```
+
+###### Step 2: Use `HttpProxy` or `HttpCompressionProxy`
```
StefanOssendorf, line 209: I think this is not necessary, because the usage of `HttpProxy` and `HttpCompressionProxy` is hidden behind the data portals.

docs/Upgrading to CSLA 9.md
```diff
@@ -174,3 +174,66 @@ CSLA 9 supports the use of nullable reference types in your code. This means tha
...
+
+##### Configuration
+
+To configure the timeout for `HttpProxy` and `HttpCompressionProxy`, you need to set the `Timeout` and `ReadWriteTimeout` properties in the `HttpProxyOptions` class.
+
+##### Example
+
+Below is an example of how to configure and use the timeout settings in `HttpProxy` and `HttpCompressionProxy`.
+
+###### Step 1: Configure `HttpProxyOptions`
```
StefanOssendorf, line 192: Seeing this. We should extend the `HttpProxyOptionsExtensions` to set these values with the fluent api.

StefanOssendorf: Something like this:
```csharp
services.AddCsla(
    o => o.DataPortal(
        o2 => o2.AddClientSideDataPortal(
            cso => cso.UseHttpProxy(
                hpo => hpo.WithTimeout(TimeSpan.FromSeconds(30)).WithReadWriteTimeout(TimeSpan.FromSeconds(30))
            )
        )
    )
);
```

luizfbicalho: I added this code, check it first and then I change the docs

StefanOssendorf: The code looks good. Please adjust the docs 👍

luizfbicalho: Changed, i hope

docs/Upgrading to CSLA 9.md
StefanOssendorf: What's with this here?

luizfbicalho: Should I remove this whole step?

StefanOssendorf: Yeah. The typical user won't see the implementation.

luizfbicalho: changed

StefanOssendorf (approved):

StefanOssendorf:
LGTM
@rockfordlhotka

Issue #4160, opened 2024-08-12
CslaImplementProperties generates invalid code


rockfordlhotka:
**Describe the bug**
I tried using `CslaImplementProperties` in the ProjectTracker sample and it didn't work as expected. I wonder if the issue is because there is a custom base class between `Csla.BusinessBase` and the business class?

The issue appears to be that the generator creates an extra `using` like this:

```csharp
using System;
using Csla;
using ;
```

**Version and Platform**
CSLA version: 9

luizfbicalho:
Can you copy the class here?

rockfordlhotka:
> Can you copy the class here?

I did the following:

1. Open the ProjectTracker solution from `main`
2. Add the generator to the `BusinessLibrary` project
3. Apply the attribute to the `ProjectEdit` class
4. Switch some of the properties to use the generator syntax

Failure occurs.

luizfbicalho:
I changed the ProjectEdit.cs like this

```cs
  [Serializable]
  [CslaImplementPropertiesInterface<IProjectEdit>]
  public partial class ProjectEdit : CslaBaseTypes.BusinessBase<ProjectEdit>
  {

    private interface IProjectEdit
    {

      [Browsable(false)]
      [EditorBrowsable(EditorBrowsableState.Never)]
      public byte[] TimeStamp
      {
        get;
        set;
      }


      [Display(Name = "Project id")]
      public int Id
      {
        get;

      }


      [Display(Name = "Project name")]
      [Required]
      [StringLength(50)]
      public string Name
      {
        get;
        set;
      }


      public DateTime? Started
      {
        get;
        set;
      }


      public DateTime? Ended
      {
        get;
        set;
      }

      public string Description
      {
        get;
        set;
      }

      public ProjectResources Resources
      {
        get;

      }
    }
```

and update the generator to implement like this

```cs
//<auto-generated/>
using System;
using Csla;
using System.ComponentModel;
using System.ComponentModel.DataAnnotations;

namespace ProjectTracker.Library;

public partial class ProjectEdit
{
	public static readonly PropertyInfo<byte[]> TimeStampProperty = RegisterProperty<byte[]>(nameof(TimeStamp));
	[System.ComponentModel.Browsable(false )]
	[System.ComponentModel.EditorBrowsable(EditorBrowsableState.Never )]
	public byte[] TimeStamp
	{
		get => GetProperty(TimeStampProperty);
		set => SetProperty(TimeStampProperty, value);
	}
[...]

PR #4173, opened 2024-08-17
Features/error source generator

luizfbicalho:
Closes: #4160

### DIFF
--- a/Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.Tests/Snapshots/GeneratorSnapshotTests.ReadOnlyBaseInterface#InterfaceImplementBusinessPOCO.g.verified.cs
+++ b/Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.Tests/Snapshots/GeneratorSnapshotTests.ReadOnlyBaseInterface#InterfaceImplementBusinessPOCO.g.verified.cs
@@ -3,6 +3,7 @@
 #nullable enable
 using System;
 using Csla;
+using System.ComponentModel.DataAnnotations;
 
 namespace Csla.Generator.AutoImplementProperties.TestObjects;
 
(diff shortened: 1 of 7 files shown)


Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.Tests/Snapshots/GeneratorSnapshotTests.ReadOnlyBaseInterface#InterfaceImplementBusinessPOCO.g.verified.cs
```diff
@@ -3,6 +3,7 @@
 #nullable enable
 using System;
```
rockfordlhotka, line 4: Should this file be in source control? It is generated?

rockfordlhotka (changes requested):

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.Tests/Snapshots/GeneratorSnapshotTests.ReadOnlyBaseInterface#InterfaceImplementBusinessPOCO.g.verified.cs
luizfbicalho: in this video https://www.youtube.com/watch?v=azJm_Y2nbAI the author says that is a good practice to put the nullable enable in teh begining

rockfordlhotka: I agree with the `#nullable enable` - I am wondering if the `.g.verified.cs` files should be in source control at all - if they are generated on each build they should not be in source control.

luizfbicalho: this is part of the snapshot unit test, when the test runs it makes a .received.cs and compares both files

the verivied is there to check the test result

rockfordlhotka (approved):

Issue #4130, opened 2024-07-31
Remove incorrect code from UseLocalProxy


rockfordlhotka:
Remove these lines from the `UseLocalProxy` method:

```csharp
      var existingOptions = config.Services.FirstOrDefault(i => i.ServiceType.Equals(typeof(IDataPortalProxy)));
      if (existingOptions?.ImplementationInstance is not LocalProxyOptions proxyOptions)
```

              Please also take a look at `Csla.DataPortalClient.LocalProxyExtensions.UseLocalProxy(config, options)`. That implementation looks broken to me 🤔

_Originally posted by @StefanOssendorf in https://github.com/MarimerLLC/csla/issues/1233#issuecomment-2260753703_

luizfbicalho:
> Please also take a look at `Csla.DataPortalClient.LocalProxyExtensions.UseLocalProxy(config, options)`. That implementation looks broken to me 🤔

Can you add more infromation on this looks broken method? what is going on? How can I reproduce the problem?

StefanOssendorf:
https://github.com/MarimerLLC/csla/blob/7d964dfba833ee33d2ef7049535191d6fd93bfcd/Source/Csla/DataPortalClient/LocalProxyExtensions.cs#L34-L43

The problem is line 36 retrieves a service registration of type `IDataPortalProxy` which represents any kind of `DataPortal` implementation. That means line 37 is _always_ `true`. So both lines are unnecessary an can be removed so only the instantiation of the `LocalProxyOptions` remain with the rest of the method.
Hth

luizfbicalho:
I understood that there were two separated problems, remove the two lines and then after check for the broken part, can you assign that to me?

rockfordlhotka:
This issue is assigned to you @luizfbicalho

PR #4195, opened 2024-08-29
Simplify LocalProxyOptions initialization in UseLocalProxy

luizfbicalho:
The method `UseLocalProxy` now directly initializes a new `LocalProxyOptions` instance without checking for an existing instance. The conditional logic that checked for an existing `IDataPortalProxy` service and its implementation instance has been removed. This change simplifies the method by ensuring that a new `LocalProxyOptions` instance is always created and used.

Closes: #4130


StefanOssendorf (approved):
LGTM

PR #4212, opened 2024-09-07
Hardening against thread aborts when on < .NET 5.0

MarkCiliaVincenti:


Source/Csla.Analyzers/Csla.Analyzers/Csla.Analyzers.csproj
```diff
@@ -16,6 +16,7 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.10.0" PrivateAssets="all" />
     <PackageReference Update="NETStandard.Library" PrivateAssets="all" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All" />
+    <PackageReference Include="Backport.System.Threading.Lock" Version="2.0.0" />
```
rockfordlhotka, line 19: I am hesitant to take on external dependencies.

Nothing personal, just an observation over the decades: when I've taken external (non-Microsoft) dependencies in the past, those dependencies don't live as long as CSLA, and then we have to remove them when they become non-maintained.

Can you explain the value of the `Backport` library vs the built-in behaviors? It _seems_ like backing this out would be easy enough if that became necessary in the future - I just would like to understand the value of using the library in the meantime.

rockfordlhotka (changes requested):

Source/Csla.Analyzers/Csla.Analyzers/Csla.Analyzers.csproj
MarkCiliaVincenti: Hi,

csla is already depending on the same library. It's also small enough that changes are not envisaged unless System.Threading.Lock changes one day in future or if the interface for lockers gets approved (there's a proposal by Stephen Toub).

The only change you'd need is that if you start supporting net5 or later only then you'd be able to change to the simpler (and current) implementation. Unfortunately this is not possible at the moment due to very rare edge cases with thread aborts that could theoretically keep something locked.

In any case, my intention is to keep maintaining this library when needed, as I do with all my other libraries.

Issue #4207, opened 2024-09-06
Cache `NonSerializedAttribute` in new `IsSerializable` property


StefanOssendorf:
https://github.com/MarimerLLC/csla/blob/8fa7d4898ecf43e29a333dd83c0650446b49e183/Source/Csla/PropertyInfo.cs#L218-L219

This call can and should be cached to avoid re-evaluating the attribute on each usage.
The easiest solution would probably be just to assign the result to `_isSerializable` which would shortcut the whole evaluation route.

luizfbicalho:
It's the same structure in the FriendlyName property, and it's not cached also. I can't see the difference.
Should both be cached?

The code is cached in the FieldData.IsSerializable property.

one alternative is to use like in this code

```cs
    protected static PropertyInfo<P> RegisterProperty<P>(string propertyName, RelationshipTypes relationship,bool isSerializable)
    {
      return RegisterProperty(Core.FieldManager.PropertyInfoFactory.Factory.Create<P>(typeof(T), propertyName, string.Empty, relationship,isSerializable));
    }
```

rockfordlhotka:
FriendlyName can't be cached due to localization.

PR #4213, opened 2024-09-07
Refactor PropertyInfo to cache IsSerializable result

luizfbicalho:
Changed _isSerializable from readonly to regular field to allow caching of the IsSerializable result. Updated IsSerializable property getter to set _isSerializable based on the presence of Csla.NonSerializedAttribute, improving performance by avoiding repeated attribute lookups.

Closes: #4207


StefanOssendorf (approved):
LGTM

Issue #4182, opened 2024-08-22
CSLA 8 uses wrong IContextManager in aspnetcore when Blazor Server is also used


rockfordlhotka:
If you have a Blazor 8 app that does the following:

1. Is configured to use CSLA state management, so state is kept in sync between server and wasm, including user identity
2. Is configured to _not_ flow the user from the wasm client to the data portal server (this is the default)
3. Makes a data portal call from a wasm client to the aspnetcore/Blazor server
4. The data portal _on the server_ makes another root data portal call (such as a UoW object)
5. That second data portal call _should_ be using the aspnetcore `IContextManager`, but it continues to use the state-aware Blazor context manager
6. This fails, because the data portal attempts to call `SetUser`, thinking it is running in aspnetcore, but having the wrong context manager (Blazor server) which doesn't support setting the user

The fix is to ensure that, with a cascading root data portal call like this, that the correct context manager is used.

Probably related to #4075

PR #4234, opened 2024-09-27
Explore possible perf issues with static property loading

rockfordlhotka:
* Fix bug: If !SyncContextWithServer properly load empty session
* #4182 Fix Blazor context manager issue


Source/Csla.AspNetCore/Blazor/ApplicationContextManagerBlazor.cs
```diff
@@ -137,19 +141,33 @@ public IPrincipal GetUser()
     /// Not supported in Blazor.
     /// </summary>
     /// <param name="principal">Principal object.</param>
-    public virtual void SetUser(IPrincipal principal) => throw new NotSupportedException(nameof(SetUser));
+    public virtual void SetUser(IPrincipal principal)
+    {
+      ArgumentNullException.ThrowIfNull(principal);
```
StefanOssendorf, line 146: I suggest to add the following after the `null` check:
```csharp
if(principal is not ClaimsPrincipal claimsPrincipal)
{
  throw new NotSupportedException("The user to be set must be of type ClaimsPrincipal");
}
```
That would make a more actionable exception than an invalid cast in line 149 :)

rockfordlhotka: I did as you suggested.

PR #4282, opened 2024-11-04
Minimize DynamicBindingList "async void" methods

### LINKED ISSUE #4281: DynamicListBase SaveItem shouldn't be async void
https://github.com/MarimerLLC/csla/blob/6789bb52ce9a55130e961896a80641220f467e27/Source/Csla/DynamicListBase.cs#L266-L279
IMHO this is bad and a bug. I don't know if this base class is really in use.
`async void` for non event handler should be banned ^^'. If I read this right this will not bubble up any exceptions to the user during a save. 

Edit:
Here too: https://github.com/MarimerLLC/csla/blob/6789bb52ce9a55130e961896a80641220f467e27/Source/Csla/DynamicListBase.cs#L346

_Originally posted by @StefanOssendorf in https://github.com/MarimerLLC/csla/issues/1233#issuecomment-2452394293_

rockfordlhotka:
Closes #4281

### DIFF
--- a/Source/Csla/BusinessBindingListBase.cs
+++ b/Source/Csla/BusinessBindingListBase.cs
@@ -264,9 +264,10 @@ namespace Csla
       AcceptChanges(EditLevel - 1);
     }
 
-    void IParent.ApplyEditChild(IEditableBusinessObject child)
+    Task IParent.ApplyEditChild(IEditableBusinessObject child)
     {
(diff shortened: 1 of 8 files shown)


rockfordlhotka:
There's one `async void` we can't avoid because the .NET base type has it as a protected method, otherwise this change eliminates the rest of them.

Issue #3347, opened 2023-05-19
Add RequiresDynamicCode attribute to dynamic methods/members


rockfordlhotka:
Generally speaking, CSLA will not work with ahead-of-time compilation (AOT) because it makes use of unbounded generic types, reflection, expression trees, and other features that are incompatible with AOT.

To be a good citizen, we should mark methods that build expressions, use reflection, etc. with the `RequiresDynamicCode` attribute.

https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/fixing-warnings

Freelancingonupwork:
I am working on this.

rockfordlhotka:
@Freelancingonupwork let us know (here or in discord) if you have questions. I suspect this issue will require some research and experimentation - at least it would for me, as I haven't explored the AOT stuff in .NET yet.

The core of the issue here, is that AOT will "optimize" code that it thinks isn't being called, so the compiler literally removes the code from the compiled output. My understanding is that there are attributes that we can apply to the code to tell AOT that the code should be kept as part of the compiled output, even if it would normally have been optimized away.

I kind of suspect this will also affect business classes. For example, all the data portal operation methods (like `[Create] private void Create()` will appear to not be invoked by any other code, and so I am pretty sure AOT will optimize away all those methods. Obviously that's a problem, and so business developers will probably need to put an attribute on the data portal operation methods to keep them as part of the compiled output.

kant2002:
Best process in my experience how to annotate project for AOT is following
- Enable AOT analyzer
- Have WarningsAsErrors enabled (not nescessary, but nag you if you miss things)
- List all AOT warnings in the NoWarn property.

Now after recording all work, you diligently find most trivial warnings which you can completely eliminate from the list. 
once done, remove that warning from the NoWarn and nevel let it appear again.

`RequiresDynamicCode` should be probably last resort, there usually possibility to solve warnings based on context. Maybe it require some refactoring, maybe some shaking of code.

Example of how legacy codebases did that https://github.com/dotnet/SqlClient/pull/1959/files#diff-844a96fe684c7aecaa27e22d292e57d4418d2c9c591b62dcf4b0a4f90d918bb1R22
They are SUPER SLOW 😄 unfortunately.

rockfordlhotka:
@kant2002 any progress on this effort?

PR #4340, opened 2024-11-24
Enable checking of the AOT and trimming compatibility.

kant2002:
This is only enable checking, but disable all warnings. Subsequent work should remove warnings from the list inside project file, and fixing source code. I envision at least 2 subsequent PR for fixing this. One with safe changes, and one in unsafe changes. It's possible that not everything is rosy in "safe" changes, and fixing some warnings should be postponed.

Also given examples of what considered safe NativeAOT annotations.

Related #3347


Source/Csla/Core/MobileBindingList.cs
```diff
@@ -24,7 +27,11 @@ namespace Csla.Core
   [System.Diagnostics.DebuggerStepThrough]
 #endif
   [Serializable]
-  public class MobileBindingList<T> : BindingList<T>, IMobileList
+  public class MobileBindingList<
+#if NET8_0_OR_GREATER
+    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)]
```
kant2002, line 32: This requirement comes from `System.ComponentModel.BindingList<T>`.just have to propagate it.

Source/Csla/Core/ExtendedBindingList.cs
```diff
@@ -20,7 +23,11 @@ namespace Csla.Core
   /// </summary>
   /// <typeparam name="T">Type of item contained in list.</typeparam>
   [Serializable]
-  public class ExtendedBindingList<T> : MobileBindingList<T>,
+  public class ExtendedBindingList<
+#if NET8_0_OR_GREATER
+    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)]
```
kant2002, line 28: This requirement appears because it appears on `MobileBindingList<T>`.

Source/Csla/Csla.csproj
```diff
@@ -12,6 +12,16 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>CslaKey.snk</AssemblyOriginatorKeyFile>
     <BaseOutputPath>..\..\Bin</BaseOutputPath>
+    <EnableTrimAnalyzer Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</EnableTrimAnalyzer>
```
StefanOssendorf, line 15: You are checking against net7.0 here. But we don't support net7.0 explicitly anymore. We are only building against 6,8 and 9 iirc

StefanOssendorf: Only net8 and 9

kant2002: Okay. I’m probably overthinking things

PR #4352, opened 2024-11-28
Add statically analyzeable AOT annotations

kant2002:
Other things which cannot be automatically analyzed by the tools, would be dealt later on. This is gives baseline on thinking about code, and can be used for future improvements. Annotation using `DynamicallyAccessedMembers` that's what should be done immdiately after capturing existing warnings.

Given cascading nature of the changes, it's better keep track of class of warnings, and don't allow re-introduce them after you cleanup codebase.

Related to #3347

### DIFF
--- a/Source/Csla/Utilities.cs
+++ b/Source/Csla/Utilities.cs
@@ -10,6 +10,7 @@ using System.Reflection;
 using Csla.Reflection;
 using System.ComponentModel;
 using Csla.Properties;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Csla
 {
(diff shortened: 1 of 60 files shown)


Source/Csla/Utilities.cs
```diff
@@ -150,7 +151,15 @@ public static Type GetChildItemType(Type listType)
     /// result is parsed to convert into the enum value.
     /// </para>
     /// </remarks>
-    public static object CoerceValue(Type desiredType, Type valueType, object oldValue, object value)
+    public static object CoerceValue(
```
kant2002, line 154: That's the method which should be very pedantically analyzed. If I can prove statically, that `DAM(All)` can be replaced with `DAM(PublicProperties | Interfaces)` for example, they it will significantly remove requirements from the caller code, no need to include everything from types then.

rockfordlhotka: If I understand this correctly, the fact that we have things like `private async Task FetchAsync()` that need to avoid optimization, the changes you've made referencing `All` is probably correct?

kant2002: The changes which I made provable correct by compiler. Basically compiler tell me this method uses other method which require this and that and I annotate method with compiler warnings. Rinse and repeat.

I was only hoping that this thing can be written in different way. Or there alternative mode which is more AOT friendly and without this code.
https://github.com/MarimerLLC/csla/blob/93f1670c09cd4846afd3bd5f988f175bd3e8a7a6/Source/Csla/Utilities.cs#L216-L221

I know that would be breaking change. I just indicate that this place alone require all types to preserve all their code which is not necessary optimal. If I drop this code hypothetically, then 80% of annotations would be completely different. You unlikely can guess this afterwards.

rockfordlhotka (approved):

Issue #4354, opened 2024-11-29
Wrong IContextManager is used in ASP.NET 5 with LocalProxy


rockfordlhotka:
When using CSLA 8+ with an ASP.NET MVC 5 (and presumably all ASP.NET 5) projects, and a LocalProxy for the data portal, the logical server-side data portal attempts to use `Csla.Web.Mvc.ApplicationContextManager`. The problem is that `HttpContext` is not available in logical server-side code, and so the data portal call fails.

The `Csla.Web.Mvc.ApplicationContextManager` knows it is invalid - but it gets used anyway.

It seems that `ApplicationContextAccessor` is returning the invalid context manager rather than a valid option, because there is no check for this case in the code.

PR #4356, opened 2024-11-29
Ensure context manager is valid before returning

rockfordlhotka:
#4354


StefanOssendorf:
Is it possible to add a test for that?

rockfordlhotka:
> Is it possible to add a test for that?

I did add one for CSLA 9 - had to create a whole new test project to reference net48.

#4361

PR #4361, opened 2024-12-04
Use correct context manager for aspnet classic

rockfordlhotka:
#4354


rockfordlhotka:
I am seriously wondering if a unit test for .NET 4.8 is worth the pain of getting the CI server to actually build old mstest project targets like this one 😠

PR #4363, opened 2024-12-05
More AOT annotations to the code from IL2072

kant2002:
I also add MSBuild file which simplify testing with local Csla for samples. Just change condition locally, and dotnet build and other commands will pickup local Csla

Related #3347


rockfordlhotka:
Thank you @kant2002 - this will have to wait for version 9.1 (or maybe 10, as it has breaking changes?).

Version 9 is about to go live, and this is too big to adopt today.

rockfordlhotka:
> as it has breaking changes?

On further review of the code, maybe there are no breaking changes to the public API?

I don't understand some of the new code though - like changing `Type` to `new(type)`? What am I missing?

kant2002:
You may take a look at this issue https://github.com/dotnet/runtime/issues/110239 it shows technique which required to mark Type inside list with attributes

Issue #4315, opened 2024-11-13
Backport IContextManager selection fixes from v8 to v7


rockfordlhotka:
The `IContextManager` changes/fixes are spread across a number of issues and PRs. These are the latest:

* https://github.com/MarimerLLC/csla/issues/4208
* https://github.com/MarimerLLC/csla/pull/4239

StefanOssendorf:
Do we still need/want this with v9 around the corner? An update to V8 from v7 should be easy.

rockfordlhotka:
> Do we still need/want this with v9 around the corner? An update to V8 from v7 should be easy.

@tkyle is working with a Xamarin project and the last version where we support Xamarin is CSLA 7. I think we need to allow folks who are invested in that platform to continue to provide PRs so they can be successful.

StefanOssendorf:
Oh yeah. I forgot. Nvm then

rockfordlhotka:
We are no longer making changes to v7.

PR #4395, opened 2024-12-16
#4315 ContextManager Backport

tkyle:
#4315 

Backported the context manager changes from 8 down to 7.


rockfordlhotka (changes requested):
Because there are some breaking changes in this PR I'll make the version go to 7.1.0 instead of 7.0.7.

Can you please edit the `/csla/releasenotes.md` file and provide some guidance to a future dev as to what are the breaking changes and what they need to change in their `AddCsla` config to move from 7.0.6 to 7.1.0?

Otherwise I'm good with the changes in the PR - totally impressed by how much work you put in!

rockfordlhotka:
Not needed as per discussion in Discord

PR #4405, opened 2024-12-17
Rename Csla.NonSerializedAttribute to NonSerializedPropertyAttribute

### LINKED ISSUE #4387: Rename Csla.NonSerializedAttribute to avoid ambiguous type issue
Hi. I've just updated our CSLA from 8 to the 9-RC1.
I encountered the following issues/compilation errors:
* The new `Csla.NonSerializedAttribute` is now ambigious with the `System.NonSerializedAttribute`. Which poses the question: Which do I need to mark the field as not participating in the serialization process? (It was "introduced" in #4197 )

I'll update this post when I encounter other problems.

_Originally posted by @ossendorf-at-hoelscher in https://github.com/MarimerLLC/csla/discussions/4383#discussioncomment-11568557_

rockfordlhotka:
Closes #4387 Rename Csla.NonSerializedAttribute to NonSerializedPropertyAttribute


ossendorf-at-hoelscher:
That attribute applies to properties and fields. So `Property` in the name might be misleading.
So instead
* `NonSerializedDataAttribute`
* `NonCslaSerialized`

rockfordlhotka:
Some AI generated attribute name suggestions

image

StefanOssendorf:
I think we should go with `DoNotSerializeAttribute`.

PR #4415, opened 2024-12-23
Enable resetting application context user to null

### LINKED ISSUE #4410: v9-RC1: FlowSecurityPrincipalFromClient = true, throws ArgumentNullException
https://github.com/MarimerLLC/csla/blob/cf867fbe1a90ecf17e0a60e2227d013b409a7b5f/Source/Csla/Server/DataPortal.cs#L776-L777

DataPortal throws an exception when `FlowSecurityPrincipalFromClient=true` when calling without an actual user set.
We either have to set it to an default unauthenticated user or make the code handle `null`.

StefanOssendorf:
Set application context user to an unauthenticated claims principal when `FlowSecurityPrincipalFromClient` is enabled.

Fixes #4410

### DIFF
--- a/Source/Csla/Server/DataPortal.cs
+++ b/Source/Csla/Server/DataPortal.cs
@@ -7,6 +7,7 @@
 //-----------------------------------------------------------------------
 
 using System.Diagnostics.CodeAnalysis;
+using System.Security.Claims;
 using System.Security.Principal;
 using Csla.Configuration;
 using Csla.Properties;
@@ -774,7 +775,7 @@ namespace Csla.Server
       if (!context.IsRemotePortal) return;
       _applicationContext.Clear();
       if (SecurityOptions.FlowSecurityPrincipalFromClient)
-        _applicationContext.User = null;
+        _applicationContext.User = new ClaimsPrincipal(new ClaimsIdentity());
     }
 
     #endregion
--- a/Source/Csla/Server/DataPortal.cs
+++ b/Source/Csla/Server/DataPortal.cs
@@ -775,7 +775,7 @@ namespace Csla.Server
       if (!context.IsRemotePortal) return;
       _applicationContext.Clear();
       if (SecurityOptions.FlowSecurityPrincipalFromClient)
-        _applicationContext.User = null;
+        _applicationContext.User = new ClaimsPrincipal(new ClaimsIdentity());
     }
 
     #endregion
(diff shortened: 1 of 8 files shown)


Source/Csla/Server/DataPortal.cs
```diff
@@ -774,7 +775,7 @@ private void ClearContext(DataPortalContext context)
       if (!context.IsRemotePortal) return;
       _applicationContext.Clear();
       if (SecurityOptions.FlowSecurityPrincipalFromClient)
-        _applicationContext.User = null;
+        _applicationContext.User = new ClaimsPrincipal(new ClaimsIdentity());
```
rockfordlhotka, line 778: You can just create the new principal - it defaults to creating a new identity if you don't provide one.

rockfordlhotka (changes requested):
This is in the clear context method - which is there to ensure nothing is left on the server once the data portal operation is complete.

I'm wondering why it _matters_ if the value is left as a null, because nobody should be using it anyway - we're just cleaning up at this point, immediately before leaving the server to return to the client.

StefanOssendorf:
> This is in the clear context method - which is there to ensure nothing is left on the server once the data portal operation is complete.
> 
> I'm wondering why it _matters_ if the value is left as a null, because nobody should be using it anyway - we're just cleaning up at this point, immediately before leaving the server to return to the client.

It matters because we have implementations which don't accept null as a new principal on the application context.

Source/Csla/Server/DataPortal.cs
StefanOssendorf: Now obsolete. I set it to `null` again.

StefanOssendorf:
Converted the PR to draft.
I'm adding tests to ensure the context managers can handle `null` as a value.
But the `InMemory` Contextmanager can't handle that case. Only a `ClaimsPrincipal`. I think we should default to an empty/unauthenticated `ClaimsPrincipal` to reduce the error cases.

Source/Csla/Server/DataPortal.cs
StefanOssendorf: That's not true. Using the default ctor the principal does not have an identity associated with it.
See https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Security.Claims/src/System/Security/Claims/ClaimsPrincipal.cs#L99-L104

rockfordlhotka (approved):

StefanOssendorf:
So. I kept the `User = null;` to make it less impactful on our users. The `ApplicationContext` now handles the `null` and translates it into a `ClaimsPrincipal(PrincipalIdentity)`.

rockfordlhotka (approved):

PR #4437, opened 2025-01-10
Possible issue with exceptions in an async business rule

keithdv:
Hello,

Is there the possibility of "lost" exceptions with async business rules? I've made the changes below and the unit tests still pass. The 'catch' at line 1124  is skipped for async Rule.Execute since the task is not awaited. Should the exception be hung on to and throw if/when they await the rules? Should there be context.AddErrorResult with the exception message?

Thanks!
Keith


rockfordlhotka:
Thank you for the ideas @keithdv. Unfortunately the PR doesn't build, which is a blocker for evaluation.

Issue #4450, opened 2025-01-21
Fix issues with Backport.System.Threading.Lock package


rockfordlhotka:
https://github.com/MarkCiliaVincenti/Backport.System.Threading.Lock/issues/6

It looks like we either 

1. Go through all the projects and fix all the references/usage of the library
   * Remove the offending lines from `Build.Build.props`
   * For projects that actually use the library, keep it and add necessary lines to csproj
   * For projects that don't use the library, but just reference it today due to the viral mistake, remove the reference entirely
3. Go through all the projects and remove all usage of the library

Right now I'm leaning toward option 2, as I'm not convinced the value of the library is worth the ongoing effort.

StefanOssendorf:
If we go route 2. How will we handle the locks?
The new `Lock` object does seem to have performance improvements we should not miss out - I think.

rockfordlhotka:
> If we go route 2. How will we handle the locks? The new `Lock` object does seem to have performance improvements we should not miss out - I think.

That is a good point, though I'd like to see some hard data about how much different the performance improvements actually make.

I must confess that I put in option 2 due to frustration, as the `Backport` package dependency seems to cause random build issues over time and that is frustrating when it happens while someone is trying to work on an actual CSLA issue.

StefanOssendorf:
I understand that :)
Here's a post from the .Net blog: https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-9/#threading
It's "only" a difference of ~0.5ns but I think with this kind of stuff it stacks up when used in hot path scenarios.
I don't have an overview when or where we use lock extensivly. So if it's not used in hot path scenarios we could stick to the old style for now.

Bowman74:
As an interesting aside, from what I can tell reading the background on System.Threading.Lock, the motivation was to give developers more control over the locking process and not necessarily performance. Looks like MS got an extra bonus with the new design. As with all new language features, their use can be problematic in libraries like CSLA which still target and compile against earlier versions.

I guess it is possible to avoid a dependency is for CSLA to create its own Locking abstraction class that contains all the required `#if NET9_0_OR_GREATER` type code to handle differences in the locking syntax when compiled in different versions of dot net. I don't know if it is a good idea or a bad idea, but it is an idea that I believe has been used before.

StefanOssendorf:
This is an old issue but I think it's the right place. So I'll reopen it until adviced to create a new issue.

I'm reopening because during analyzing the build time of our solution at work I found out:
[...]

PR #4473, opened 2025-02-08
Addresses #4450 (Fix issues with Backport.System.Threading.Lock package)

MarkCiliaVincenti:
Addresses #4450

### DIFF
--- a/Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.Attributes.CSharp/Csla.Generator.AutoImplementProperties.Attributes.CSharp.csproj
+++ b/Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.Attributes.CSharp/Csla.Generator.AutoImplementProperties.Attributes.CSharp.csproj
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Backport.System.Threading.Lock" Version="3.1.0" />
+    <PackageReference Include="Backport.System.Threading.Lock" Version="3.1.4" />
     <Using Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))" Alias="Lock" Include="System.Threading.Lock" />
     <Using Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))" Alias="Lock" Include="Backport.System.Threading.Lock" />
     <Using Alias="LockFactory" Include="Backport.System.Threading.LockFactory" />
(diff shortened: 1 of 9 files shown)


Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.Attributes.CSharp/Csla.Generator.AutoImplementProperties.Attributes.CSharp.csproj
```diff
@@ -18,6 +18,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Backport.System.Threading.Lock" Version="3.1.0" />
+    <PackageReference Include="Backport.System.Threading.Lock" Version="3.1.4" />
+    <Using Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))" Alias="Lock" Include="System.Threading.Lock" />
+    <Using Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))" Alias="Lock" Include="Backport.System.Threading.Lock" />
```
CptWesley, line 23: At this point, wouldn't it make more sense to make the `Lock` alias simply refer to `System.Object`? This is assuming that the only way the `Lock` type is currently being used is through the `lock (...)` syntax and not through any method calls. That way the external dependency can be dropped in its entirety.

MarkCiliaVincenti: That would not allow the new locking API though. If an external dependency is an issue it can be used as the source generator instead.

rockfordlhotka (commented):
Does this change mean that when people create a .NET 4.8 WinForms project, they won't need to add their own reference to the `BackPort` package?

I found that I had to add this superfluous reference when updating the samples.

For example, [WinSortFilter](https://github.com/MarimerLLC/csla/tree/main/Samples/WinSortFilter)

rockfordlhotka:
If I read this right, the only projects that should have a reference to `Backport` are `Csla` and `Csla.Channels.RabbitMq`, is that right?

All other csproj files with any mention of the package should have it removed?

image

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.Attributes.CSharp/Csla.Generator.AutoImplementProperties.Attributes.CSharp.csproj
rockfordlhotka: See my other comment regarding usage of the `Backport` package - it is only used in a couple projects, and so no other projects (including the generators) should reference the package at all right?

PR #4474, opened 2025-02-11
Properly serialize data in a FieldData object

### LINKED ISSUE #4472: Using MobileFormatter Custom Serializers throws Exception on BeginEdit
**Describe the bug**
I want to use a MobileFormatter Custom Serializers as described in [Upgrading to CSLA 9](https://github.com/MarimerLLC/csla/blob/main/docs/Upgrading%20to%20CSLA%209.md).

When using the BeginEdit() methode of the businessobject, an System.NotSupportedException is thrown with the message "Serialization of type object is not supported".

**Version and Platform**
CSLA version: 9.0.0-rc3
OS: Windows 10
Platform: all

**Code that Fails**
[https://github.com/jensbrand/CslaCustomSerializerTest](https://github.com/jensbrand/CslaCustomSerializerTest)

   at Csla.Serialization.Mobile.CslaBinaryWriter.Write(Object target, BinaryWriter writer)
   at Csla.Serialization.Mobile.CslaBinaryWriter.Write(Stream serializationStream, List`1 objectData)
   at Csla.Serialization.Mobile.MobileFormatter.Serialize(Stream serializationStream, Object graph)
   at Csla.Core.FieldManager.FieldDataManager.Csla.Core.IUndoableObject.CopyState(Int32 parentEditLevel, Boolean parentBindingEdit)
   at Csla.Core.UndoableBase.CopyState(Int32 parentEditLevel)
   at Csla.Core.BusinessBase.BeginEdit()
   at CslaTestCustomSerializer.Main.ExecuteAsync(CancellationToken stoppingToken) in D:\CSharp-Projekte\Test\CslaTestCustomSerializer\CslaTestCustomSerializer\Program.cs:line 51

**Additional context**
I tried to debug the issue and I think the problem is caused by the serialization of the FielddataManager. This class uses the MobileFormatter to serialize the list of fields (IFieldData) and the implementation FieldData stores the value of the field as it is:

```csharp
        info.AddValue("_data", _data);
```

Later then, this value is serialized with CslaBinaryWriter and this class throws the error.

Unfortunately I am not familiar enough with the subject to find a good solution for this

rockfordlhotka:
Fixes #4472

### DIFF
--- a/Source/Csla.test/Serialization/CustomSerializerTests.cs
+++ b/Source/Csla.test/Serialization/CustomSerializerTests.cs
@@ -110,6 +110,63 @@ namespace Csla.Test.Serialization
       var clone = (SerializablePoco)cloner.Clone(poco);
       Assert.AreEqual(poco.Name, clone.Name);
     }
+
+    [TestMethod]
+    public async Task ClonePoco()
+    {
+      var services = new ServiceCollection();
+      services.AddCsla(o => o
+        .Serialization(o => o
+          .UseMobileFormatter(o => o
+            .CustomSerializers.Add(new TypeMap<object, PocoSerializer<SerializablePoco>>(PocoSerializer<SerializablePoco>.CanSerialize)))));
+      services.AddScoped<Csla.Core.IContextManager, Csla.Core.ApplicationContextManagerAsyncLocal>();
+      var provider = services.BuildServiceProvider();
+      var applicationContext = provider.GetRequiredService<ApplicationContext>();
+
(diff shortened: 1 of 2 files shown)


rockfordlhotka:
@jensbrand I didn't (and still don't) see a comment in the PR prior to this one from 2 hours ago.

That is a good catch and I'll address that.

jensbrand:
That is very strange, I posted the following comment yesterday

grafik

PR #4501, opened 2025-02-16
fix incorrect as TypeDeclarationSyntax usage

SimonCropp:
since it has to be a TypeDeclarationSyntax  or it would be a null ref in the following lines


rockfordlhotka:
Thank you for the wonderful flurry of PRs @SimonCropp !

Can you please resolve this conflict?

rockfordlhotka (approved):

PR #4507, opened 2025-02-16
avoid string naming of members

SimonCropp:


SimonCropp:
hmm i think i broke something here. i will take another look

SimonCropp:
good to go

PR #4526, opened 2025-02-17
remove redundant base calls

SimonCropp:


rockfordlhotka (approved):

rockfordlhotka:
It is amazing some of the cruft that makes its way into a codebase...

PR #4544, opened 2025-02-20
cleanup BindingSourceHelper

SimonCropp:
* move to file scoped namespace
 * use enumerable for GetChildBindingSources instead of assignable lists for each level in the hierarchy 
 * use pattern matching in GetChildBindingSources


rockfordlhotka:
@SimonCropp there aren't really unit tests for the Windows UI components. Can you confirm that this change doesn't break existing functionality?

PR #4554, opened 2025-02-20
use type equality instead of .Equals

SimonCropp:


StefanOssendorf:
Out of curiosity: What's the benefit using one over the other?

hurcane:
I see that all of these changes involve System.Type comparisons. If I am remembering correctly, the == operator is stricter than the Equals method. The == operator does reference equality. The Equals method returns true if the underlying types are the same. System.Type.Equals will return true on types that are different but inherit from the same root.

SimonCropp:
> Out of curiosity: What's the benefit using one over the other?

mainly consistency and readability. 

> If I am remembering correctly, the == operator is stricter than the Equals method. The == operator does reference equality. The Equals method returns true if the underlying types are the same. System.Type.Equals will return true on types that are different but inherit from the same root.

nope. that is what IsAssignableTo and IsAssignableFrom are for

https://learn.microsoft.com/en-us/dotnet/api/system.type.isassignableto?view=net-9.0#system-type-isassignableto(system-type)
https://learn.microsoft.com/en-us/dotnet/api/system.type.isassignablefrom?view=net-9.0#system-type-isassignablefrom(system-type)

SimonCropp:
@StefanOssendorf  want me to handle the conflicts?

StefanOssendorf:
> @StefanOssendorf want me to handle the conflicts?

I don't know. I think we should wait for this one for @rockfordlhotka .

PR #4556, opened 2025-02-20
Add some solution level usings

SimonCropp:
needs a squash on merge


SimonCropp:
@rockfordlhotka is this something u want? if so i will handle the conflicts

PR #4557, opened 2025-02-20
use some null coalescing

SimonCropp:


SimonCropp:
@rockfordlhotka is this something you want? if so i will handle the conflicts

Issue #4565, opened 2025-03-02
Abstracting the data persistence of the Blazor state management session storage


joshhanson314:
The Csla.Blazor.State.SessionManager uses a concurrent dictionary as a build in session store. In a load balanced environment, multiple instances of a Blazor server application using the state management subsystem can each only access sessions stored in their own memory. This causes failures to look up and update these sessions for a client's requests that may be handled by multiple instances of the application.

The storage mechanism can be abstracted to use any implementation of data persistence. The default implementation can continue to be an in memory session store that still implements a concurrent dictionary. Extensibility points can be provided to override this storage abstraction rather than needing to override the entire logic of the ISessionManager implementation.

(This was discussed in the Discord channel on 1/22/25)

joshhanson314:
I have created a fork and branch to contribute this feature.

luizfbicalho:
Is this in development or abandoned?

joshhanson314:
I have been focused with other issues, but I can reengage with it and complete it. Give me a week or 2 please.

PR #4567, opened 2025-03-02
Abstract out session store from SessionManager

joshhanson314:
#4565 Added abstractions and a default implementation for a session store to be used with the Blazor state SessionManager. The default implementation continues to use in-memory persistence in a ConcurrentDictionary.

### DIFF
--- a/Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
+++ b/Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
@@ -6,7 +6,6 @@
 // <summary>Manages all user session data</summary>
 //-----------------------------------------------------------------------
 
-using System.Collections.Concurrent;
 using Csla.State;
 
 namespace Csla.Blazor.State
@@ -16,23 +15,30 @@ namespace Csla.Blazor.State
   /// root DI container.
   /// </summary>
   /// <param name="sessionIdManager"></param>
-  public class SessionManager(ISessionIdManager sessionIdManager) : ISessionManager
+  /// <param name="sessionStore"></param>
+  public class SessionManager(ISessionIdManager sessionIdManager, ISessionStore sessionStore) : ISessionManager
   {
-    private readonly ConcurrentDictionary<string, Session> _sessions = [];
     private readonly ISessionIdManager _sessionIdManager = sessionIdManager;
+    private readonly ISessionStore _sessionStore = sessionStore;
 
     /// <summary>
     /// Gets the session data for the current user.
     /// </summary>
     public Session GetSession()
     {
-      Session result;
       var key = _sessionIdManager.GetSessionId();
-      if (!_sessions.ContainsKey(key))
-        _sessions.TryAdd(key, []);
-      result = _sessions[key];
-      result.Touch();
-      return result;
+      var session = _sessionStore.GetSession(key);
+      if (session == null)
+      {
+        session = [];
+        session.Touch();
+        _sessionStore.CreateSession(key, session);
+        return session;
+      }
+
+      session.Touch();
+      _sessionStore.UpdateSession(key, session);
+      return session;
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


Source/Csla/State/ISessionStore.cs
rockfordlhotka, line 1: I think @StefanOssendorf would be much happier if this had async methods instead of sync methods 😁 

Seriously though, if we're going to improve the code like this, I do think we should expect that people will use async stores for the data.

Source/Csla.Blazor/State/DefaultSessionStore.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /// <inheritdoc />
+    public void DeleteSession(string key)
+    {
+      _store.TryRemove(key, out _);
+    }
+
+    /// <inheritdoc />
+    public void DeleteSessions(SessionsFilter filter)
```
rockfordlhotka, line 52: I don't like this method name. Perhaps it should be more explicit that this is _expiring_ sessions?

Source/Csla/State/ISessionStore.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <param name="key"></param>
+    /// <returns>The session for the given key, or default if not found</returns>
+    Session? GetSession(string key);
+
+    /// <summary>
+    /// Creates a session
+    /// </summary>
+    /// <param name="key"></param>
+    /// <param name="session"></param>
+    void CreateSession(string key, Session session);
```
rockfordlhotka, line 29: Could the create/update be combined to a `TryUpdate` or something similar?

Source/Csla/State/SessionsFilter.cs
```diff
@@ -0,0 +1,33 @@
...
+    public TimeSpan? Expiration { get; set; }
+
+    /// <summary>
+    /// Validates
+    /// </summary>
+    public void Validate()
+    {
+      if (!Expiration.HasValue)
+      {
+        throw new ArgumentNullException("Expiration is required.");
```
rockfordlhotka, line 29: No hard-coded use of English is allowed in the framework. This needs to be from a resource string or be `ArgumentException(nameof(HasValue))`

Source/Csla/State/SessionsFilter.cs
rockfordlhotka, line 1: This is basically internal to the default implementation right?

As in, if someone were to use a different session store they'd probably use their own approach for expiring old data.

Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
rockfordlhotka, line 1: I really appreciate this refactoring - absolutely on the right track. While we're doing this, might as well also embrace async.

Perhaps also consider using file scoped namespaces so Simon doesn't need to come back later and clean that up?

rockfordlhotka (changes requested):

Source/Csla/State/SessionsFilter.cs
ossendorf-at-hoelscher: No. Currently this class is part of the public API.
`ISessionStore.DeleteSession(SessionsFilter)`

Source/Csla/State/SessionsFilter.cs
ossendorf-at-hoelscher: My suggestion to get rid of this exception would be: Don't _allow_ the creation of an invalid object state.
1. Make `Expiration` non nullable.
2. Either add a constructor expecting the value and/or make the property `required`
With those changes the whole `Validate()` method becomes obsolete.

Source/Csla/State/ISessionStore.cs
ossendorf-at-hoelscher: Yes please :)
You can always do sync with a  `Task`/`ValueTask` method but not async with a `void` method :)

Source/Csla/State/ISessionStore.cs
ossendorf-at-hoelscher: `AddOrUpdate` like the `ConcurrentDictionary` methods?

Source/Csla.AspNetCore/Blazor/State/SessionManager.cs
```diff
@@ -44,9 +50,19 @@ public void UpdateSession(Session newSession)
     {
       ArgumentNullException.ThrowIfNull(newSession);
       var key = _sessionIdManager.GetSessionId();
-      var existingSession = _sessions[key];
+      var existingSession = _sessionStore.GetSession(key)!;
```
ossendorf-at-hoelscher, line 53: Why do you suppress with `!`? You should check for it and handle the `default` case.

ossendorf-at-hoelscher:
@rockfordlhotka This is a change for v10. It's a breaking one ;)

rockfordlhotka:
Although this is a good change, I think this particular PR is so out of date now that it is beyond hope. If I'm wrong, please reopen and update, otherwise it'll stay closed.

PR #4585, opened 2025-03-08
use Guard for null checks

SimonCropp:


StefanOssendorf:
Hi. Can you please explain the changes here? What's the benefit of the `Guard` and where does it come from?

hurcane:
Simon will need to address the build failures, but I see the benefit as a more concise implementation of boilerplate guard clauses. It comes from the Polyfill package, which I believe is already referenced by Csla for other purposes.

StefanOssendorf:
Yeah I added the PolyFill package for attributes and the like. To get rid of a lot of #if NET.... stuff.
I agree with boilerplate code, but my "problem" with those Guards is. They are not auto generated. I have to type it for every usage. The normal argument null exception stuff is auto generated via vs. So no effort at all.

SimonCropp:
fair enough. will close

SimonCropp:
> I agree with boilerplate code, but my "problem" with those Guards is. They are not auto generated. I have to type it for every usage. The normal argument null exception stuff is auto generated via vs. So no effort at all.

i should address that in more details. while i too prefer not to have the ide write code for me, it is problematic when that code it writes is overly verbose. if you google "time writing code versus time reading code" you can see estimates vary wildly. anywhere from 1/5 to 1/100. but taking any of those figures it is clear that we should give stronger weighting to code impact on readability over the ease of writing code. And (in so many cases) having "what the method does" buried so far under verbose argument checks negatively effects my ability to read and understand the code.

but that is just IMHO

Issue #4530, opened 2025-02-18
Analyzer throws exceptions during build


ossendorf-at-hoelscher:
**Describe the bug**
When building our solution I get a couple of exceptions in the build output logs.
`112>CSC : warning AD0001: Analyzer 'Csla.Analyzers.EvaluatePropertiesForSimplicityAnalyzer' threw an exception of type 'System.ArgumentNullException' with message 'Value cannot be null.`

**Version and Platform**
CSLA version: 9.0
OS: Windows

**Additional context**
Doing a build or rebuild all

ProDInfo:
I have the same error:
```
CSC : warning AD0001: El analizador "Csla.Analyzers.EvaluatePropertiesForSimplicityAnalyzer" produjo una excepción de tipo "System.ArgumentNullException" con el mensaje "Value cannot be null. (Parameter 'classDeclaration')".

```

**Version and Platform**
CSLA version: 9.0
OS: Windows

**Additional context**
Doing a build or rebuild all

JurjendeGroot:
I have the same error

warning AD0001: Analyzer 'Csla.Analyzers.EvaluatePropertiesForSimplicityAnalyzer' threw an exception of type 'System.ArgumentNullException' with message 'Value cannot be null.

Version and Platform
CSLA version: 9.0
OS: Windows

Additional context
Doing a build or rebuild all

ossendorf-at-hoelscher:
I found the issue. It's this new check/validation.
https://github.com/MarimerLLC/csla/blob/f8ad0074d0b0e88111d74259b1f413fcfc00b341/Source/Csla.Analyzers/Csla.Analyzers/EvaluatePropertiesForSimplicityAnalyzer.cs#L249-L257

Instead of throwing an exeption I'll change it to
```csharp
if (classDeclaration is null)
{
  return [];
}
```
With this the analyzer doesnt die.
The actual underlying problem is that it's trying to analyze interface property declaration which don't have a containing class.

PR #4587, opened 2025-03-10
make csla.analyzers nullable aware

StefanOssendorf:
Make Csla.Analyzers nullable aware to avoid possible null reference exceptions.

Maybe this solves #4530


StefanOssendorf:
Did something went wrong? Those changes are not in main 🤔

PR #4594, opened 2025-03-14
Remove redundant comments

SimonCropp:
since they negatively effect readability and add no value


StefanOssendorf:
Why do they negatively affect readability?

I'm personally for full comments here. Especially with public types.
This PR also disables these warnings. So instead of removing we should fill them out correctly.

rockfordlhotka:
While I would prefer completing the comments, I am ok with removing empty comment blocks.

I do NOT want to disable warnings about comments, because comments should exist - especially for public APIs.

I have a local branch where I've been working on automatic creation of docs as part of the build. In the past people have volunteered to create the docs, but that should be part of the GitHub pipeline.

PR #4597, opened 2025-03-14
fix type compare bug in IsNativeType

SimonCropp:
An instance of a Type can never be a IMobileObject. i assume this code should be doing a IsAssignableTo?


StefanOssendorf:
Can you please add a test for this case. Probably won't ever happen/change again. But I think a test to proof it's fixed is better.

SimonCropp:
@StefanOssendorf done. and i added a few other tests for the method as well

StefanOssendorf (approved):
Thanks a lot. lgtm

PR #4603, opened 2025-03-14
Allow disabling scan for data annotations attributes

### LINKED ISSUE #4492: Add option to ignore data annotation attributes

rockfordlhotka:
Closes #4492


StefanOssendorf:
Please fix the merge conflict

Source/Csla.test/Rules/BusinessRuleTests.cs
```diff
@@ -0,0 +1,88 @@
...
+      TestResults.Reinitialise();
+    }
+
+    [TestMethod]
+    public void DefaultDataAnnotationsScan()
+    { 
+      var portal = _testDIContext.ServiceProvider.GetRequiredService<IDataPortal<TestBusinessRule>>();
+      var obj = portal.Create();
+      obj.Name = "";
+      Assert.IsFalse(obj.IsValid);
```
StefanOssendorf, line 39: Shouldn't we use FluentAssertions here?

Source/Csla.test/Rules/BusinessRuleTests.cs
```diff
@@ -0,0 +1,88 @@
...
+    [TestMethod]
+    public void DisableDataAnnotationsScan()
+    {
+      var options = _testDIContext.ServiceProvider.GetRequiredService<CslaOptions>();
+      options.ScanForDataAnnotations(false);
+
+      var portal = _testDIContext.ServiceProvider.GetRequiredService<IDataPortal<TestBusinessRule2>>();
+      var obj = portal.Create();
+      obj.Name = "";
+      Assert.IsTrue(obj.IsValid);
```
StefanOssendorf, line 51: Shouldn't we use FluentAssertions here?

Source/Csla.test/Rules/BusinessRuleTests.cs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    [Create]
+    private void Create()
+    {
+      BusinessRules.CheckRules();
+    }
+  }
+
+  public class TestBusinessRule2 : BusinessBase<TestBusinessRule2>
```
StefanOssendorf, line 72: Why do you need two types for this test? Shouldn't be the first sufficient?

Source/Csla/ApplicationContext.cs
```diff
@@ -26,12 +27,19 @@ public class ApplicationContext
...
     {
       ApplicationContextAccessor = applicationContextAccessor;
       ApplicationContextAccessor.GetContextManager().ApplicationContext = this;
+      Options = options;
     }
 
+    /// <summary>
+    /// Holds configuration options for the CSLA framework.
+    /// </summary>
+    public CslaOptions Options { get; private set; }
```
StefanOssendorf, line 41: Remove `private set`. This should be "readonly" even within the class.

rockfordlhotka:
Oh, I was fixing the merge conflict and now realize that this PR should be against the v9.x branch, not main - that's why that conflict came into being.

I'll close and resubmit the PR.

PR #4605, opened 2025-03-16
remove redundant dispose code from SortedBindingList

SimonCropp:


Source/Csla/SortedBindingList.cs
```diff
@@ -132,42 +129,9 @@ public void Reset()
...
-          {
-            // free unmanaged resources when explicitly called
-          }
-          // free shared unmanaged resources
-        }
-        _disposedValue = true;
-      }
-
-      // this code added to correctly implement the disposable pattern.
       public void Dispose()
```
StefanOssendorf, line 132: Why isn't this method removed, too?

StefanOssendorf (changes requested):

Source/Csla/SortedBindingList.cs
SimonCropp: required by IEnumerator

StefanOssendorf (approved):

PR #4607, opened 2025-03-16
remove redundant dispose code in SafeDataReader

SimonCropp:


StefanOssendorf (changes requested):
This class is public and has to implement Dispose. Why do you remove the default implementation code for IDisposable for classes which can be derived from? 

I don't know if this is actually used in the wild but it's public and someone deriving from this class must have a way to add code to the disposing. 
So I'm against this change.

SimonCropp:
the `Dispose(bool disposing)` and destructor pattern is the approach for cleaning up unmanaged resources. for types with only managed resources a single dispose approach is sufficient

StefanOssendorf:
That's right. But we don't know what a user of our class does. So we should provide the necessary hook to participate in a proper disposal, no?

PR #4608, opened 2025-03-16
remove redundant nullable

SimonCropp:


StefanOssendorf (dismissed):
Thanks for catching these. Probably left overs from my work. I appreciate it!

StefanOssendorf:
Ah please solve the conflict. Thanks

Source/Csla/Core/MobileObservableCollection.cs
```diff
@@ -136,7 +136,7 @@ protected virtual void OnGetChildren(SerializationInfo info, MobileFormatter for
       List<int> references = new List<int>();
       for (int x = 0; x < Count; x++)
       {
-        T? child = this[x];
+        T child = this[x];
```
StefanOssendorf, line 139: Shouldn`t this still be `T?` because of the next line? Otherwise the check could be removed due to the missing annotation?

SimonCropp: no because ` MobileObservableCollection<T>` has no `notnull` constraint.

StefanOssendorf: But exactly because of that we have to assume it can be nullable.
So `T? child` is more correct than `T`. With `T` we don't express the possibility of the `child` being `null` so the compiler can help us avoiding possible `null` errors?

PR #4614, opened 2025-03-26
Allow disabling scan for data annotations attributes

### LINKED ISSUE #4492: Add option to ignore data annotation attributes

### LINKED ISSUE #4615: Ensure ApplicationContext isn't null during initialization
As business base types are instantiated, the `ApplicationContext` is set via the `IUseApplicationContext` interface.

Within the implementations of that interface, we should ensure that the provided `ApplicationContext` value is not null. If it is null, that leads to other bugs that are hard to debug, and it would be better to just throw an exception right away to indicate that the app environment isn't set up right, or that the business class isn't being created correctly.

rockfordlhotka:
Closes #4492 
Closes #4615


rockfordlhotka:
Already reviewed in #4603 - this PR retargets to the correct v9.x branch

rockfordlhotka:
@StefanOssendorf can you please review the changes in https://github.com/MarimerLLC/csla/pull/4614/commits/aa301d3372de70674e53237abfade92a4bb08053?

The other changes in this PR were already reviewed - but did reveal the need for these new changes.

Source/Csla/ApplicationContext.cs
```diff
@@ -26,12 +27,19 @@ public class ApplicationContext
...
     {
       ApplicationContextAccessor = applicationContextAccessor;
       ApplicationContextAccessor.GetContextManager().ApplicationContext = this;
+      Options = options;
     }
 
+    /// <summary>
+    /// Holds configuration options for the CSLA framework.
+    /// </summary>
+    public CslaOptions Options { get; }
```
ossendorf-at-hoelscher, line 41: I'm not sure if we should add it there. 
I'm not sure why the usage below (`PropertyChangedMode`) isn't set in the constructor? With that the `_propertyChangedModeSet` flag could be removed.

If you want it as an property here it should at least be `internal`. It's less to worry about regarding a bigger public API to maintain :D

Source/Csla/Core/BusinessBase.cs
```diff
@@ -1158,7 +1160,8 @@ void IHostRules.AllRulesComplete()
     /// </remarks>
     protected virtual void AddBusinessRules()
     {
-      BusinessRules.AddDataAnnotations();
+      if (ApplicationContext.Options.ScanDataAnnotations)
```
ossendorf-at-hoelscher, line 1163: Depending on the answer to my comment on the ApplicationContext retrieving the options here once should be okay.
This method is called once per type per process. So it's negligible.

ossendorf-at-hoelscher (changes requested):

Source/Csla/ApplicationContext.cs
```diff
@@ -26,12 +27,17 @@ public class ApplicationContext
     /// Creates a new instance of the type
     /// </summary>
     /// <param name="applicationContextAccessor"></param>
-    public ApplicationContext(ApplicationContextAccessor applicationContextAccessor)
+    /// <param name="options"></param>
+    public ApplicationContext(ApplicationContextAccessor applicationContextAccessor, CslaOptions options)
```
StefanOssendorf, line 31: As already mentioned in discord.
This is technically a breaking change. We should implement it this way in v10. But here we must use `ApplicationContextAccessor.ServiceProvider.GetRequiredService<CslaOptions>` to avoid the breaking change.

Source/Csla/Configuration/Fluent/CslaOptions.cs
```diff
@@ -64,6 +64,22 @@ public CslaOptions RegisterPropertyInfoFactory<
       return this;
     }
 
+    /// <summary>
+    /// Indicates whether the data annotations scan is disabled.
```
StefanOssendorf, line 68: enabled not disabled. The property is not a negation.

rockfordlhotka:
Oh, this is annoying. The test fails here, but doesn't fail on my machine. That's a bit hard to debug!

PR #4645, opened 2025-05-05
Fix missing generic support for auto implement properties generator

StefanOssendorf:
Fix missing generic support for auto implement properties generator
Add basic tests to verify the code is correctly generated

Instead of generating a lot of `using`s in the generated file the types are now globally fully qualified to distinguish them.


StefanOssendorf:
This must be ported/cherry picked to main. I'll do it after this is merged.

StefanOssendorf:
Marked as draft. I want to add a test with nested generics to make sure they work as expected.

StefanOssendorf:
Added. Ready to go :)

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp.Tests/Snapshots/AutoImplementPropertiesGeneratorTests.Case01#ReadOnlyPOCO.g.verified.cs
rockfordlhotka, line 1: Should these `.g.` files be in source control if they are generated?

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/AutoImplement/Discovery/DefinitionExtractionContext.cs
```diff
@@ -59,13 +60,86 @@ public string GetTypeNamespace(TypeSyntax typeSyntax)
...
+      else
+      {
+        typeSymbol = _semanticModel.GetSymbolInfo(typeSyntax).Symbol as INamedTypeSymbol;
+      }
+      if (typeSymbol is null || typeSymbol.ContainingNamespace is null)
+        return string.Empty;
+
+      var typeString = typeSymbol.ToString();
+#pragma warning disable IDE0066 // Convert switch statement to expression, This is more readable than the switch-expression
+      switch (typeString)
```
rockfordlhotka, line 90: Is there no other way to detect a "native" .NET type?

Perhaps, at least, this chunk of code should be a static method like `IsNativeDotNetType()` or something?

rockfordlhotka (changes requested):

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp.Tests/Snapshots/AutoImplementPropertiesGeneratorTests.Case01#ReadOnlyPOCO.g.verified.cs
StefanOssendorf: Yes. They are used to verify if the tests are still successful or not.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/AutoImplement/Discovery/DefinitionExtractionContext.cs
StefanOssendorf: I have to double check that. I took this snippet from my source generator and I found no other way at that time. 

I'll move that into some kind of method.

StefanOssendorf: Good catch. I got rid of it entirely :) Was not necessary. I'll also update my source gen because it's not necessary there, too xD

rockfordlhotka (approved):

PR #4656, opened 2025-05-17
Some fixes to AutoSerializable source generator

StefanOssendorf:
Fixes file scope namespace handling
Fixes generics handling through globally fully qualifying types


rockfordlhotka (changes requested):
There are a bunch of build warnings in autoserialization tests that need to be addressed.

StefanOssendorf:
> There are a bunch of build warnings in autoserialization tests that need to be addressed.

Should be done. I'm waiting for the build to check for any missed ones.

PR #4658, opened 2025-05-19
Add option to disable scan for DataAnnotations

### LINKED ISSUE #4492: Add option to ignore data annotation attributes

rockfordlhotka:
Closes #4492

### DIFF
--- a/Source/Csla.test/BusyStatus/BusyStatusTests.cs
+++ b/Source/Csla.test/BusyStatus/BusyStatusTests.cs
@@ -145,11 +145,13 @@ namespace cslalighttest.BusyStatus
       ItemWithAsynchRuleList items = ItemWithAsynchRuleList.GetListWithItems(dataPortal);
       items[0].ValidationComplete += async (_, _) =>
       {
+#pragma warning disable MSTEST0040 // Do not assert inside 'async void' contexts
         Assert.IsFalse(items.IsBusy);
         Assert.IsTrue(items.IsSavable);
         items = await items.SaveAsync();
         string actual = items[0].OperationResult;
         Assert.AreEqual("DataPortal_Update", actual);
+#pragma warning restore MSTEST0040 // Do not assert inside 'async void' contexts
       };
 
       items[0].RuleField = "some value";
--- a/Source/Csla/Configuration/Fluent/CslaOptions.cs
+++ b/Source/Csla/Configuration/Fluent/CslaOptions.cs
@@ -78,6 +78,22 @@ namespace Csla.Configuration
     }
 
     /// <summary>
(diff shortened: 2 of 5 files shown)


Source/Csla/Configuration/Fluent/CslaOptions.cs
```diff
@@ -77,6 +77,22 @@ public CslaOptions UseReflectionFallback(bool value)
...
+    /// <summary>
+    /// Indicates whether the data annotations scan is enabled.
+    /// </summary>
+    public bool ScanDataAnnotations { get; private set; } = true;
+
+    /// <summary>
+    /// Configures the scanning for data annotations based on the provided flag.
+    /// </summary>
+    /// <param name="flag">True to scan for data annotations, false to disable scanning. (default: true)</param>
+    /// <returns>Returns the current instance of CoreOptions.</returns>
```
StefanOssendorf, line 89: Not `CoreOptions` but `CslaOptions` :)

Source/Csla/Core/BusinessBase.cs
```diff
@@ -1159,7 +1160,9 @@ void IHostRules.AllRulesComplete()
     /// </remarks>
     protected virtual void AddBusinessRules()
     {
-      BusinessRules.AddDataAnnotations();
+      var options = ApplicationContext.GetRequiredService<CslaOptions>();
```
StefanOssendorf, line 1163: What about doing this instead of here there: https://github.com/MarimerLLC/csla/blob/faa3c2d973cf7b9e21f864b45df0b1b07d6d782d/Source/Csla/Rules/BusinessRules.cs#L1167 ?

StefanOssendorf (changes requested):

Source/Csla.test/BusyStatus/BusyStatusTests.cs
```diff
@@ -145,11 +145,13 @@ public void ListTestSaveWhileNotBusy()
       ItemWithAsynchRuleList items = ItemWithAsynchRuleList.GetListWithItems(dataPortal);
       items[0].ValidationComplete += async (_, _) =>
       {
+#pragma warning disable MSTEST0040 // Do not assert inside 'async void' contexts
```
StefanOssendorf, line 148: Don't suppress it. We have to fix it properly. This is a lying test. When the test method finishes the handler does not fail the test if it fails.

StefanOssendorf (changes requested):

Source/Csla.test/BusyStatus/BusyStatusTests.cs
rockfordlhotka: Noted in #4668

PR #4659, opened 2025-05-19
Stop caching current principal

### LINKED ISSUE #4612: Stop caching the principal in wasm Blazor context mgr
In the Blazor WebAssembly `IContextManager` implementation, stop caching the user principal, and get it from the authentication state provider each time.

The reasoning for this is documented in #4486

rockfordlhotka:
Closes #4612

### DIFF
--- a/Source/Csla.Blazor.WebAssembly/ApplicationContextManager.cs
+++ b/Source/Csla.Blazor.WebAssembly/ApplicationContextManager.cs
@@ -11,180 +11,177 @@ using Microsoft.AspNetCore.Components.Authorization;
 using System.Security.Claims;
 using System.Security.Principal;
 
-namespace Csla.Blazor.WebAssembly
+namespace Csla.Blazor.WebAssembly;
+
+/// <summary>
+/// Application context manager that uses HttpContextAccessor when 
+/// resolving HttpContext to store context values.
+/// </summary>
+public class ApplicationContextManager : IContextManager, IDisposable
 {
   /// <summary>
-  /// Application context manager that uses HttpContextAccessor when 
-  /// resolving HttpContext to store context values.
+  /// Creates an instance of the object, initializing it
+  /// with the required IServiceProvider.
   /// </summary>
-  public class ApplicationContextManager : IContextManager, IDisposable
+  /// <param name="authenticationStateProvider">AuthenticationStateProvider service</param>
+  public ApplicationContextManager(AuthenticationStateProvider authenticationStateProvider)
   {
-    /// <summary>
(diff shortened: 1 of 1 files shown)


Source/Csla.Blazor.WebAssembly/ApplicationContextManager.cs
```diff
@@ -11,180 +11,177 @@
...
+
+  private static readonly IPrincipal _anonymous = new ClaimsPrincipal(new ClaimsIdentity());
+
+  /// <summary>
+  /// Gets the current principal.
+  /// </summary>
+  public IPrincipal GetUser()
+  {
+    IPrincipal result;
+    if (AuthenticationState.IsCompleted && AuthenticationState.Result != null)
```
StefanOssendorf, line 78: Shouldn't you test for `IsCompletedSuccessfully` instead of `IsCompleted`. The latter can also mean cancelled or faulted (exception). I'm not sure if this would result in an exception here.

Source/Csla.Blazor.WebAssembly/ApplicationContextManager.cs
```diff
@@ -11,180 +11,177 @@
...
-    /// Gets the local context.
-    /// </summary>
-    public IContextDictionary GetLocalContext()
+  /// <summary>
+  /// Gets the local context.
+  /// </summary>
+  public IContextDictionary GetLocalContext()
+  {
+    IContextDictionary localContext;
+    var sessionManager = ApplicationContext.GetRequiredService<ISessionManager>();
```
StefanOssendorf, line 102: I know not part of this PR. But is there any reason why `ISessionManager` can't be a dependency provided through the constructor?

StefanOssendorf (changes requested):
Should you add tests for this to document that?

Source/Csla.Blazor.WebAssembly/ApplicationContextManager.cs
rockfordlhotka: I think that's a good improvement. I believe the null check prevents an exception, but this will be better.

Source/Csla.Blazor.WebAssembly/ApplicationContextManager.cs
rockfordlhotka: Agreed, I'll make this change

rockfordlhotka: LOL! Or not, as that would result in a circular DI dependency

PR #4670, opened 2025-05-25
Fix bugs of source gens found during main port v9

StefanOssendorf:
Nullable handling improved
Arrays are now correctly handled (again)

Tests for all built in types added.


rockfordlhotka (changes requested):
There are build warnings that need to be addressed in unchanged files (that I assume were affected by the other changes).

rockfordlhotka (approved):

PR #4690, opened 2025-06-20
4385- Changes to the build for maui

### LINKED ISSUE #4385: Split the MAUI build/pack into its own process
Because .NET MAUI has a separate release cycle from .NET itself, the `Csla.Xaml.Maui` project and NuGet package also needs to have a separate release cycle from the rest of CSLA.

@Bowman74 has some good insights into this in #4349

I think it will require

* Pulling the `Csla.Xaml.Maui` project out of the `csla.build` solution
* Changing `Csla.Xaml.Maui.csproj` to reference CSLA via NuGet instead of as a project reference
* Creating a new solution specifically for building the CSLA MAUI project
* Creating a new GitHub action yaml file to build/pack/push the CSLA MAUI project

Bowman74:
#4385 

- csla.build.sln no longer contains Csla.xaml.maui project 
- csla.maui.build.sln created to build csla.xaml.maui.csproj 
- csla.xaml.maui.csproj changed to reference CSLA via a nuget reference 
- directory.package.props created and placed in the csla.xaml.maui directory. The primary purpose of this is to change the output directory to bin\maui-packages instead of bin\packages 
- release.maui.yaml created that builds the new csla.mauil.build.sln and then pushes the nuget packages found under bin\maui-packages

NOTE: The only way to fully test this end to end would be via a release.


Bowman74:
@rockfordlhotka I have also noticed that the ci build uses csla.test.sln. This solution did not before, nor does it now, contain a reference to csla.xaml.maui.csproj. This means that not only does this project not have any unit tests (which may be fine), but it also does not build as a part of a ci build (which also may be fine).

It's probably OK as there is no code directly in it (only code from csla.xaml.shared). But if code ever were added to that project as things stand now, it would never go through any CI verification.

rockfordlhotka:
iirc, we don't include the MAUI projects in the CI build because that requires a Windows runner and also requires installation of the MAUI workload on each run - which consumes build minutes.

Now that we have a separate release for MAUI, I think we could consider adding a new CI build action to detect when changes have happened within the MAUI subtree and to then run a CI build using a Windows runner and install the workload. That would be a good improvement - for a different PR.

Source/csla.build.sln
```diff
@@ -3804,7 +3601,6 @@ Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
 		Csla.Web.Mvc.Shared\Csla.Web.Mvc.Shared.projitems*{31affcd2-f8fe-4b0b-868f-36258f421148}*SharedItemsImports = 5
 		Csla.Xaml.Shared\Csla.Xaml.Shared.projitems*{3698d49c-25b5-4f46-bde2-14008231e68d}*SharedItemsImports = 13
-		Csla.Xaml.Shared\Csla.Xaml.Shared.projitems*{396810b8-3881-4ab7-8f72-663d0690e9d7}*SharedItemsImports = 5
```
rockfordlhotka, line 3807: Isn't this shared project still required by the WPF project? It isn't specific to MAUI, as it is code that is common for all XAML projects (iirc).

.github/workflows/release-maui.yaml
```diff
@@ -0,0 +1,32 @@
+name: release
+
+# trigger on a push to any release/* branch
+on:
+  push:
+    branches:
+      - 'release/*'
```
rockfordlhotka: Does this need to be a different branch prefix name? Otherwise won't we end up releasing everything all the time like today?

Right now when I create a new `release/xyz` branch it builds and pushes the nuget packages for everything - core and maui.

The goal is to be able to create a `release/xyz` branch to build core, and a different branch to build maui? So this might need to be `release-maui/xyz`?

Or am I misunderstanding?

rockfordlhotka (changes requested):

Source/csla.build.sln
Bowman74: Yes, that's the shared project entry from the csla.xaml.maui project in the csla.build.sln file that no longer builds the maui project (the project id for csla.xaml.maui is 396810b8-3881-4ab7-8f72-663d0690e9d7, the shared project reference that was removed). The guid for csla.xaml.wpf project is c5dd3150-ada0-4a47-8cc1-8d9020278b30 and the shared project entry for that is on line 3606 of the csla.build.sln file.

.github/workflows/release-maui.yaml
Bowman74: Good catch, fixed.

rockfordlhotka (approved):

PR #4695, opened 2025-06-25
Feature/4691 MAUI CI build

### LINKED ISSUE #4691: Add MAUI CI build
Create a new CI build that triggers only when code is changed in the MAUI subtree. This CI build should build and run the MAUI code and any MAUI related unit tests.

I don't know if there are any MAUI unit tests at this time? If not, perhaps create a MAUI test project with at least one test just so we can establish the pattern and ensure that the CI action does the right thing.

Bowman74:
Changes to build maui xaml project if anything maul specific has changed. The following changes were made:

- new csla.maui.test.sln made to do just the maui build
- new job added to the CI build that is dependent on the current job. new job runs on MacOS and only runs if the current job succeeds and any maui specific files changed, including the shared xaml project
- The build tries platform builds for Android, iOS and MacOS
- Csla.Xaml.Maui.csproj changed to meet .net 9 minimum versions for iOS and MacOS


Bowman74:
@rockfordlhotka It appears this PR build the new CI build, including the maui job as a maui file did indeed change. This should allow you to look at it. The Build MAUI job will only run if it detects some file important to CSLA MAUI has changed and the Build Main job will always run (it is the old CI build with the addition of taking a look to see if any MAUI files have changed).

Bowman74:
@rockfordlhotka And if you don't normally, I think you definitely want to squash this PR when merging!

Bowman74:
@rockfordlhotka 
I further modified the maui xaml file to use the old minimum iOS and MacOS versions if compiled under .net 8 and the new ones if compiled under any other .Net version. I don't know if we ever compile the projects like that (it doesn't seem we do, at least for the nuget packages) but this will leave the existing minimums in place where they can be used (.Net 8).

Bowman74:
I did change the name of the original job back so the PR dependency doesn't have to change. However, there is not a PR dependency on the MAUI job. That would have to be a conditional dependency for the PR only if the job runs and I don't think GitHub supports that. So with this, the PR won't have a dependency on the new MAUI job but you will be able to see in the PR if it ran and if it was successful.

rockfordlhotka:
.NET 8 will go out of support in November, when .NET 10 supercedes it. Our `main` branch will stop targeting .NET 8 at that point.

Bowman74:
> .NET 8 will go out of support in November, when .NET 10 supercedes it. Our `main` branch will stop targeting .NET 8 at that point.

@rockfordlhotka 
Let me know if you want to change the project file to pull out the conditional .Net 8 stuff. If not, this PR is ready to go.

rockfordlhotka:
It is good to leave it in for now, as we still build dotnet 8 today

Bowman74:
> It is good to leave it in for now, as we still build dotnet 8 today

Sounds good. Then I'll consider this one done.

PR #4724, opened 2025-09-18
Remove Backport.System.Threading.Lock dependency

### LINKED ISSUE #4450: Fix issues with Backport.System.Threading.Lock package

StefanOssendorf:
Removes the `Backport.System.Threading.Lock` dependency from `Csla` to avoid shipping it as a source generator to consumers.

Closes #4450

### DIFF
--- a/Source/Csla.Analyzers/Csla.Analyzers/Csla.Analyzers.csproj
+++ b/Source/Csla.Analyzers/Csla.Analyzers/Csla.Analyzers.csproj
@@ -41,11 +41,4 @@
     </EmbeddedResource>
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="Backport.System.Threading.Lock" Version="3.1.4" />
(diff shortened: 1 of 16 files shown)


MarkCiliaVincenti:
You could have used it as a source generator and that would have not made any transitive dependencies to the consumers. It is only transitive as it is not being used as a source generator.

StefanOssendorf:
I tried that but got errors that the type is already present in other assemblies. I think we have InternalsVisibleTo in use which causes that. Instead of investing time to fix that I opted and discussed the removal with Rocky.

PR #4743, opened 2025-10-16
Document the process of creating docs with doxygen

### LINKED ISSUE #4742: Automate creation of documentation
To date poeple have (on an ad-hoc basis) stepped up to create docs for CSLA based on the XML comments in the code. Which has been very much appreciated!

It would be better, however, if there was a documented process to create the documentation, possibly with some automation or a GitHub Action or something.

At the very least, having a documented process flow for creating the docs would make it easier to generate them for each release.

rockfordlhotka:
Closes #4742 

I have created a `csla/Source/docs` folder that contains a `readme.md` describing how to use doxygen to create docs for CSLA based on the XML comments in code (and readme.md files in the source tree).

I also had copilot do a bunch of cleanup of the XML comments that were generating warnings from doxygen, and I created the content for the landing page in the generated help web site.

### DIFF
--- a/Source/Csla.Analyzers/Csla.Analyzers/Csla.Analyzers.csproj
+++ b/Source/Csla.Analyzers/Csla.Analyzers/Csla.Analyzers.csproj
@@ -15,7 +15,7 @@
     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.11.0" PrivateAssets="all" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.10.0" PrivateAssets="all" />
     <PackageReference Update="NETStandard.Library" PrivateAssets="all" />
-    <PackageReference Include="Polyfill" Version="7.4.0">
+    <PackageReference Include="Polyfill" Version="8.9.0">
       <PrivateAssets>all</PrivateAssets>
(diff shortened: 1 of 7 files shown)


Source/Csla/Server/ObjectFactoryLoader.cs
```diff
@@ -38,7 +38,7 @@ public ObjectFactoryLoader(ApplicationContext applicationContext)
     /// on the business object.
     /// </param>
     /// <exception cref="ArgumentException"><paramref name="factoryName"/> is <see langword="null"/>, <see cref="string.Empty"/> or only consists of white spaces.</exception>
-    /// <exception cref="InvalidOperationException"><see cref="Type"/> for <paramref name="factoryName"/> could not be loaded.</exception>
+  /// <exception cref="InvalidOperationException"><c>System.Type</c> for <paramref name="factoryName"/> could not be loaded.</exception>
```
StefanOssendorf: This looks wrong. Why should we give up on referencing another type the correct way?  See cred is the way to go to point to another type instead of using the fixed string for that type.

rockfordlhotka: I reverted the change - doxygen throws warnings about the external cref links, but we can live with those.

Issue #4622, opened 2025-04-01
Tweak/Cleanup of ISerializationNotification.Deserialized()


swegele:
Per Rocky's [request here in discord](https://discord.com/channels/795719451148156938/795719451148156941/1356662270859415714)...

_Background:  While upgrading from CSLA8 to CSLA9 I noticed the overridable OnDeserialized() method is gone (presumably because of the changes made related to removing .net binary de/serialization)._

Three things to look at in V9 (and fourth for v10) related to `ISerializationNotification.Deserialized()`:

**FIRST:**
Can the method `ISerializationNotification.Deserialized` be marked overridable so we can hook into it?...as a replacement for the old `OnDeserialized`.  No need for any parameters in the new method AFAIK.

**SECOND:**
See `ReadOnlyBase.cs` Line 513 -> Inside the method `void ISerializationNotification.Deserialized()`, why does the code new up an `System.Runtime.Serialization.StreamingContext()` and pass it to the method `private void OnDeserializedHandler(`?  That `context` parameter is unused so it is not needed.  Also that extra method seems pointless, the little code there could simply be moved into original method.
Same thing in `BusinessRules.cs` 

**THIRD:**
Why is the attribute [System.Runtime.Serialization.OnDeserialized] added to the previously mentioned (seemingly useless) method `private void OnDeserializedHandler` only in the `ReadOnlyBase` and `BusinessRules` classes? That attribute is not used or referenced anywhere.

**FOURTH:**
For CSLA10, add async support related to this (asked for by Stefan)

rockfordlhotka:
@StefanOssendorf so we're clear, adding an `OnDeserializedAsync` means the entire MobileFormatter needs to be changed to be async.

StefanOssendorf:
@rockfordlhotka Yes.
We should porbably add
* `OnDeserializingAsync()`
* `OnDeserializedAsync()`

as hooks into the deserialization pipeline.
And when we are at it probably
* `OnSerializingAsync()`
* `OnSerializedAsync()`

too.

That would mean we have to extend the `ISerializationFormatter` interface and add `Async` variants.

I don't see why we have to change `IMobileObject` though? I can't remember our exact conversation on discord, but I'm happy with hooks in the process. I don't think we have to make the `Get-/SetState` or `Get-/SetChildren` methods async.
Maybe I'm forgetting something important here - dunno 😅.

rockfordlhotka:
[...]

PR #4744, opened 2025-10-17
#4622 Implement virtual Deserialized method

rockfordlhotka:
FIRST:
Can the method ISerializationNotification.Deserialized be marked overridable so we can hook into it?...as a replacement for the old OnDeserialized. No need for any parameters in the new method AFAIK.

SECOND:
See ReadOnlyBase.cs Line 513 -> Inside the method void ISerializationNotification.Deserialized(), why does the code new up an System.Runtime.Serialization.StreamingContext() and pass it to the method private void OnDeserializedHandler(? That context parameter is unused so it is not needed. Also that extra method seems pointless, the little code there could simply be moved into original method. Same thing in BusinessRules.cs

THIRD:
Why is the attribute [System.Runtime.Serialization.OnDeserialized] added to the previously mentioned (seemingly useless) method private void OnDeserializedHandler only in the ReadOnlyBase and BusinessRules classes? That attribute is not used or referenced anywhere.

### DIFF
--- a/Source/Csla/ReadOnlyBase.cs
+++ b/Source/Csla/ReadOnlyBase.cs
@@ -530,6 +530,9 @@ namespace Csla
 
     void ISerializationNotification.Deserialized()
     {
+      if (_fieldManager != null)
+        FieldManager.SetPropertyList(GetType());
+      InitializeBusinessRules();
       Deserialized();
     }
 


rockfordlhotka:
Note: This does not include making anything async. That is a much bigger task and will be either a separate PR or will happen in a future version.

PR #4745, opened 2025-10-17
CSLA 10 upgrade doc

rockfordlhotka:
This PR is for updates to docs and information relative to CSLA 10


ossendorf-at-hoelscher:
We have to add a warning:
With CSLA 10 it's not possible anymore to set csla-properties in a BO/Command constructor. This will result in a ArgumentNullException due to the missing ApplicationContext for the FieldManager.

PR #4750, opened 2025-10-23
Implement IMobileObjectMetastate interface

### LINKED ISSUE #4263: Add a "Metastate" Interface for BOs to Implement to Simplify Serialization
Right now, a serialization author needs to know inaccessible information about a BO to read/write it correctly via a serialization.

I'm currently working on a source generator-based serializer for CSLA (https://github.com/JasonBock/CslaGeneratorSerialization). Recently, I updated to a preview version of CSLA (9.0.0-R24070301), and....something changed internally on a BO (`BusinessBase<>` I believe) such that it completely broke my Reflection-based code. I'd rather not write that code, but it's essential to properly handle serialization of BO types correctly. Really, what should happen is that a BO should know how to read/write its' own metastate, and a serialization author could then ask a BO to read/write its' state without having to worry about private field names and whatnot.

Therefore, I propose that CSLA adds an interface (I'll call it `IMetastate` for now, but the final name I'll leave up to Rocky) that has methods that a BO would implement:

```c#
public interface IMetastate
{
  void Write(BinaryWriter writer);
  void Read(BinaryReader reader);
}
```

Then, as a serialization author, I could invoke these methods at the appropriate times so the values for "is new", "is dirty", etc. are saved and retrieved correctly, and I no longer care how that's done.

Leave things the way they are, and hope that future CSLA changes don't change how internal metastate is managed. (Or, if such a mechanism already exists in CSLA, **please** tell me :) ).

rockfordlhotka:
Closes #4263 

@JasonBock does this match your thoughts?

### DIFF
--- a/Source/Directory.Package.props
+++ b/Source/Directory.Package.props
@@ -1,8 +1,19 @@
 <Project>
 
+ <!-- Read version from version.json -->
+  <Target Name="ReadVersionFromJson" BeforeTargets="Build;Pack">
+    <ReadLinesFromFile File="$(MSBuildThisFileDirectory)version.json">
+      <Output TaskParameter="Lines" ItemName="VersionJsonLines" />
+    </ReadLinesFromFile>
+    <PropertyGroup>
+      <VersionJsonContent>@(VersionJsonLines)</VersionJsonContent>
+      <PackageVersionFromJson>$([System.Text.RegularExpressions.Regex]::Match($(VersionJsonContent), '"version"\s*:\s*"([^"]+)"').Groups[1].Value)</PackageVersionFromJson>
+    </PropertyGroup>
+  </Target>
  
  <!-- Common Package Settings -->
   <PropertyGroup>
+   <PackageVersion>$(PackageVersionFromJson)</PackageVersion>
    <PackageLicenseExpression>MIT</PackageLicenseExpression>
    <PackageProjectUrl>https://cslanet.com</PackageProjectUrl>
    <OutputPath>..\..\bin\packages\</OutputPath>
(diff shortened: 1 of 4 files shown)


JasonBock:
If it handles round-tripping [these fields](https://github.com/JasonBock/CslaGeneratorSerialization/blob/main/src/CslaGeneratorSerialization.Analysis/Extensions/IncrementalGeneratorInitializationContextExtensions.cs) and lets me get rid of [this code](https://github.com/JasonBock/CslaGeneratorSerialization/blob/main/src/CslaGeneratorSerialization.Analysis/Builders/BusinessBaseBuilder.cs#L42),, then yes. Specifically, fields like `_editLevelAdded`, `_identity`, etc. Right now I generate code that uses `UnsafeAccessor` to help a bit with Reflection performance, and the goal of this feature would be that I wouldn't care about any of this. I could tell a BO to serialize its state and/or metastate, and it would just work.

But, the way this is implemented, there's still Reflection involved. I'm not sure that's ideal. Couldn't each BO implement `IMobileObjectMetastate`, which would make it such that it knows which metastate fields to roundtrip, and that would eliminate Reflection usage altogether?

Basically, `IMobileObjectMetastate` is fine. But I don't get why `MobileObjectMetastateHelper` is there, using Reflection to do the work. Seems like each stereotype should implement this and handle its fields directly.

rockfordlhotka:
The updated implementation is much simpler - doesn't use the `StateMode` property and therefore only works for serialization, not undo. By doing this it eliminates the need for reflection.

@JasonBock you don't care what is in the byte array I assume? It is JSON in this case, but you are just going to treat it like a blob I assume?

JasonBock:
> The updated implementation is much simpler - doesn't use the `StateMode` property and therefore only works for serialization, not undo. By doing this it eliminates the need for reflection.
> 
> @JasonBock you don't care what is in the byte array I assume? It is JSON in this case, but you are just going to treat it like a blob I assume?

Correct, I don't care what's in the array. I assume the implementor knows what they're doing ;)

rockfordlhotka:
> Correct, I don't care what's in the array. I assume the implementor knows what they're doing ;)

That's just silliness!

PR #4777, opened 2025-12-08
Update to use binary serialization instead of JSON

### LINKED ISSUE #4746: Add async Deserialized methods
Right now (as of #4622) there are virtual `Deserialized` methods that are invoked as each object in the graph is deserialized.

It would be ideal to have async versions of this method as well. This may be a fairly large task, because the code that _invokes_ `Deserialized` is not currently async, and so there is likely to be a cascade effect of making a lot of things async that are currently sync.

rockfordlhotka:
Closes #4746

### DIFF
--- a/Source/Csla/Server/TransactionalDataPortal.cs
+++ b/Source/Csla/Server/TransactionalDataPortal.cs
@@ -89,6 +89,7 @@ namespace Csla.Server
         TransactionIsolationLevel.RepeatableRead => IsolationLevel.RepeatableRead,
         TransactionIsolationLevel.ReadCommitted => IsolationLevel.ReadCommitted,
         TransactionIsolationLevel.ReadUncommitted => IsolationLevel.ReadUncommitted,
+        TransactionIsolationLevel.Snapshot => IsolationLevel.Snapshot,
         _ => IsolationLevel.Unspecified
       };
     }
--- a/Source/Csla/TransactionIsolationLevel.cs
+++ b/Source/Csla/TransactionIsolationLevel.cs
@@ -38,7 +38,14 @@ namespace Csla
     /// <summary>
     /// Shared locks are issued and no exclusive locks are honored.
     /// </summary>
-    ReadUncommitted
+    ReadUncommitted,
+    /// <summary>
+    /// Volatile data can be read. Before a transaction modifies data, it verifies if
+    /// another transaction has changed the data after it was initially read. If the
+    /// data has been updated, an error is raised. This allows a transaction to get to
+    /// the previously committed value of the data.
+    /// </summary>
+    Snapshot
   }
 }
 #endif


JasonBock:
@rockfordlhotka There are no files changes in this PR?

rockfordlhotka:
> @rockfordlhotka There are no files changes in this PR?

Whoa! Where'd they go???? Well damn!

rockfordlhotka:
> > @rockfordlhotka There are no files changes in this PR?
> 
> Whoa! Where'd they go???? Well damn!

No, there are here:

https://github.com/MarimerLLC/csla/pull/4777/commits/dd9a25013baad7efc05e222bdd0ced36f8139ee8#diff-bac8c334722ef8a8a5f7145e4cdf071c73ef8a38578a6559642d3ecbf92e3ecb

JasonBock:
> > > @rockfordlhotka There are no files changes in this PR?
> > 
> > 
> > Whoa! Where'd they go???? Well damn!
> 
> No, there are here:
> 
> [dd9a250#diff-bac8c334722ef8a8a5f7145e4cdf071c73ef8a38578a6559642d3ecbf92e3ecb](https://github.com/MarimerLLC/csla/commit/dd9a25013baad7efc05e222bdd0ced36f8139ee8#diff-bac8c334722ef8a8a5f7145e4cdf071c73ef8a38578a6559642d3ecbf92e3ecb)

Odd, because the PR itself shows no files.

JasonBock:
@rockfordlhotka I added two comments in that diff

rockfordlhotka:
> @rockfordlhotka I added two comments in that diff

I wonder if GitHub is having issues - that file is absolutely changed, and yet I don't see your comments - something is very messed up.

JasonBock:
Maybe. Dunno. Here are my two comments:

"Isn't this code copy/pasted from somewhere else in CSLA? This feels like something I've seen in serialization before - that is, each type is handled separately." - this is referencing the Write/ReadValue calls.

"With structs, this is going to cause boxing. Could this be made generic?" - this is referring to the Value field in FieldDataResult.

rockfordlhotka:
@JasonBock that last commit eliminates the FieldDataResult to stop that level of boxing. As Copilot points out, there's a deeper issue here that would be hard to address:

> Note: The value itself is still boxed because ReadValue(BinaryReader) returns object? and AddValue(string, object?) accepts object?. To completely eliminate boxing would require a much more invasive change to make the entire serialization infrastructure generic, which would be a significant architectural change beyond the scope of this refactoring.

JasonBock:
I didn't expect metastate to be as complicated as it's turning out to be. I thought it was basically fields based on very primitive types, like `bool` or `int`. Doing serde on them seemed like just calling method directly on the binary reader and writer that handle those types directly. Moreover, you'd know the order that you would write to and read from the streams.

rockfordlhotka:
I think it is simple now - because that helper class has become super short - just handling write/read from the stream as you say.

The "big" change was de-duplicating the existing code to read/write different types.

At least GitHub is behaving now - showing the list of changed files, etc.

JasonBock:
What I meant was, for the standard stereotypes in CSLA, you should know what the metastate is, correct? You're calling `GetState()`, which I'm guessing returns the metastate for each stereotype? But if so, you should be able to still call the type-safe versions of the binary streams without a big if/switch statement.

rockfordlhotka:
I thought about that, but it would mean duplicating the existing get/set state methods (basically) - which could easily be a long-term maintainability issue as someone adding a new property/field would need to define the field and then add it to all _four_ serialization methods instead of the existing two.

Maybe it isn't that big of a deal, as we rarely add new fields. On the other hand, _because_ we rarely add new fields, this is a situation that is ripe for sneaky future bugs.

JasonBock:
Personally, I don't see it as a big deal. Seriously, how often does the stereotype's metastate change? I mean, I don't know, you would know far better than I would :), but I can't imagine it changes often. And if there is a change, there are ways to mitigate potential issues with change to minimize any bugs.

rockfordlhotka:
> Personally, I don't see it as a big deal. Seriously, how often does the stereotype's metastate change? I mean, I don't know, you would know far better than I would :), but I can't imagine it changes often. And if there is a change, there are ways to mitigate potential issues with change to minimize any bugs.

I think you are right, so I did update the implementation to directly use a binary reader/writer. I think this is better - some minor duplication of functionality, but faster and cleaner at runtime.

What do you think @JasonBock ?

JasonBock (approved):

JasonBock:
I think overall this is a step in the right direction. You may want to add documentation somewhere if you haven't already to inform developers that implementing the metastate interface should **only** be done by types are representing a "stereotype" and it isn't meant for serialization of a specific business object. I can see someone using (abusing?) this interface and the stereotype's protected members to do that, and I don't think that's correct.

PR #4782, opened 2025-12-16
#4779 Add Snapshot TransactionIsolationLevel

### LINKED ISSUE #4779: Add Snapshot TransactionIsolationLevel
The transactional attribute currently allows you to choose your isolation level. However, CSLA's TransactionIsolationLevel enum does not include the option for Snapshot.

Add Snapshot as an option to TransactionIsolationLevel. Also the new option needs to be handled in TransactionalDataPortal.GetIsolationLevel

I looked for ways to handle this without this change but the TransactionalDataPortal doesn't seem to be extendable or replaceable.

I would be more than happy to send a PR for this. Hopefully this can be included in a CSLA 9 update!

xal1983:
Closes #4779

### DIFF
--- a/Source/Csla/Core/FieldManager/FieldDataManager.cs
+++ b/Source/Csla/Core/FieldManager/FieldDataManager.cs
@@ -790,7 +790,7 @@ namespace Csla.Core.FieldManager
         _stateStack.Clear();
         if (info.Values.ContainsKey("_stateStack"))
         {
-          var stackArray = info.GetValue<byte[][]>("_stateStack");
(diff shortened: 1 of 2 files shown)


rockfordlhotka:
@xal1983 there's some build issue - please make sure the csla.test.sln builds on your machine.

xal1983:
Hi @rockfordlhotka, interestingly, it builds on VS2022 but fails in 2026. The same issue was fixed on main by casting those values to IEnumerable<byte[]>, so I applied the same fix.

Issue #4789, opened 2026-01-09
CSLA 10 release


rockfordlhotka:
Finalize the CSLA 10 release - including Csla.Maui, the release notes, docs, samples, and other assets.

PR #4795, opened 2026-01-14
Add ICslaObject constraint to IDataPortal<T> and IChildDataPortal<T>

StefanOssendorf:
Add constraint to IDataPortal<T> and IChildDataPortal<T> to satisfy the constraint on the actual implementations.

The actual implementations `DataPortal<T>` or `IChildDataPortal` have these constraints. So we have to propagate them through the interfaces to avoid the runtime exception where T does not satisfy the constraint and instead make it a compile time error.


StefanOssendorf:
Link to #4789

rockfordlhotka (approved):

Issue #4794, opened 2026-01-13
EditLevelMismatch when cloning object


xal1983:
**Describe the bug**
When cloning an object graph that contains children, the edit level of the children is not correctly preserved.
Example scenario is Invoice with Cost line items. There is a BindingSource for the invoice and one for the child cost list.
Creating a clone and showing a new form with that cloned object triggers an EditLevelMismatch exception.

**Version and Platform**
CSLA version: 9.1.0
OS: Windows
Platform: WinForms

**Code that Fails**
I created a simple demo project to showcase the issue here: https://github.com/xal1983/CslaEditLevelMismatch/blob/master/EditLevelMismatch/Form1.cs

But the gist of it is, the child cost and it's field manager have an EditLevel = 1 at the time of cloning the invoice.
The cloned cost however ends up with EditLevel = 0 but it's FieldManager has an EditLevel = 1.
When setting the binding source, the binding source will automaticall call IEditableObject.BeginEdit() on the cost and Csla will raise the exception because the FieldManager and the Cost's EditLevels do not match.

**Stack Trace or Exception Detail**
   at Csla.Core.FieldManager.FieldDataManager.Csla.Core.IUndoableObject.CopyState(Int32 parentEditLevel, Boolean parentBindingEdit) in /_/Source/Csla/Core/FieldManager/FieldDataManager.cs:line 512
   at Csla.Core.UndoableBase.CopyState(Int32 parentEditLevel) in /_/Source/Csla/Core/UndoableBase.cs:line 167
   at Csla.Core.BusinessBase.System.ComponentModel.IEditableObject.BeginEdit() in /_/Source/Csla/Core/BusinessBase.cs:line 778
   at System.Windows.Forms.CurrencyManager.OnCurrentChanged(EventArgs e)
   at System.Windows.Forms.CurrencyManager.ChangeRecordState(Int32 newPosition, Boolean validating, Boolean endCurrentEdit, Boolean firePositionChange, Boolean pullData)
   at System.Windows.Forms.CurrencyManager.List_ListChanged(Object sender, ListChangedEventArgs e)
   at System.Windows.Forms.BindingSource.OnListChanged(ListChangedEventArgs e)
   at System.Windows.Forms.BindingSource.ResetBindings(Boolean metadataChanged)
   at System.Windows.Forms.BindingSource.SetList(IList list, Boolean metaDataChanged, Boolean applySortAndFilter)
   at System.Windows.Forms.BindingSource.ResetList()
[...]

PR #4796, opened 2026-01-16
Fix issue with editlevel and clone

rockfordlhotka:
#4794

  Fix edit level preservation when cloning object graphs with children

  When cloning an object graph during an active edit session, child objects
  were having their edit levels incorrectly reset. This occurred because
  InsertItem called UndoableBase.ResetChildEditLevel when adding deserialized
  children back to their parent collection, which cleared the child's
  already-restored state stack.

  The fix adds an _isDeserializing flag to BusinessListBase and
  BusinessBindingListBase that is set during OnSetChildren. When this flag
  is true, InsertItem skips the ResetChildEditLevel call, preserving the
  child's correctly deserialized edit level and undo state.

  This enables the clone-edit-save pattern where users can:
  - Begin editing an object
  - Clone it to show in a separate view
  - Continue editing or cancel on either the original or clone independently

  Files changed:
  - Source/Csla/BusinessListBase.cs
  - Source/Csla/BusinessBindingListBase.cs
  - Source/Csla.test/BusinessListBase/BusinessListBaseTests.cs

### DIFF
--- a/Source/Csla.test/BusinessListBase/BusinessListBaseTests.cs
+++ b/Source/Csla.test/BusinessListBase/BusinessListBaseTests.cs
@@ -421,6 +421,45 @@ namespace Csla.Test.BusinessListBase
       Assert.AreEqual(0, clonedRoot.Children.Count, "After CancelEdit, child added during edit should be removed");
     }
 
+    [TestMethod]
+    public void ClonePreservesChildEditLevelWhenUsingBindingSource()
+    {
+      // Arrange: Create a root with children and begin editing
+      var root = CreateRoot();
+      var child = root.Children.AddNew();
+      child.Data = "original";
+
+      ((System.ComponentModel.IEditableObject)root).BeginEdit();
+      ((System.ComponentModel.IEditableObject)child).BeginEdit();
+
+      // Make changes after BeginEdit
+      child.Data = "modified";
+
+      // Verify edit levels before clone
+      var rootEditLevel = ((Core.IUndoableObject)root).EditLevel;
+      var childEditLevel = ((Core.IUndoableObject)child).EditLevel;
+      var childListEditLevel = ((Core.IUndoableObject)root.Children).EditLevel;
+
+      Assert.AreEqual(1, rootEditLevel, "Root EditLevel should be 1 before clone");
+      Assert.AreEqual(1, childEditLevel, "Child EditLevel should be 1 before clone");
+      Assert.AreEqual(0, childListEditLevel, "ChildList EditLevel should be 0 before clone");
+
+      // Act: Clone the object graph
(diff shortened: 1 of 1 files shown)


rockfordlhotka:
@xal1983 I think this fixes the issue, do you concur?

xal1983:
Hi @rockfordlhotka ,
Yes, that is exactly what I was looking for! Tested in my sample code and works fine. Thank you!! 

Happy to approve as is, but I did want to mention the test itself doesn't cover the specific scenario I was having. Here's another test you could add if you want to cover it. The key point is that the BindingSource will not call BeginEdit on the collection itself and the EditLevel in the collection is 0, while in the root and child Editlevel is 1. I couldn't quite make out why calling it via IEditableObject has a different behavior but it is consistent.

`
[TestMethod]
public void ClonePreservesChildEditLevelWhenUsingBindingSource()
{
  // Arrange: Create a root with children and begin editing
  var root = CreateRoot();
  var child = root.Children.AddNew();
  child.Data = "original";

  ((System.ComponentModel.IEditableObject)root).BeginEdit();
  ((System.ComponentModel.IEditableObject)child).BeginEdit();

  // Make changes after BeginEdit
  child.Data = "modified";

  // Verify edit levels before clone
  var rootEditLevel = ((Core.IUndoableObject)root).EditLevel;
  var childEditLevel = ((Core.IUndoableObject)child).EditLevel;
  var childListEditLevel = ((Core.IUndoableObject)root.Children).EditLevel;

  Assert.AreEqual(1, rootEditLevel, "Root EditLevel should be 1 before clone");
  Assert.AreEqual(1, childEditLevel, "Child EditLevel should be 1 before clone");
  Assert.AreEqual(0, childListEditLevel, "ChildList EditLevel should be 0 before clone");

  // Act: Clone the object graph
  var clonedRoot = root.Clone();

  // Assert: Verify edit levels are preserved after clone
  var clonedRootEditLevel = ((Core.IUndoableObject)clonedRoot).EditLevel;
  var clonedChildEditLevel = ((Core.IUndoableObject)clonedRoot.Children[0]).EditLevel;
  var clonedChildListEditLevel = ((Core.IUndoableObject)clonedRoot.Children).EditLevel;

  Assert.AreEqual(1, clonedRootEditLevel, "Cloned Root EditLevel should be 1");
  Assert.AreEqual(1, clonedChildEditLevel, "Cloned Child EditLevel should be 1");
  Assert.AreEqual(0, clonedChildListEditLevel, "Cloned ChildList EditLevel should be 0");

  // Verify the modified data is preserved
  Assert.AreEqual("modified", clonedRoot.Children[0].Data, "Modified data should be preserved in clone");
}
`

PR #4815, opened 2026-02-05
Implement BusinessDocumentBase combining BusinessBase and BusinessListBase

### LINKED ISSUE #1830: Add rules engine to BusinessListBase
In the past there was no way to data bind to properties of a collection, and so the collection base types have never had a rules engine built in.

Really since XAML though, it _has_ been possible to bind to properties of a list. That's now true in Blazor as well as XAML (and I think in Razor Pages as well).

At the very least, it should be possible to have rules that run when a child changes, resulting in output property values that change properties of the list itself.

This requires some research and serious thought - as it _could_ lead to BusinessListBase having most/all the same logic and capabilities of BusinessBase. ???

rockfordlhotka:
## Summary

- Implements `BusinessDocumentBase<T,C>`, a new base class that inherits from `BusinessBase<T>` and adds full `BusinessListBase<T,C>` collection capabilities
- Enables the "document pattern" where a single business object has its own managed properties, validation rules, and authorization AND contains a collection of child items
- Adds `IBusinessDocumentBase<C>` consolidated interface

## Key Features

- Full `IList<C>` collection support via composition (`MobileList<C>`)
- Deleted child tracking (`IContainsDeletedList`)
- N-level undo cascading to collection children
- `IsDirty`/`IsValid`/`IsBusy` aggregation across properties and children
- `MobileFormatter` serialization of collection items
- `CollectionChanged`/`PropertyChanged` events
- `LoadListMode` for bulk loading without events
- Child data access (`Child_Update`/`Child_UpdateAsync`)
- `RegisterProperty` / `RegisterMethod` with correct type resolution

## Test plan

- [x] 29 unit tests covering:
  - Create and fetch operations
  - Collection operations (Add, Insert, Remove, RemoveAt, Clear, Contains, IndexOf, Enumerator, Indexer)
  - Status aggregation (IsDirty, IsValid, IsNew across properties and children)
  - CollectionChanged events
  - Clone / serialization preservation
  - N-level undo (BeginEdit/CancelEdit/ApplyEdit for properties and children)

Closes #1830


hurcane:
I looked for a related discussion or issue for this PR, but didn't find one. What is the intended utility for this kind of class? 

Looking at this initial plan, I wonder how more complex documents would be handled. By more complex, I mean having multiple child lists. In the invoice example, the editable Invoice class in my project would also have a comments collection, a contacts collection. Thinking across the full project, I don't have any documents with only a single collection.

StefanOssendorf:
You are describing a business (base) object. I think the purpose of this PR is to get the whole rules stuff etc. _into_ a business list class.
For example you have a PositionList which has rules to enforce (e.g. only one item with Type X is allowed) and so on. This kind of rule is currently not possible because the list itself can't define rules.

hurcane:
I get that use case. Those business rules currently have to be handled in the parent class. We typically use properties on the child classes to indicate their error state. For example, the child will have an IsDuplicate property with a rule that ensures the property is false.

I'm all for simplifying this scenario, but it wasn't clear in the description and the generated plan. Sorry for butting in without understanding the full context.

rockfordlhotka:
> I get that use case. Those business rules currently have to be handled in the parent class. We typically use properties on the child classes to indicate their error state. For example, the child will have an IsDuplicate property with a rule that ensures the property is false.
> 
> I'm all for simplifying this scenario, but it wasn't clear in the description and the generated plan. Sorry for butting in without understanding the full context.

#1830

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+using Csla.Properties;
+using Csla.Reflection;
+using Csla.Serialization.Mobile;
+using Csla.Server;
+
+namespace Csla
+{
+  /// <summary>
+  /// Base class for an editable business object that has its own
+  /// properties AND contains a collection of child items.
```
StefanOssendorf: `contains` or `is` a collection of child items?
For me it should be an `is`?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] C> :
+    BusinessBase<T>,
+    IEditableCollection,
+    IContainsDeletedList,
+    IObservableBindingList,
+    INotifyCollectionChanged,
+    IList<C>,
+    IBusinessDocumentBase<C>
+    where T : BusinessDocumentBase<T, C>
+    where C : IEditableBusinessObject
```
StefanOssendorf: should we add the `notnull` constraint here to make clear a
`FooDocuments<FooDocuments, FooDocument?>` isn't allowed?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+        _deletedList ??= new MobileList<C>();
+        return _deletedList;
+      }
+    }
+
+    #endregion
+
+    #region IContainsDeletedList
+
+    IEnumerable<IEditableBusinessObject> IContainsDeletedList.DeletedList
```
StefanOssendorf, line 77: If we want to be memory efficient here we could to:
`(IEnumerable<IEditableBusinessObject>)(_deletedList ?? Enumerable.Empty<IEditableBusinessObject>());`

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+      => (IEnumerable<IEditableBusinessObject>)DeletedList;
+
+    /// <summary>
+    /// Returns true if the internal deleted list
+    /// contains the specified child object.
+    /// </summary>
+    /// <param name="item">Child object to check.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="item"/> is <see langword="null"/>.</exception>
+    [EditorBrowsable(EditorBrowsableState.Advanced)]
+    public bool ContainsDeleted(C item)
```
StefanOssendorf, line 87: Should this be `public` or better be `protected` and the user should implement functionality to check that? 🤔

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+    {
+      // reset the child edit level
+      UndoableBase.ResetChildEditLevel(child, EditLevel, false);
+      // mark the object as deleted
+      child.DeleteChild();
+      // add to deleted collection for storage
+      DeletedList.Add(child);
+    }
+
+    private void UnDeleteChild(C child)
```
StefanOssendorf, line 109: Isn't there missing a `UnDeleteChild` or the like? I just tested this and Deleting+Readding still deletes the item on the server side ^^'

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+      set => SetItem(index, value);
+    }
+
+    /// <summary>
+    /// Adds an item to the collection.
+    /// </summary>
+    /// <param name="item">The child object to add.</param>
+    public void Add(C item)
+    {
+      InsertItem(_items.Count, item);
```
StefanOssendorf, line 217: null guard missing?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+    }
+
+    /// <summary>
+    /// Inserts an item at the specified index.
+    /// </summary>
+    /// <param name="index">Zero-based index.</param>
+    /// <param name="item">The child object to insert.</param>
+    public void Insert(int index, C item)
+    {
```
StefanOssendorf, line 227: null guard missing?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+    }
+
+    /// <summary>
+    /// Removes the first occurrence of a specific item from the collection.
+    /// </summary>
+    /// <param name="item">The child object to remove.</param>
+    /// <returns>True if the item was found and removed.</returns>
+    public bool Remove(C item)
+    {
+      int index = _items.IndexOf(item);
```
StefanOssendorf, line 245: null guard missing?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+    public void Clear()
+    {
+      ClearItems();
+    }
+
+    /// <summary>
+    /// Determines whether the collection contains a specific item.
+    /// </summary>
+    /// <param name="item">The item to locate.</param>
+    public bool Contains(C item) => _items.Contains(item);
```
StefanOssendorf: null guard missing?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+
+    /// <summary>
+    /// Determines the index of a specific item in the collection.
+    /// </summary>
+    /// <param name="item">The item to locate.</param>
+    public int IndexOf(C item) => _items.IndexOf(item);
```
StefanOssendorf: null guard missing?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+
+    /// <summary>
+    /// Copies the elements to an array starting at the specified index.
+    /// </summary>
+    /// <param name="array">Destination array.</param>
+    /// <param name="arrayIndex">Start index in array.</param>
+    public void CopyTo(C[] array, int arrayIndex) => _items.CopyTo(array, arrayIndex);
```
StefanOssendorf: null guard missing?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+
+    #region Insert, Remove, Set, Clear
+
+    /// <summary>
+    /// Sets the edit level of the child object as it is added.
+    /// </summary>
+    /// <param name="index">Index of the item to insert.</param>
+    /// <param name="item">Item to insert.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="item"/> is <see langword="null"/>.</exception>
+    protected virtual void InsertItem(int index, C item)
```
StefanOssendorf, line 327: null guard missing?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+
+    /// <summary>
+    /// Replaces the item at the specified index with
+    /// the specified item, first moving the original
+    /// item to the deleted list.
+    /// </summary>
+    /// <param name="index">The zero-based index of the item to replace.</param>
+    /// <param name="item">The new value for the item at the specified index.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="item"/> is <see langword="null"/>.</exception>
+    protected virtual void SetItem(int index, C item)
```
StefanOssendorf, line 400: null guard missing?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+    }
+
+    object IEditableCollection.GetDeletedList()
+    {
+      return DeletedList;
+    }
+
+    void IEditableCollection.SetParent(IParent? parent)
+    {
+      SetParent(parent!);
```
StefanOssendorf: The `!` is wrong here. We have to handle  the `IParent?`.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+    /// <summary>
+    /// Indicates that the specified property belongs
+    /// to the business object type.
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
+    protected static new PropertyInfo<P> RegisterProperty<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] P>(string propertyName)
+    {
+      if (propertyName is null)
```
StefanOssendorf: `string.IsNullOrWhiteSpace` should here be used. An empty property is also invalid.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1115 @@
...
+
+    /// <summary>
+    /// Indicates that the specified property belongs
+    /// to the business object type.
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="info">PropertyInfo object for the property.</param>
+    /// <returns>The provided IPropertyInfo object.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="info"/> is <see langword="null"/>.</exception>
+    protected static new PropertyInfo<P> RegisterProperty<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] P>(PropertyInfo<P> info)
```
StefanOssendorf: Idea: Should we still support the "old way" registrations or skip them in newer types and only use the modern version/art?

StefanOssendorf (commented):
I skipped the tests for now until the implementation is discussed.

rockfordlhotka:
Addressed the review feedback from @StefanOssendorf and Copilot:

- **Doc comment** (:26): Changed "contains" → "is" a collection of child items
- **`notnull` constraint** (:44): Added `where C : notnull, IEditableBusinessObject` to prevent nullable child types
- **Memory efficiency** (:76): `IContainsDeletedList.DeletedList` now uses `_deletedList ?? Enumerable.Empty<>()` to avoid lazy allocation of the deleted list
- **UnDeleteChild on re-add** (:108): `InsertItem` now checks the deleted list and calls `UnDeleteChild` if the item is being re-added, fixing the bug where delete + re-add still deleted on the server
- **Null guards** (:208-353): Added `ArgumentNullException` checks on `Add`, `Insert`, `Remove`, `Contains`, `IndexOf`, and `CopyTo`
- **`IParent?` null-forgiving** (:526): Removed the `!` operator — `SetParent` already accepts `IParent?`
- **`IsNullOrWhiteSpace`** (:926): All `RegisterProperty(string)` and `RegisterMethod(string)` overloads now use `string.IsNullOrWhiteSpace` with `ArgumentException`

**Not changed:**
- **`ContainsDeleted` visibility** (:86): Kept `public` to match `BusinessListBase` convention
- **Legacy `RegisterProperty` support** (:909): Left as-is — this is a design decision for broader discussion
- **`SetItem` IsChild check** (Copilot :388): Already fixed in a prior commit
- **`RegisterProperty` overloads** (Copilot :1112): Already complete in a prior commit

All 29 existing tests pass.

Source/Csla/BusinessDocumentBase.cs
StefanOssendorf: I don't  see this addressed. Yes the child is _removed_ from  the `DeletedList` but it's still marked as Deleted (Done in Line 103).

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    {
+      get => _raiseListChangedEvents;
+      protected set => _raiseListChangedEvents = value;
+    }
+
+    /// <summary>
+    /// Use this object to suppress list changed events
+    /// during bulk operations. Equivalent to <see cref="LoadListMode"/>.
+    /// </summary>
+    public IDisposable SuppressListChangedEvents => LoadListMode;
```
StefanOssendorf, line 155: Why do we need `SuppressListChangedEvents` and `LoadListMode`? They do the same? Is it to have a similar api to the existing list types?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    }
+
+    /// <summary>
+    /// Adds an item to the collection.
+    /// </summary>
+    /// <param name="item">The child object to add.</param>
+    public void Add(C item)
+    {
+      if (item is null)
+        throw new ArgumentNullException(nameof(item));
```
StefanOssendorf, line 215: Exception not documented in the methods xml-doc.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+
+    /// <summary>
+    /// Inserts an item at the specified index.
+    /// </summary>
+    /// <param name="index">Zero-based index.</param>
+    /// <param name="item">The child object to insert.</param>
+    public void Insert(int index, C item)
+    {
+      if (item is null)
+        throw new ArgumentNullException(nameof(item));
```
StefanOssendorf, line 229: Exception not documented in the methods xml-doc.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+
+    /// <summary>
+    /// Removes the first occurrence of a specific item from the collection.
+    /// </summary>
+    /// <param name="item">The child object to remove.</param>
+    /// <returns>True if the item was found and removed.</returns>
+    public bool Remove(C item)
+    {
+      if (item is null)
+        throw new ArgumentNullException(nameof(item));
```
StefanOssendorf, line 243: Exception not documented in the methods xml-doc.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    }
+
+    /// <summary>
+    /// Determines whether the collection contains a specific item.
+    /// </summary>
+    /// <param name="item">The item to locate.</param>
+    public bool Contains(C item)
+    {
+      if (item is null)
+        throw new ArgumentNullException(nameof(item));
```
StefanOssendorf, line 277: Exception not documented in the methods xml-doc.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    }
+
+    /// <summary>
+    /// Determines the index of a specific item in the collection.
+    /// </summary>
+    /// <param name="item">The item to locate.</param>
+    public int IndexOf(C item)
+    {
+      if (item is null)
+        throw new ArgumentNullException(nameof(item));
```
StefanOssendorf, line 290: Exception not documented in the methods xml-doc.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+
+    /// <summary>
+    /// Copies the elements to an array starting at the specified index.
+    /// </summary>
+    /// <param name="array">Destination array.</param>
+    /// <param name="arrayIndex">Start index in array.</param>
+    public void CopyTo(C[] array, int arrayIndex)
+    {
+      if (array is null)
+        throw new ArgumentNullException(nameof(array));
```
StefanOssendorf, line 304: Exception not documented in the methods xml-doc.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    /// <summary>
+    /// Marks the child object for deletion and moves it to
+    /// the collection of deleted objects.
+    /// </summary>
+    /// <param name="index">Index of the item to remove.</param>
+    protected virtual void RemoveItem(int index)
+    {
+      C child = _items[index];
+      OnRemoveEventHooks((IBusinessObject)child);
+      using (LoadListMode)
```
StefanOssendorf, line 377: Is this necessary?  It's not done in the `Add` case.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+        OnAddEventHooks((IBusinessObject)item);
+        var collectionArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Add, item, index);
+        if (RaiseListChangedEvents)
+          OnCollectionChanged(collectionArgs);
+        OnChildChanged(new Core.ChildChangedEventArgs(item, null, collectionArgs));
+      }
+      else
+      {
+        // item must be marked as a child object
+        throw new InvalidOperationException(Resources.ListItemNotAChildException);
```
StefanOssendorf, line 364: Should this be documented and bubble up the documentation chain?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+      if (item is IUseApplicationContext iuac)
+        iuac.ApplicationContext = ApplicationContext;
+      // set child edit level
+      UndoableBase.ResetChildEditLevel(item, EditLevel, false);
+      // reset EditLevelAdded
+      item.EditLevelAdded = EditLevel;
+      _items[index] = item;
+      OnAddEventHooks((IBusinessObject)item);
+      if (RaiseListChangedEvents)
+        OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Replace, item, (object?)child, index));
```
StefanOssendorf: Why is only in the Add/Remove case `OnChildChanged` invoked but not during a replacement?

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    }
+
+    #endregion
+
+    #region IParent override
+
+    /// <summary>
+    /// This method is called by a child object when it
+    /// wants to be removed from the collection.
+    /// </summary>
```
StefanOssendorf, line 586: This should take the doc from the interface with `/// <inheritdoc />`

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    public override bool IsDirty
+    {
+      get
+      {
+        // Check object's own dirty state (properties + FieldManager children)
+        if (base.IsDirty)
+          return true;
+
+        // Any non-new deletions make us dirty
+        foreach (C item in DeletedList)
```
StefanOssendorf, line 628: This will _always_ allocate the `DeletedList` which we try to avoid until necessary.
So either we should use the implementation which already handles that: `((IContainsDeletedList)this).DeletedList` or use the field with an appropiate `null` check.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+          if (!item.IsNew)
+            return true;
+
+        // Check all collection children
+        foreach (C child in _items)
```
StefanOssendorf, line 633: This complete foreach could be reduced to `return _items.Any(static c => c.IsDirty);`

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    public override bool IsValid
+    {
+      get
+      {
+        // Check object's own validity (rules + FieldManager children)
+        if (!base.IsValid)
+          return false;
+
+        // Check all collection children
+        foreach (C child in _items)
```
StefanOssendorf, line 655: This can be reduced to `return !_items.Any(static c => !c.IsValid);`

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    public override bool IsBusy
+    {
+      get
+      {
+        // Check object's own busy state
+        if (base.IsBusy)
+          return true;
+
+        // Check deleted children
+        foreach (C item in DeletedList)
```
StefanOssendorf, line 676: This will _always_ allocate the `DeletedList` which we try to avoid until necessary.
So either we should use the implementation which already handles that: `((IContainsDeletedList)this).DeletedList` or use the field with an appropiate `null` check.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+          if (item.IsBusy)
+            return true;
+
+        // Check active collection children
+        foreach (C child in _items)
```
StefanOssendorf, line 681: This complete foreach could be reduced to `return _items.Any(static c => c.IsBusy);`

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    #endregion
+
+    #region N-Level Undo
+
+    /// <summary>
+    /// Cascades CopyState to collection children after the
+    /// object's own state has been copied.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Advanced)]
+    protected override void CopyStateComplete()
```
StefanOssendorf, line 698: Shouldn't this somehow override `CopyState` and perform it's _additional_  copy state operations instead of hijacking the "complete" notification?
I'm not knowledgeable enough  about this so may I'm wrong here.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+      foreach (C child in DeletedList)
+        child.CopyState(EditLevel, false);
+    }
+
+    /// <summary>
+    /// Cascades UndoChanges to collection children after the
+    /// object's own state has been restored.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Advanced)]
+    protected override void UndoChangesComplete()
```
StefanOssendorf, line 719: Same as for `CopyStateComplete` but for the undo part.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    }
+
+    /// <summary>
+    /// Cascades AcceptChanges to collection children after the
+    /// object's own changes have been accepted.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Advanced)]
+    protected override void AcceptChangesComplete()
+    {
+      base.AcceptChangesComplete();
```
StefanOssendorf, line 784: Same as for `CopyStateComplete` but for the accept changes part.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+
+      // Cascade to all active children
+      for (int x = 0; x < _items.Count; x++)
+      {
+        C child = _items[x];
+        child.CopyState(EditLevel, false);
+      }
+
+      // Cascade to all deleted children
+      foreach (C child in DeletedList)
```
StefanOssendorf, line 710: This will _always_ allocate the `DeletedList` which we try to avoid until necessary.
So either we should use the implementation which already handles that: `((IContainsDeletedList)this).DeletedList` or use the field with an appropiate `null` check.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+              }
+              finally
+              {
+                _completelyRemoveChild = false;
+              }
+            }
+          }
+
+          // Cancel edit on all deleted items (reverse order)
+          for (int index = DeletedList.Count - 1; index >= 0; index--)
```
StefanOssendorf, line 752: This will _always_ allocate the `DeletedList` which we try to avoid until necessary.
So either we should use the implementation which already handles that: `((IContainsDeletedList)this).DeletedList` or use the field with an appropiate `null` check.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+      foreach (C child in _items)
+      {
+        child.AcceptChanges(EditLevel, false);
+        // If item is below its point of addition, lower point of addition
+        if (child.EditLevelAdded > EditLevel)
+          child.EditLevelAdded = EditLevel;
+      }
+
+      // Cascade to deleted children (reverse order)
+      for (int index = DeletedList.Count - 1; index >= 0; index--)
```
StefanOssendorf, line 796: This will _always_ allocate the `DeletedList` which we try to avoid until necessary.
So either we should use the implementation which already handles that: `((IContainsDeletedList)this).DeletedList` or use the field with an appropiate `null` check.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+        throw new ArgumentNullException(nameof(info));
+      if (formatter is null)
+        throw new ArgumentNullException(nameof(formatter));
+
+      base.OnGetChildren(info, formatter);
+
+      var itemsInfo = formatter.SerializeObject(_items);
+      info.AddChild("_bdb_items", itemsInfo.ReferenceId);
+
+      if (_deletedList != null)
```
StefanOssendorf, line 828: Shouldnt we check here for `if(_deletedList?.Count > 0)`? With that we would avoid the transfer of an empty `DeletedList` which does not provide any value.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+
+      // Re-establish parent references and event hooks for active items
+      foreach (var item in _items)
+      {
+        item.SetParent(this);
+        OnAddEventHooks((IBusinessObject)item);
+      }
+
+      // Re-establish parent references for deleted items
+      foreach (var item in DeletedList)
```
StefanOssendorf, line 877: This will _always_ allocate the `DeletedList` which we try to avoid until necessary.
So either we should use the implementation which already handles that: `((IContainsDeletedList)this).DeletedList` or use the field with an appropiate `null` check.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    [EditorBrowsable(EditorBrowsableState.Advanced)]
+    protected virtual void Child_Update(params object?[] parameters)
+    {
+      if (parameters is null)
+        throw new ArgumentNullException(nameof(parameters));
+
+      using (LoadListMode)
+      {
+        var dp = ApplicationContext.CreateInstanceDI<DataPortal<C>>();
+        foreach (var child in DeletedList)
```
StefanOssendorf, line 905: This will _always_ allocate the `DeletedList` which we try to avoid until necessary.
So either we should use the implementation which already handles that: `((IContainsDeletedList)this).DeletedList` or use the field with an appropiate `null` check.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    [UpdateChild]
+    protected virtual async Task Child_UpdateAsync(params object?[] parameters)
+    {
+      if (parameters is null)
+        throw new ArgumentNullException(nameof(parameters));
+
+      using (LoadListMode)
+      {
+        var dp = ApplicationContext.CreateInstanceDI<DataPortal<C>>();
+        foreach (var child in DeletedList)
```
StefanOssendorf, line 933: This will _always_ allocate the `DeletedList` which we try to avoid until necessary.
So either we should use the implementation which already handles that: `((IContainsDeletedList)this).DeletedList` or use the field with an appropiate `null` check.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    /// Indicates that the specified property belongs
+    /// to the business object type.
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
+    protected static new PropertyInfo<P> RegisterProperty<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] P>(string propertyName)
+    {
+      if (string.IsNullOrWhiteSpace(propertyName))
+        throw new ArgumentException("Property name must not be null or empty.", nameof(propertyName));
```
StefanOssendorf, line 957: Text must come from the resource file. There should already be an entry exist for this.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+      return Core.FieldManager.PropertyInfoManager.RegisterProperty<P>(typeof(T), info);
+    }
+
+    /// <summary>
+    /// Indicates that the specified property belongs
+    /// to the business object type.
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
```
StefanOssendorf, line 953: Doc outdated. Only covers the null.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    /// to the business object type.
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <param name="relationship">Relationship with property value.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
+    protected static new PropertyInfo<P> RegisterProperty<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] P>(string propertyName, RelationshipTypes relationship)
+    {
+      if (string.IsNullOrWhiteSpace(propertyName))
+        throw new ArgumentException("Property name must not be null or empty.", nameof(propertyName));
```
StefanOssendorf, line 1060: Text must come from the resource file. There should already be an entry exist for this.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    }
+
+    /// <summary>
+    /// Indicates that the specified property belongs
+    /// to the business object type.
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <param name="relationship">Relationship with property value.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
```
StefanOssendorf, line 1056: Outdated.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    /// to the business object type.
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <param name="friendlyName">Friendly description for a property to be used in databinding.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
+    protected static new PropertyInfo<P> RegisterProperty<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] P>(string propertyName, string? friendlyName)
+    {
+      if (string.IsNullOrWhiteSpace(propertyName))
+        throw new ArgumentException("Property name must not be null or empty.", nameof(propertyName));
```
StefanOssendorf, line 1076: Text must come from the resource file. There should already be an entry exist for this.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    }
+
+    /// <summary>
+    /// Indicates that the specified property belongs
+    /// to the business object type.
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <param name="friendlyName">Friendly description for a property to be used in databinding.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
```
StefanOssendorf, line 1072: Outdated.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <param name="friendlyName">Friendly description for a property to be used in databinding.</param>
+    /// <param name="defaultValue">Default value for the property.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
+    protected static new PropertyInfo<P> RegisterProperty<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] P>(string propertyName, string? friendlyName, P? defaultValue)
+    {
+      if (string.IsNullOrWhiteSpace(propertyName))
+        throw new ArgumentException("Property name must not be null or empty.", nameof(propertyName));
```
StefanOssendorf, line 1093: Text must come from the resource file. There should already be an entry exist for this.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+
+    /// <summary>
+    /// Indicates that the specified property belongs
+    /// to the business object type.
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <param name="friendlyName">Friendly description for a property to be used in databinding.</param>
+    /// <param name="defaultValue">Default value for the property.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
```
StefanOssendorf, line 1089: Outdated.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <param name="friendlyName">Friendly description for a property to be used in databinding.</param>
+    /// <param name="defaultValue">Default value for the property.</param>
+    /// <param name="relationship">Relationship with property value.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
+    protected static new PropertyInfo<P> RegisterProperty<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] P>(string propertyName, string? friendlyName, P? defaultValue, RelationshipTypes relationship)
+    {
+      if (string.IsNullOrWhiteSpace(propertyName))
+        throw new ArgumentException("Property name must not be null or empty.", nameof(propertyName));
```
StefanOssendorf, line 1111: Text must come from the resource file. There should already be an entry exist for this.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    /// <summary>
+    /// Indicates that the specified property belongs
+    /// to the business object type.
+    /// </summary>
+    /// <typeparam name="P">Type of property.</typeparam>
+    /// <param name="propertyName">Property name from nameof().</param>
+    /// <param name="friendlyName">Friendly description for a property to be used in databinding.</param>
+    /// <param name="defaultValue">Default value for the property.</param>
+    /// <param name="relationship">Relationship with property value.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <see langword="null"/>.</exception>
```
StefanOssendorf, line 1107: Outdated.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+
+    /// <summary>
+    /// Registers a method for use in Authorization.
+    /// </summary>
+    /// <param name="methodName">Method name from nameof().</param>
+    /// <exception cref="ArgumentNullException"><paramref name="methodName"/> is <see langword="null"/>.</exception>
+    protected static new MethodInfo RegisterMethod(string methodName)
+    {
+      if (string.IsNullOrWhiteSpace(methodName))
+        throw new ArgumentException("Method name must not be null or empty.", nameof(methodName));
```
StefanOssendorf, line 1124: Text must come from the resource file. There should already be an entry exist for this.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+        throw new ArgumentException("Property name must not be null or empty.", nameof(propertyName));
+
+      return RegisterProperty(Core.FieldManager.PropertyInfoFactory.Factory.Create<P>(typeof(T), propertyName, friendlyName, defaultValue, relationship));
+    }
+
+    /// <summary>
+    /// Registers a method for use in Authorization.
+    /// </summary>
+    /// <param name="methodName">Method name from nameof().</param>
+    /// <exception cref="ArgumentNullException"><paramref name="methodName"/> is <see langword="null"/>.</exception>
```
StefanOssendorf, line 1120: outdated.

Source/Csla/IBusinessDocumentBase.cs
```diff
@@ -0,0 +1,26 @@
...
+  /// Consolidated interface for the BusinessDocumentBase type,
+  /// which combines BusinessBase and BusinessListBase capabilities.
+  /// A business document has its own properties AND is
+  /// a collection of child items.
+  /// </summary>
+  /// <typeparam name="C">Type of the child objects contained in the collection.</typeparam>
+  public interface IBusinessDocumentBase<C> :
+    IBusinessBase,
+    IBusinessListBase<C>
+    where C : IEditableBusinessObject
```
StefanOssendorf, line 24: `notnull` constraint must be added, too.

Source/Csla/BusinessDocumentBase.cs
```diff
@@ -0,0 +1,1149 @@
...
+    where C : notnull, IEditableBusinessObject
+  {
+    #region Collection Storage
+
+    [NotUndoable]
+    [NonSerialized]
+    private bool _completelyRemoveChild;
+
+    [NotUndoable]
+    private MobileList<C> _items = new();
```
StefanOssendorf, line 54: This should be `readonly` and the deserialization should use `AddRange` to add the items instead of assigning the list to this field.

StefanOssendorf (changes requested):

Source/Csla/Core/BusinessBase.cs
```diff
@@ -994,6 +994,20 @@ internal void DeleteChild()
...
+    /// Called by a parent object to reverse a
+    /// previous call to <see cref="DeleteChild"/>,
+    /// restoring the child to a non-deleted state.
+    /// </summary>
+    internal void UnDeleteChild()
+    {
+      if (!IsChild)
+        throw new NotSupportedException(Resources.NoDeleteRootException);
+
+      IsDeleted = false;
```
StefanOssendorf, line 1011: It should check  if it's even deleted if  not skip this and the metapropertychanged call.

Source/tests/Csla.test/BusinessDocumentBase/BusinessDocumentBaseMetastateTests.cs
```diff
@@ -0,0 +1,197 @@
...
+    private MetastateDocument NewDocument()
+    {
+      var portal = _testDIContext.CreateDataPortal<MetastateDocument>();
+      return portal.Create();
+    }
+
+    #region MakeOld
+
+    [TestMethod]
+    [TestCategory("SkipOnCIServer")]
```
StefanOssendorf, line 45: Why are all tests skipped on the CI server?

rockfordlhotka:
## Restrict new types to modern .NET and fix breaking interface change

**Changes:**

- **`BusinessDocumentBase.cs` / `IBusinessDocumentBase.cs`** — Wrapped entirely in `#if NET8_0_OR_GREATER` so these new types only compile for .NET 8+. They are excluded from netstandard2.0, net462, net472, and net48 builds.
- **`IEditableBusinessObject.UnDeleteChild()`** — Wrapped in `#if NET8_0_OR_GREATER` with a default interface implementation (`=> throw new NotImplementedException()`) so it is no longer a breaking change for existing implementors on modern .NET.
- **`BusinessBase.UnDeleteChild()`** — Both the `internal` method and the explicit interface implementation wrapped in `#if NET8_0_OR_GREATER`.
- **`FakeEntity.cs` (test)** — `UnDeleteChild()` wrapped in `#if NET8_0_OR_GREATER` to match.

Build succeeds across all target frameworks (netstandard2.0, net4x, net8.0, net9.0, net10.0) and all 29 BusinessDocumentBase tests pass.

JasonBock:
Just throwing this out here, why support .NET 8 (and .NET 9) when they're end-of-line in 7 months?

rockfordlhotka:
> Just throwing this out here, why support .NET 8 (and .NET 9) when they're end-of-line in 7 months?

We'll drop support for them in version 11.

Source/tests/Csla.test/BusinessDocumentBase/BusinessDocumentBaseMetastateTests.cs
rockfordlhotka: Root cause: DefaultDataPortalActivator.CreateInstance() (line 43) resolves ApplicationContext from its ServiceProvider via GetRequiredService<ApplicationContext>().
  In the Csla.test assembly, other test classes that run earlier set a Windows-mode ApplicationContext into the AsyncLocal context manager. When our test's data portal
  creates a business object, the activator resolves this stale Windows-mode context onto the object instead of the Xaml-mode one.

  Why csla.netcore.test works: That assembly has fewer tests with less AsyncLocal contamination.

  Why SkipOnCIServer is the right fix for now: This is a pre-existing framework limitation (same reason BasicModernTests in Csla.test skips identical tests). Fixing it
  requires changes to how DefaultDataPortalActivator propagates ApplicationContext, which is a separate issue.

PR #4842, opened 2026-03-13
Replace duplicate dictionary lookups with TryGetValue and cached locals

SimonCropp:
- UndoableBase: cache state[fieldName] in a local variable instead of looking it up three times during undo operations
  - UndoableBase/FieldDataManager: replace ContainsKey + GetRequiredValue with single TryGetValue call for _stateStack deserialization
  - FieldDataManager: replace ContainsKey + GetValue with single TryGetValue for child property lookup, also caching the concatenated key to avoid redundant string allocation
  - Remove unnecessary double cast in _stateStack deserialization


StefanOssendorf (approved):
LGTM

PR #4866, opened 2026-06-03
Use registred CustomSerializer for Criteria-Objects

### LINKED ISSUE #4865: Custom Serializer for Criteria-Objects
Since CSLA 9, it has been possible to register custom serializers for data types that do not inherit from IMobileObject. This allows these data types to be used natively as properties, which is particularly useful when using third-party packages.

However, this does not work when using these data types as criteria objects. In that case, the registered custom serializers are bypassed, and the application encounters the following error:

`
Serialization of type object is not supported.
` 

It would be a significant improvement if custom serializers were also utilized for criteria objects.

Alternatively, these data types must be encapsulated in a separate criteria object based on BusinessBase etc., which introduces significant overhead when there is only a single property required for the criteria.

**Additional context**
An analysis of the CSLA source code regarding criteria object serialization revealed that the issue originates in the `DataPortalProxy` class:

```csharp
    private CriteriaRequest GetBaseCriteriaRequest(object criteria)
    {
      if (criteria is not IMobileObject)
        criteria = new PrimitiveCriteria(criteria);
      ...
``` 

Here, all criteria objects that do not inherit from `IMobileObject` are encapsulated in a `PrimitiveCriteria` object. However, this object is incapable of serializing these data types; instead, it simply writes them as-is.

To address this limitation, I modified the implementation of the `PrimitiveCriteria` class as follows:

```csharp
namespace Csla.DataPortalClient
{
  public class PrimitiveCriteria : ReadOnlyBase<PrimitiveCriteria>
  {
    public static readonly PropertyInfo<object> ValueProperty = RegisterProperty<object>(nameof(Value));

    public object Value
    {
      get { return GetProperty(ValueProperty)!; }
      set { LoadProperty(ValueProperty, value); }
    }
  
    [Create]
    [RunLocal]
    private void Create(object value)
    {
      Value = value;
    }
  }
}
``` 

Corresponding changes were made in `DataPortalProxy` at line 331 and following:

```csharp
    private CriteriaRequest GetBaseCriteriaRequest(object criteria)
    {
      if (criteria is not IMobileObject)
        criteria = ApplicationContext.GetRequiredService<IDataPortal<PrimitiveCriteria>>().Create(criteria);
``` 

As a result, these criteria objects are now serialized in the same manner as standard properties, leveraging all registered custom serializers.

jensbrand:
### Summary
* **Problem:** Criteria objects not inheriting from `IMobileObject` bypass registered custom serializers, causing serialization errors.
* **Solution:** Refactored `PrimitiveCriteria` into a `ReadOnlyBase` and updated `DataPortalProxy` to instantiate it via the data portal.
* **Result:** Criteria objects now serialize like standard properties and correctly utilize custom serializers.

### Related Issue
Fixes #4865 ([Custom Serializer for Criteria-Objects](https://github.com/MarimerLLC/csla/issues/4865))


StefanOssendorf (commented):
Hi. Thank you for your contribution.

However I think this is the wrong approach to solve this issue. The `PrimitiveCriteria` was meant to deal with the .net primitives (int, string, DateTime, ...) and not with custom serializers.

I think the correct way would be something like adding a way to `ISerializationFormatter` to ask if it can check a particular type.
If yes: Skip `PrimitiveCriteria`
If no: Use `PrimitiveCriteria`

But this is something more fundamental where I'd like to get @rockfordlhotka opinion this. Your way works, maybe as a temporary implementation, but it should be implemented in another way for the future.

rockfordlhotka:
Looking at the changes, this might be the simplest overall solution. There's no real downside (I don't think) to having it subclass `ReadOnlyBase` instead of directly implementing `IMobileObject`.

The drawback is that we're now involving DI to create instances of something that used to be cheap and trivial, but it is _probably_ not going to be a noticeable perf issue.

On the other hand, what @StefanOssendorf says could be valid - other serializers (like the super-fast one from @JasonBock) could, in theory, support serialization of types in ways different from MobileFormatter - with support for different types entirely. For example, I don't think Jason's serializer relies on `IMobileObject` at all.

JasonBock:
The only thing my serializer does with `IMobileObject` is to check if an object in the graph doesn't implement `IGeneratorSerializable` (which is what the source generator does for you). This is for the (hopefully rare) case where a type definition couldn't be updated and we have to fall back to `MobileFormatter`.

Also, my SG allows a user to register custom serialization implementations for types if needed.

jensbrand:
Even with @JasonBock's serializer, serializing Criteria objects still doesn't work, even after registering a CustomSerializer for it. I suspect the fallback to `IMobileFormatter` that Jason mentioned is the culprit here, since `PrimitiveCriteria` derives from it.

JasonBock:
@jensbrand probably because I don't support `CriteriaBase` right now - https://github.com/JasonBock/CslaGeneratorSerialization/blob/main/src/CslaGeneratorSerialization.Analysis/Builders/GeneratorSerializationBuilder.cs#L65-L88. Not that I couldn't, I just need to add that support in - I added an issue to address this: https://github.com/JasonBock/CslaGeneratorSerialization/issues/46

JasonBock:
@jensbrand BTW can you submit a bug report to my serializer with your `PrimitiveCriteria` issue? My serializer **should** work with any type that derives from `IMobileObject`, but there may be something subtle that I'm missing.

jensbrand:
> @jensbrand BTW can you submit a bug report to my serializer with your `PrimitiveCriteria` issue? My serializer **should** work with any type that derives from `IMobileObject`, but there may be something subtle that I'm missing.

Done

rockfordlhotka:
I'm having claude add a suggested alternative to the original issue, and closing this PR in favor of that alternate approach.

Issue #4891, opened 2026-09-04
LazySingleton<T> parameterless constructor is unusable


rockfordlhotka:
**Describe the bug**

`LazySingleton<T>`'s parameterless constructor cannot be used successfully in CSLA 10.x. It has two independent defects, and it is public API, so fixing either one is a breaking change.

**1. The default constructor's delegate throws for most `T`**

`Source/Csla/LazySingleton.cs:29-32`:

```csharp
public LazySingleton()
{
  _delegate = () => (T)_applicationContext.CreateInstanceDI(typeof(T));
}
```

`CreateInstanceDI` with no parameters reaches `ActivatorUtilities.CreateInstance`. Its constructor selection treats *every* public constructor as applicable when zero arguments are supplied, so any `T` with more than one public constructor throws:

> Multiple constructors accepting all given argument types have been found in type '...'. There should only be one applicable constructor.

`Dictionary<TKey, TValue>` — the `T` used by the old `RuleContext` — is one such type.

**2. `_applicationContext` is never set unless the instance is created through the activator**

`_applicationContext` is only assigned via `IUseApplicationContext`, which `DefaultDataPortalActivator.CreateInstance` does. So plain `new LazySingleton<T>()` followed by `.Value` throws `NullReferenceException`.

**3. `LazySingleton<T>` itself is ambiguous to the activator**

Because the type has two public constructors, `CreateInstanceDI(typeof(LazySingleton<X>))` — the only way to get `_applicationContext` populated — hits the same ambiguity error described in (1). There is no way to reach a working default-constructor instance.

This is exactly the exception reported in #4856 (CSLA 8.2.9), where `RuleContext` created its collections with `ApplicationContext.CreateInstanceDI(typeof(LazySingleton<Dictionary<IPropertyInfo, object>>))`. That call site was removed in #4528, so the rules engine is no longer affected in 10.x, but the type itself is still broken.

**Why it is uncovered**

`Source/tests/Csla.test/LazySingleton/LazySingletonTest.cs` never reads `.Value` on the default-constructor path, and never creates the type through DI.

**Version and Platform**

CSLA version: 10.x (`main`)

**Possible resolutions (v11)**

- Remove `LazySingleton<T>` — the framework has no callers left; `Lazy<T>` covers the use case.
- Or remove the parameterless constructor, leaving only `LazySingleton(Func<T>)`, which works correctly.
[...]

PR #4528, opened 2025-02-17
Add NRT to Csla-Core library

StefanOssendorf:
Adding nullable reference annotations to the core csla project. 
There are a lot of API changes which I tried to document in the added csla 10 upgrade document.

PR #4892, opened 2026-09-04
#4856 Add regression tests for Required rule on value type properties

### LINKED ISSUE #4856: Error where adding a Required rule on an int property
**Describe the bug**
When adding a Required Rule on an int property the error is 

Multiple constructors accepting all given argument types have been found in type 'Csla.LazySingleton1[System.Collections.Generic.Dictionary2[Csla.Core.IPropertyInfo,System.Object]]'. There should only be one applicable constructor.
**Version and Platform**
CSLA version: 8.2.9
OS: Windows.
Platform: WinForms, Net Framework 4.8

**Code that Fails**

 BusinessRules.AddRule(new Required(XYZIntegerProperty, "Cannot be null"));

Microsoft.Extensions.DependencyInjection.ActivatorUtilities.TryFindMatchingConstructor(Type instanceType, Type[] argumentTypes, ConstructorInfo& matchingConstructor, Nullable1[]& parameterMap)
   en Microsoft.Extensions.DependencyInjection.ActivatorUtilities.FindApplicableConstructor(Type instanceType, Type[] argumentTypes, ConstructorInfo& matchingConstructor, Nullable1[]& matchingParameterMap)
   en Microsoft.Extensions.DependencyInjection.ActivatorUtilities.CreateInstance(IServiceProvider provider, Type instanceType, Object[] parameters)
   en Csla.Server.DefaultDataPortalActivator.CreateInstance(Type requestedType, Object[] parameters)
   en Csla.ApplicationContext.CreateInstanceDI(Type objectType, Object[] parameters)
   en Csla.Rules.RuleContext..ctor(ApplicationContext applicationContext, Action1 completeHandler)
   en Csla.Rules.BusinessRules.RunRules(IEnumerable1 rules, Boolean cascade, RuleContextModes executionContext)
   en Csla.Rules.BusinessRules.CheckRulesForProperty(IPropertyInfo property, Boolean cascade, RuleContextModes executionMode)
   en Csla.Rules.BusinessRules.CheckRules(IPropertyInfo property, RuleContextModes executionContext)
   en Csla.Core.BusinessBase.CheckPropertyRules(IPropertyInfo property)
   en Csla.Core.BusinessBase.SetProperty[P](PropertyInfo`1 propertyInfo, P newValue, NoAccessBehavior noAccess)

**Additional context**
The problem is resolve when the rule is comment out

rockfordlhotka:
## Summary

Investigating #4856 showed the reported exception cannot occur on `main`.

The user's stack trace came from CSLA 8.2.9, where `RuleContext`'s constructor built its
output/dirty collections through DI:

```csharp
// origin/v9.x Source/Csla/Rules/RuleContext.cs:212
_outputPropertyValues = (LazySingleton<Dictionary<IPropertyInfo, object>>)
  ApplicationContext.CreateInstanceDI(typeof(LazySingleton<Dictionary<IPropertyInfo, object>>));
```

`CreateInstanceDI` with zero parameters reaches `ActivatorUtilities.CreateInstance`, which
treats every public constructor as applicable when no arguments are supplied — so both
`LazySingleton<T>()` and `LazySingleton<T>(Func<T>)` matched, producing the "Multiple
constructors accepting all given argument types" error. This ran on every rule execution;
`int` was incidental. #4528 rewrote `RuleContext` to hold plain `Dictionary`/`List`
collections, removing the call site.

Changes here:

- New test fixture `HasRequiredValueTypeRules` with `PropertyInfo<int>` and
  `PropertyInfo<int?>` properties, both carrying a `Required` rule. Every existing
  `CommonRules.Required` usage in the test tree targeted a `PropertyInfo<string>`, so this
  path had no coverage.
- New `RequiredRuleTests` covering the original repro (setting an `int` property that has a
  `Required` rule), that `0` satisfies the rule on a non-nullable `int`, and that `int?`
  breaks the rule when null and satisfies it when set.
- XML docs on `CommonRules.Required` now state that the rule fails when the value is null or
  its string form is null/empty/whitespace, and that a non-nullable value type property can
  therefore never break it — use a nullable type or `MinValue<T>` instead. The reporter's
  `"Cannot be null"` message on a non-nullable `int` would never have appeared; behavior is
  unchanged, only documented.

Filed #4891 for the still-latent `LazySingleton<T>` defects (its parameterless constructor is
unusable in three separate ways). Fixing that is a breaking change, so it belongs in v11.

## Test plan

- [x] `dotnet build Source\csla.test.sln` — 0 errors
- [x] `dotnet test Source\csla.test.sln --no-build --filter TestCategory!=SkipOnCIServer --settings Source/test.runsettings` — all green (667 passed in `Csla.Tests.dll`, including the 4 new tests)

Closes #4856

https://claude.ai/code/session_011VEzABB5KneYkRmUx9hFD2

### DIFF
--- a/Source/Csla/Rules/CommonRules.cs
+++ b/Source/Csla/Rules/CommonRules.cs
@@ -102,6 +102,18 @@ namespace Csla.Rules.CommonRules
   /// <summary>
   /// Business rule for a required string.
   /// </summary>
+  /// <remarks>
+  /// The rule is broken when the property value is <see langword="null"/>, or when the
+  /// string form of the value is <see langword="null"/>, <see cref="string.Empty"/> or
+  /// consists only of white space.
+  /// <para>
+  /// A non-nullable value type property can therefore never break this rule, because it
+  /// always has a value that produces a non-empty string. For example an
+  /// <see langword="int"/> property with a value of <c>0</c> satisfies the rule. Use a
+  /// nullable property type (such as <see langword="int"/>?) to require that a value was
+  /// supplied, or <see cref="MinValue{T}"/> to require a value within a range.
+  /// </para>
+  /// </remarks>
   public class Required : CommonBusinessRule
   {
     /// <summary>
--- a/Source/tests/Csla.test/ValidationRules/HasRequiredValueTypeRules.cs
+++ b/Source/tests/Csla.test/ValidationRules/HasRequiredValueTypeRules.cs
@@ -1,0 +1,44 @@
+//-----------------------------------------------------------------------
+// <copyright file="HasRequiredValueTypeRules.cs" company="Marimer LLC">
+//     Copyright (c) Marimer LLC. All rights reserved.
+//     Website: https://cslanet.com
+// </copyright>
+// <summary>Business object with Required rules on value type properties</summary>
+//-----------------------------------------------------------------------
+
+namespace Csla.Test.ValidationRules
+{
+  [Serializable]
+  public class HasRequiredValueTypeRules : BusinessBase<HasRequiredValueTypeRules>
+  {
+    public static PropertyInfo<int> IdProperty = RegisterProperty<int>(c => c.Id);
+    public int Id
+    {
+      get { return GetProperty(IdProperty); }
+      set { SetProperty(IdProperty, value); }
+    }
+
+    public static PropertyInfo<int?> NullableIdProperty = RegisterProperty<int?>(c => c.NullableId);
+    public int? NullableId
+    {
+      get { return GetProperty(NullableIdProperty); }
+      set { SetProperty(NullableIdProperty, value); }
+    }
+
+    protected override void AddBusinessRules()
+    {
+      BusinessRules.AddRule(new Rules.CommonRules.Required(IdProperty));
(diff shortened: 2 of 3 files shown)


rockfordlhotka:
Rebased onto current `main` and migrated to `CslaTestHost`.

This PR was about to merge cleanly and then fail to build. It only adds new files, so there was no textual conflict and GitHub reported it mergeable — but `RequiredRuleTests` used `TestDIContext`, `TestDIContextFactory.CreateDefaultContext()` and `.CreateDataPortal<T>()`, and those were migrated in #4894 and deleted in #4896 after the green checkmark here was recorded. The checks above were from 05:35 today, five commits behind.

The change is confined to the test fixture's plumbing:

- `TestDIContext` -> `CslaTestHost`, `CreateDefaultContext()` -> `CslaTestHost.Create()`, `CreateDataPortal<T>()` -> `GetDataPortal<T>()`
- added a `[ClassCleanup]` disposing the host, matching the rest of the suite since #4894

No assertion changed, and the `CommonRules` documentation change is untouched.

Verified on the new base: `dotnet build Source\csla.test.sln` clean, the four regression tests pass, and the CI-filtered suite is green across all 12 assemblies (`Csla.Tests` 668 -> 672).

PR #4925, opened 2026-09-14
Call data portal operation methods explicitly via bundled source generators

### LINKED ISSUE #4359: Call data portal operation methods explicitly

rockfordlhotka:
Closes #4359

Data portal operation methods (`[Fetch]`, `[FetchChild]`, …) are private and invoked dynamically, so IDEs gray them out and the trimmer can remove them. This PR makes the CSLA source generators part of the `Csla` package, makes every operation method statically referenced through a generated interface, and adds opt-in strongly typed data portal extension methods.

## Generators bundled in the Csla package
- The DataPortalInterfaces generator and its tests are merged into the AutoImplementProperties generator project, which is now packed into the Csla nupkg (`analyzers/dotnet/cs`) instead of shipping as its own package.
- `[CslaImplementProperties]`, `[CslaImplementPropertiesInterface<T>]` and `[CslaIgnoreProperty]` move into `Csla.dll`; the Attributes project and the `Microsoft.Bcl.HashCode` dependency are removed.
- `buildTransitive/Csla.targets` raises `CSLABUILD001` when a project still references the retired `Csla.Generator.AutoImplementProperties.CSharp` package.

## Always-generated operations interface (server side)
- For every `partial` class declaring operation methods (abstract classes included), the generator emits an internal nested `IDataPortalOperations` interface with one explicitly implemented member per operation method, named with the operation name (e.g. `Fetch__Int32`). Concrete classes also get `IDataPortalOperationMapping` / `IDataPortalOperationNamedMapping` implementations that call through the interface.
- Discovery now uses `ForAttributeWithMetadataName` with equatable models grouped by type, so partial classes spread across files work and the pipeline caches correctly.
- Fixes found along the way:
  - inject parameter order
  - invalid identifiers for arrays and tuples
  - nullable value types
  - `InjectAttribute` subclasses and keyed services
  - `ValueTask`
  - generic criteria
  - sync/async parity with reflection dispatch
  - overloads that share an operation name, with a CSLADP002 info diagnostic
- Csla's abstract base classes that declare operation methods are now `partial`.

## Pre-resolved invokers (runtime)
- New `IDataPortalOperationInvoker<T>` and `IChildDataPortalOperationInvoker<T>` (`[EditorBrowsable(Never)]`), implemented by `DataPortal<T>`. They take a generated operation name and RunLocal flag, so the client skips reflection-based method lookup.
- `ChildDataPortal` and `DataPortalTarget` pass operation names through for CreateChild/FetchChild named dispatch.
- When the invoker receives a single covariant array criteria value such as `string[]`, it is no longer split into separate criteria values.

## `[DataPortalExtensions]` generator (client side)
- Opt-in per class via `[DataPortalExtensions(Prefix = "...")]`. It generates sync and async extension methods on `IDataPortal<T>` (Create/Fetch/Execute/Delete) and `IChildDataPortal<T>` (CreateChild/FetchChild).
- `[Inject]` parameters are omitted and default values are preserved. The extension method is internal when the business type or a parameter type is internal.
- Each method falls back to the existing `params` methods for custom or mock `IDataPortal<T>` implementations.
- Warnings are reported for:
  - names that `IDataPortal<T>` instance methods would hide
  - duplicate signatures
  - private parameter types
  - generic, abstract or non-CSLA types
- `[NoDataPortalExtension]` excludes a method.
- Adapted from Stefan Ossendorf's MIT-licensed [Csla.DataPortalExtensions](https://github.com/StefanOssendorf/Csla.DataPortalExtensions), with attribution headers and a third-party notice in the release notes.

## Analyzers
- **CSLA0024** (Warning): type with data portal operation methods should be partial, plus a code fix that adds `partial` to the type and its containing types.
- **CSLA0025** (Warning): use the generated data portal extension method instead of `IDataPortal<T>` methods with untyped criteria.

## Tests and verification
- Generator snapshot tests for the operations and extensions generators, plus stage-caching tests for both.
- `csla.netcore.test` integration tests:
  - every generated root and child operation, sync and async, asserting that named dispatch was used
  - `[RunLocal]` bypassing the configured proxy
  - fallback for a custom `IDataPortal<T>`
  - sync/async exception parity with reflection
  - generated interface member names matching `DataPortalOperationNameHelper`
- Analyzer tests for CSLA0024 (and its code fix) and CSLA0025.
- `dotnet build Source\csla.test.sln` and the full test run pass.
- Packaging was checked with a scratch consumer project using the local nupkg:
  - generated code works
  - the retired-package guard fires
  - IDE0051 no longer fires on private operation methods
  - a trimmed publish keeps the private operation method and generated interface, with no trim warnings about them

## Not in this PR
- `Samples/` still reference the retired generator package. They pin CSLA 10 packages, so the reference should be removed when they move to CSLA 11.
- Follow-up items from the plan: AutoSerialization packaging, Execute(T command) extensions, named dispatch for Update-family child operations, the existing sync `Delete` and `IChildDataPortal.CreateChildAsync` bugs, and the inject analyzer.

https://claude.ai/code/session_01Dn4aKN7kqcG78x1HBWApk3


Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalExtensions/DataPortalExtensionsBuilder.cs
```diff
@@ -0,0 +1,242 @@
...
+        if (baseName.EndsWith("Async", StringComparison.Ordinal) && baseName.Length > "Async".Length)
+          baseName = baseName.Substring(0, baseName.Length - "Async".Length);
+
+        var isChild = method.Kind is "CreateChild" or "FetchChild";
+        var receiverInterface = isChild ? "IChildDataPortal" : "IDataPortal";
+        var parameterTypes = string.Join(",", criteria.Select(p => p.PatternTypeDisplay + (p.IsNullableValueType ? "?" : string.Empty)));
+
+        foreach (var isAsync in new[] { true, false })
+        {
+          var name = type.Prefix + baseName + (isAsync ? "Async" : string.Empty);
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 77e215825. The generated extension name is now escaped with `@` when it's a keyword (`@class` → `@class` / `classAsync`). The `Prefix` is also checked now: if it isn't a valid identifier, the new CSLADP008 warning is reported and no extensions are generated. Covered by the `KeywordAndCollidingNames` and `InvalidPrefix` generator tests.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalExtensions/DataPortalExtensionsBuilder.cs
```diff
@@ -0,0 +1,242 @@
...
+
+      return result;
+    }
+
+    private static void AppendMethod(IndentedTextWriter writer, ExtensionTypeModel type, ExtensionMethod extension)
+    {
+      var method = extension.Method;
+      var criteria = method.CriteriaParameters.ToList();
+      var isDelete = method.Kind == "Delete";
+      var receiver = criteria.Any(p => p.Name == ReceiverName) ? "__" + ReceiverName : ReceiverName;
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 77e215825. The receiver name now gets `__` prepended until it matches none of the criteria parameter names, so `portal` and `__portal` together produce `____portal`. Covered by the `KeywordAndCollidingNames` snapshot.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalExtensions/DataPortalExtensionsBuilder.cs
```diff
@@ -0,0 +1,242 @@
...
+      var fallbackArguments = criteria.Count == 0 ? string.Empty : criteriaArray;
+      var returnKeyword = isDelete && !extension.IsAsync ? string.Empty : "return ";
+
+      writer.WriteLine("/// <summary>");
+      writer.WriteLine($"/// Invokes the <c>{Escape(method.MethodDisplay)}</c> {method.Kind} operation of <see cref=\"{type.FullyQualifiedName}\"/>.");
+      writer.WriteLine("/// </summary>");
+      writer.WriteLine($"{visibility} static {returnType} {extension.Name}({string.Join(", ", parameters)})");
+      OpenBlock(writer);
+      writer.WriteLine($"if ({receiver} is {invokerType} __invoker)");
+      writer.Indent++;
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 77e215825. The invoker pattern variable is picked the same way as the receiver name, so it can no longer collide with a parameter named `__invoker`. Covered by the `KeywordAndCollidingNames` snapshot.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalOperations/DataPortalOperationsBuilder.cs
```diff
@@ -0,0 +1,339 @@
...
+        else
+        {
+          var i = criteriaIndex++;
+          string argument;
+          if (byName && parameter.AcceptsNull)
+            argument = parameter.IsNullableValueType
+              ? $"({parameter.TypeDisplay})criteria[{i}]"
+              : $"({parameter.TypeDisplay})criteria[{i}]!";
+          else
+            argument = CriteriaVariable(candidateIndex, i);
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 77e215825. In named dispatch, an `in` criterion that accepts null is now assigned to a local first, and that local is passed with `in`. Covered by the `InCriteriaAndNullableInject` generator test (`Fetch(in string? name, in int id)`), whose output compiles cleanly.

Source/Csla/Server/DataPortalTarget.cs
```diff
@@ -200,9 +200,26 @@ await _operationMapping.InvokeOperationAsync(
       await CallMethodTryAsyncDI<T>(isSync, parameters).ConfigureAwait(false);
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 77e215825, in the generated code instead of `DataPortalTarget`. The generated dispatch now wraps the operation call in try/catch and rethrows via `DataPortalOperationHelper.CreateCallMethodException`, producing the same `CallMethodException` as reflection dispatch. A `DataPortalOperationNotSupportedException` thrown by an operation now arrives wrapped, so the fallback `catch` blocks (root and child, both lines) never see it and the operation is not run again. Only the generated "not matched" `throw` reaches those blocks. The XML docs on `IDataPortalOperationMapping` and `IDataPortalOperationNamedMapping` now state this contract for hand-written implementations. New integration tests: `FetchAsync_OperationThrowsNotSupported_IsNotInvokedAgain` (operation runs exactly once) and `FetchAsync_OperationThrows_WrapsExceptionLikeReflectionDispatch` (exception chain matches the reflection path).

Source/Csla.Analyzers/Csla.Analyzers/UseGeneratedDataPortalExtensionAnalyzer.cs
```diff
@@ -0,0 +1,168 @@
...
+      if (!IsCriteriaMethod(method, isRootPortal, isChildPortal))
+      {
+        return;
+      }
+
+      context.CancellationToken.ThrowIfCancellationRequested();
+
+      if (containingType.TypeArguments[0] is not INamedTypeSymbol businessType ||
+        !HasDataPortalExtensionsAttribute(businessType, symbols.DataPortalExtensionsAttribute))
+      {
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 77e215825. CSLA0025 now applies the generator's eligibility rules before reporting. It skips generic, abstract, private and non-CSLA types, and types with an invalid prefix. It requires an operation method of the matching kind that is not `[NoDataPortalExtension]`, has no `ref`/`out` parameters, isn't a single `object[]` parameter, has no private criteria types, and whose extension name isn't hidden on the receiving interface. When the criteria are passed in expanded `params` form, the argument count must also fit that method's criteria. The analyzer tests' sample type used names the generator hides (`Fetch`, `Create`, …), so it was renamed. New tests cover the no-extension cases, count mismatch, and an invalid prefix. The CSLA0025 doc now describes this.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalExtensions/IncrementalDataPortalExtensionsGenerator.cs
```diff
@@ -0,0 +1,125 @@
...
+
+      var compilation = context.SemanticModel.Compilation;
+      var hiddenNames = new SortedSet<string>(StringComparer.Ordinal);
+      foreach (var portalInterface in new[] { "Csla.IDataPortal`1", "Csla.IChildDataPortal`1" })
+      {
+        var symbol = compilation.GetTypeByMetadataName(portalInterface);
+        if (symbol is null)
+          continue;
+        foreach (var member in symbol.GetMembers())
+          hiddenNames.Add(member.Name);
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 77e215825. The model now carries separate `RootHiddenNames` and `ChildHiddenNames`, and each extension is checked only against its receiving interface. Covered by the `ReceiverSpecificHiddenNames` test: a `[Create] CreateChild(int)` produces `IDataPortal<T>` extensions and a `[FetchChild] Fetch(int)` produces `IChildDataPortal<T>` extensions.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalOperations/DataPortalOperationsBuilder.cs
```diff
@@ -0,0 +1,339 @@
...
+
+      var call = $"{OperationsLocal}.{memberName}({string.Join(", ", arguments)})";
+      writer.WriteLine(method.IsAsync ? $"await {call}.ConfigureAwait(false);" : $"{call};");
+      writer.WriteLine("return;");
+      CloseBlock(writer);
+    }
+
+    private static string FormatServiceResolution(OperationParameterModel parameter)
+    {
+      var serviceType = $"typeof({parameter.PatternTypeDisplay})";
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 77e215825. Injected `Nullable<T>` parameters now resolve `typeof(T?)`. Nullable reference types still resolve by the unannotated type. Covered by the `InCriteriaAndNullableInject` snapshot.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalOperations/OperationDiscovery.cs
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    private static OperationParameterModel BuildParameter(IParameterSymbol parameter)
+    {
+      var type = parameter.Type;
+      var isNullableValueType = type is INamedTypeSymbol { OriginalDefinition.SpecialType: SpecialType.System_Nullable_T };
+      var patternType = isNullableValueType ? ((INamedTypeSymbol)type).TypeArguments[0] : type;
+
+      var isInjected = false;
+      var allowNull = false;
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 77e215825. `AllowNull` is now also true when the injected parameter is `Nullable<T>` or an annotated nullable reference type, matching `ParameterAllowsNull`. Covered by the `InCriteriaAndNullableInject` snapshot and the `FetchAsync_NullableInjectedService_IsOptional` integration test (an unregistered `[Inject] IUnregisteredService?` receives null).

Source/Csla/buildTransitive/Csla.targets
```diff
@@ -0,0 +1,15 @@
...
+    The standalone Csla.Generator.AutoImplementProperties.CSharp package
+    is retired; referencing it alongside Csla would run the generators twice.
+  -->
+  <Target Name="CslaCheckRetiredGeneratorPackage" BeforeTargets="BeforeBuild;CoreCompile">
+    <ItemGroup>
+      <_CslaRetiredGeneratorPackage Include="@(PackageReference)" Condition="'%(Identity)' == 'Csla.Generator.AutoImplementProperties.CSharp'" />
+    </ItemGroup>
+    <Error Condition="'@(_CslaRetiredGeneratorPackage)' != ''"
+           Code="CSLABUILD001"
+           Text="The Csla.Generator.AutoImplementProperties.CSharp package is retired. Starting with CSLA 11 its source generators are included in the Csla package. Remove the PackageReference to Csla.Generator.AutoImplementProperties.CSharp (and any matching PackageVersion entry)." />
```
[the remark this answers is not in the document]
rockfordlhotka: No change. `Csla.Generator.AutoImplementProperties.Attributes.CSharp` was never published: the project had `IsPackable=false`, the attributes shipped inside the generator package, and nuget.org returns 404 for that package id. No consumer can have a PackageReference to it, so the guard for `Csla.Generator.AutoImplementProperties.CSharp` covers the upgrade path.

rockfordlhotka:
Responses to the suppressed comments in the Copilot review summary (all addressed in 77e215825):

- **`DataPortalExtensionsBuilder.cs:133` — arbitrary `Prefix`**: Fixed. A prefix that isn't a valid C# identifier now reports the new CSLADP008 warning and no extensions are generated for the type. CSLA0025 also skips such types. Covered by the `InvalidPrefix` test.
- **`DataPortalExtensionsBuilder.cs:135` — union of hidden names**: Fixed. Hidden names are kept per receiver interface; see the inline reply.
- **`DataPortalOperationsBuilder.cs:295` — no `CallMethodException` wrapping**: Fixed. Generated calls are wrapped via the new `DataPortalOperationHelper.CreateCallMethodException`, so the exception shape (and `DataPortalExceptionHandler` inspection) matches reflection dispatch. Service resolution stays outside the wrapper, as in reflection. `FetchAsync_OperationThrows_WrapsExceptionLikeReflectionDispatch` compares the exception chain of both paths.
- **`DataPortalOperationsDiagnostics.cs:66` — CSLADP006 severity**: Fixed. It is now a Warning, consistent with the other skipped-extension diagnostics and with the PR description.
- **`DataPortalT.cs:610` — exact `object[]` single criterion**: Fixed. Any single reference-type array value, including an exact `object[]`, is now wrapped so it stays one criterion. `FetchAsync_ObjectArrayForSingleObjectCriteria_IsNotSplit` passes an `object[]` to an operation taking `object` through named dispatch.
- **`DataPortalTarget.cs:220` — child fallback ambiguity**: Fixed by the same generated-code wrapping as the root path. Operation exceptions reach `DataPortalTarget` as `CallMethodException`, so only the generated "not matched" exception triggers fallback.

`dotnet build Source\csla.test.sln` and the full CI test run (`--filter TestCategory!=SkipOnCIServer --settings Source/test.runsettings`) pass.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalExtensions/IncrementalDataPortalExtensionsGenerator.cs
```diff
@@ -0,0 +1,128 @@
...
+      hintParts.Add(type.Name);
+
+      var compilation = context.SemanticModel.Compilation;
+
+      return new ExtensionTypeModel
+      {
+        MetadataName = OperationDiscovery.GetFullMetadataName(type),
+        Namespace = @namespace,
+        FullyQualifiedName = type.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat),
+        ExtensionClassName = string.Concat(containers.Select(c => c.Name + "_")) + type.Name + "DataPortalExtensions",
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 4c411d403, though not by renaming. The generated class is `static partial`, so `Outer_InnerDataPortalExtensions` from a top-level `Outer_Inner` and from a nested `Outer.Inner` merge into one class instead of being duplicate definitions. Their methods don't clash, because the receivers are `IDataPortal<Outer_Inner>` and `IDataPortal<Outer.Inner>`. The one real conflict was accessibility: a public part and an internal part fail with CS0262. Now only a public business type puts `public` on its part; other parts omit the modifier, so the parts always agree. Each method still gets `public` or `internal` from its business type. The new `ExtensionClassNameCollision` test (public `Outer_Inner` + internal `Outer.Inner`) compiles with no errors. I kept the naming, since an encoded name would change every nested type's class name.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalOperations/DataPortalOperationsBuilder.cs
```diff
@@ -0,0 +1,365 @@
...
+      {
+        var ordered = group.ToList();
+        var winner = ordered[0];
+        foreach (var candidate in ordered.Skip(1))
+        {
+          if (candidate.InjectCount > winner.InjectCount)
+            winner = candidate;
+        }
+        selected.Add(winner);
+        foreach (var loser in ordered.Where(m => !ReferenceEquals(m, winner)))
```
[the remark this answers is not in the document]
rockfordlhotka: Fixed in 4c411d403. `SelectNamedDispatchMethods` now marks a collision as ambiguous when several methods share the highest inject count. Ambiguous groups are left out of both named and type-based dispatch, so the call falls through to reflection and throws the same `AmbiguousMatchException`. They're reported with the new CSLADP009 warning; CSLADP002 stays for collisions with a clear winner. Covered by the `AmbiguousOperationName` generator test.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalOperations/OperationDiscovery.cs
```diff
@@ -0,0 +1,445 @@
...
+        { IsRecord: true, TypeKind: TypeKind.Struct } => "record struct",
+        { IsRecord: true } => "record",
+        { TypeKind: TypeKind.Struct } => "struct",
+        { TypeKind: TypeKind.Interface } => "interface",
+        _ => "class"
+      };
+      var builder = new StringBuilder("partial ").Append(keyword).Append(' ').Append(EscapeIdentifier(type.Name));
+      if (type.TypeParameters.Length > 0)
+        builder.Append('<').Append(string.Join(", ", type.TypeParameters.Select(t => EscapeIdentifier(t.Name)))).Append('>');
+      return builder.ToString();
```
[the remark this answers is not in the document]
rockfordlhotka: No change. C# only requires constraints to match on partial parts that declare them; a part with no `where` clauses takes the constraints from the parts that have them (CS0265 is only for two parts with different constraints). The `AbstractClass` and `DerivedFromBaseWithOperations` tests use `PersonBase<T> where T : PersonBase<T>`, and their helper asserts the output compilation has no errors or warnings — both pass. Omitting the constraints also avoids copying constraints that reference types needing extra qualification.

rockfordlhotka:
Responses to the suppressed comments in the second Copilot review (changes in 4c411d403):

- **`TypeWithOperationsShouldBePartialAnalyzer.cs:115` and `IncrementalDataPortalOperationsGenerator.cs:65` — derived operation attributes:** No change for now. The generator and CSLA0024 both deliberately recognize only the built-in operation attributes, so they are consistent with each other. A method marked with a custom subclass of `FetchAttribute` still works through reflection dispatch, as before; it just doesn't get generated dispatch. Finding derived attributes would mean semantic analysis of every attributed method in the compilation instead of using `ForAttributeWithMetadataName`, at a real IDE performance cost, for a pattern that's rarely used. If that scenario needs trim safety, it can be a follow-up issue.
- **`UseGeneratedDataPortalExtensionAnalyzer.cs:236` — argument conversions:** Fixed. When criteria are passed in expanded `params` form, each argument must implicitly convert to the matching criteria parameter (including `params` element types and `null` for nullable targets). New tests: `AnalyzeWhenCriteriaTypesDoNotMatchAnyExtension` (`portal.FetchAsync("x")` against `Fetch(int)` is not reported) and `AnalyzeWhenCriteriaConvertToExtensionParameters`.
- **`UseGeneratedDataPortalExtensionAnalyzer.cs:242` — collision selection:** Fixed. The analyzer skips operation methods that lose to another method with the same criteria types, using the generator's rule (most injected parameters, ties to the first declared). A `[NoDataPortalExtension]` winner therefore suppresses the diagnostic. Covered by `AnalyzeWhenPreferredOverloadHasNoDataPortalExtension`.
- **`DataPortalOperationsBuilder.cs:279` — service resolution outside the wrapper:** Fixed. Injected service resolution (and binding `in` criteria to locals) now happens inside the `try`, so resolution failures are wrapped in `CallMethodException` as in `LateBoundObject.CallMethodTryAsyncDI`. The async-on-sync check stays outside, since it already throws `CallMethodException`, just as the reflection path rethrows it without wrapping again. New integration test: `FetchAsync_MissingRequiredService_WrapsExceptionLikeReflectionDispatch` compares the exception chain with reflection dispatch.
- **`OperationDiscovery.cs:441` — small integral keys:** Fixed. `short`, `ushort`, `byte` and `sbyte` constants are emitted with casts (e.g. `(short)(-1)`, `(byte)2`), so boxed keyed-service keys keep their type. The `KeyedInject` test now includes `short` and `byte` keys.

The full CI test run passes.

ossendorf-at-hoelscher:
I'll review it at home later. Two things:
1. Is there an option to disable the generation of sync-extensions? If not could we add a compiler flag to disable them? At work we must only use the async variations to improve throughput.
2. It's best practice to only add diagnostics from analyzers NOT generators. A generator has to run to emit diagnostics which we don't want as little as possible to run. https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.cookbook.md#issue-diagnostics

rockfordlhotka:
Thanks @ossendorf-at-hoelscher, both addressed in cbe92b03a.

**1. Opting out of sync extensions:** Added. Set the `CslaGenerateSyncDataPortalExtensions` MSBuild property to `false` and only the async extension methods are generated:

```xml
<PropertyGroup>
  <CslaGenerateSyncDataPortalExtensions>false</CslaGenerateSyncDataPortalExtensions>
</PropertyGroup>
```

It defaults to `true`. The property is exposed to the compiler through `CompilerVisibleProperty` in the package's `buildTransitive/Csla.targets`, so no other setup is needed. When sync generation is off:
- Diagnostics about sync names that would be hidden or duplicated (CSLADP005/CSLADP007) are no longer reported.
- CSLA0025 no longer flags calls to the sync `IDataPortal<T>` methods, since there is no extension to suggest.

New tests: `AsyncOnly` (generator) and `AnalyzeWhenSyncExtensionsAreNotGenerated` (CSLA0025). I also tested it end to end with a project that imports `Csla.targets`.

**2. Diagnostics from analyzers instead of generators:** Agreed, and done. The generators no longer report any diagnostics. Two new analyzers in the bundled generator assembly report the same IDs:
- `DataPortalOperationsAnalyzer`: CSLADP002, CSLADP009
- `DataPortalExtensionsAnalyzer`: CSLADP003–CSLADP008

The analyzers call the same code the generators use to decide what to emit (`OperationDiscovery`, `SelectNamedDispatchMethods`, and a new `DataPortalExtensionsBuilder.Analyze`), so they can't drift apart. The diagnostic snapshot tests now run the analyzers and assert that the generators report nothing. The messages and locations match what the generators produced before. Diagnostics now also have real source locations, so `#pragma warning disable` works on them, which it didn't when the generator reported them.

The full CI test run passes.

StefanOssendorf:
Two other things:
1. Is it possible to enable the extension generation for a whole assembly without annotating every business object? Would that make sense? Since my implementation is used on a dedicated type as the extensions host it scanned the assembly and used all types.

I had something like this in mind:
```csharp
[assembly: DataPortalExtensions(Prefix = "Portal")]
```

2. I'm not a friend for the `Async`-Methodsuffix which is the reason why my implementation doesn't add it by default but it's possible through a compiler flag. My question is: Would it be bad to make that possible due to my taste (compiler flag with Async as default suffix but overridable) or should we stick with the async and I have to deal with it? :sweat_smile:

docs/analyzers/CSLA0025-UseGeneratedDataPortalExtensionAnalyzer.md
```diff
@@ -0,0 +1,41 @@
...
+
+`Update`, `UpdateAsync`, `UpdateChild`, `UpdateChildAsync`, and `Execute`/`ExecuteAsync` called with a command object are not reported, because they do not take untyped criteria.
+
+Generated code is not analyzed, because the generated extension methods themselves call the untyped data portal methods.
+
+## Code Fix
+
+No code fix exists for this analyzer. Replace the call with the corresponding generated extension method.
```
StefanOssendorf: Can't we implement this code fix?

docs/analyzers/index.md
```diff
@@ -21,4 +21,29 @@ While we try to make the analyzers as stable as we can, we may run into coding p
...
+## Analyzer Reference
+* [CSLA0001](CSLA0001-IsBusinessObjectSerializableAnalyzer.md) - CSLA business objects must be serializable
+* [CSLA0002](CSLA0002-IsOperationMethodPublicAnalyzer.md) - CSLA operations should not be public
+* [CSLA0003](CSLA0003-CheckConstructorsAnalyzer.md) - CSLA business objects must have a public constructor with no arguments
+* [CSLA0004](CSLA0004-CheckConstructorsAnalyzer.md) - CSLA business objects should not have public constructors with parameters
+* [CSLA0005](CSLA0005-FindSaveAssignmentIssueAnalyzer.md) - Do not ignore the result of `Save()`
+* [CSLA0006](CSLA0006-FindSaveAssignmentIssueAnalyzer.md) - Do not ignore the result of `SaveAsync()`
+* [CSLA0007](CSLA0007-EvaluatePropertiesForSimplicityAnalyzer.md) - Properties that use managed backing fields should only use Get/Set/Read/Load methods
+* [CSLA0008](CSLA0008-EvaluateManagedBackingFieldsAnalayzer.md) - Managed backing fields must be public, static and read-only
+* [CSLA0009](CSLA0009-IsOperationMethodPublicAnalyzer.md) - CSLA operations should not be public
```
StefanOssendorf: That's the same as CSLA0002.

docs/code-generators.md
```diff
@@ -4,7 +4,7 @@
 
 Starting with CSLA 9, there are code generators available as NuGet packages. These packages are designed to work with the CSLA .NET framework to generate code for your business objects as part of the .NET build process.
 
-* [AutoImplementProperties](https://github.com/MarimerLLC/csla/tree/main/Source/Csla.Generators/cs/AutoImplementProperties) - Uses code generation to radically reduce the amount of code you need to write to declare properties in a business domain class
+* [AutoImplementProperties](https://github.com/MarimerLLC/csla/tree/main/Source/Csla.Generators/cs/AutoImplementProperties) - Uses code generation to radically reduce the amount of code you need to write to declare properties in a business domain class. Starting with CSLA 11 this generator is included in the `Csla` package, along with the data portal operations generator (explicit, reflection-free calls to data portal operation methods) and the `[DataPortalExtensions]` generator (strongly typed data portal extension methods).
```
StefanOssendorf: DataPortalExtensionGenerator should get it's own paragraph

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalExtensions/DataPortalExtensionsBuilder.cs
```diff
@@ -0,0 +1,284 @@
...
+    /// <summary>
+    /// Selects the extension methods generated for a business type, and
+    /// explains, as diagnostics, why any are not generated. The generator
+    /// uses the methods and the analyzer reports the diagnostics, so both
+    /// apply the same rules.
+    /// </summary>
+    public static (List<ExtensionMethod> Methods, List<DiagnosticInfo> Diagnostics) Analyze(ExtensionGenerationModel model)
+    {
+      var type = model.Type;
+      var diagnostics = new List<DiagnosticInfo>();
```
StefanOssendorf: They are not used anymore so we should get rid of them.

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalOperations/Models/LocationInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+      {
+        if (tree.FilePath == FilePath)
+          return Location.Create(tree, TextSpan);
+      }
+      return Location.Create(FilePath, TextSpan, LineSpan);
+    }
+
+    public static LocationInfo? From(Location? location)
+    {
+      if (location is null || location.SourceTree is null)
```
StefanOssendorf: `location is null || location.SourceTree is null` can be shortened to `location?.SourceTree is null`

Source/Csla/Server/ChildDataPortal.cs
```diff
@@ -172,10 +208,45 @@ public object Fetch([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.P
...
+    /// <param name="objectType">Type of business object to retrieve.</param>
+    /// <param name="operationName">Generated operation name.</param>
+    /// <param name="parameters">Criteria values, one element per criteria parameter.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="objectType"/> or <paramref name="operationName"/> is <see langword="null"/>.</exception>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public object FetchByOperation([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type objectType, string operationName, object?[]? parameters)
+    {
+      if (objectType is null)
+        throw new ArgumentNullException(nameof(objectType));
+      if (operationName is null)
```
StefanOssendorf, line 234: Why is an empty or whitespace operationName allowed? Shouldn't we guard against that, too?

Source/Csla/Server/ChildDataPortal.cs
```diff
@@ -172,10 +208,45 @@ public object Fetch([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.P
...
+    /// <summary>
+    /// Get an existing business object using a pre-resolved fetch child operation.
+    /// </summary>
+    /// <param name="operationName">Generated operation name.</param>
+    /// <param name="parameters">Criteria values, one element per criteria parameter.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="operationName"/> is <see langword="null"/>.</exception>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public async Task<T> FetchByOperationAsync<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] T>(string operationName, object?[]? parameters)
+    {
+      if (operationName is null)
```
StefanOssendorf, line 254: Why is an empty or whitespace operationName allowed? Shouldn't we guard against that, too?

Source/Csla/Server/ChildDataPortal.cs
```diff
@@ -81,10 +82,45 @@ public object Create([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.
...
+    /// <summary>
+    /// Create a new business object using a pre-resolved create child operation.
+    /// </summary>
+    /// <param name="operationName">Generated operation name.</param>
+    /// <param name="parameters">Criteria values, one element per criteria parameter.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="operationName"/> is <see langword="null"/>.</exception>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public async Task<T> CreateByOperationAsync<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] T>(string operationName, object?[]? parameters)
+    {
+      if (operationName is null)
```
StefanOssendorf, line 122: Why is an empty or whitespace operationName allowed? Shouldn't we guard against that, too?

Source/Csla/Server/ChildDataPortal.cs
```diff
@@ -81,10 +82,45 @@ public object Create([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.
...
+    /// <param name="objectType">Type of business object to create.</param>
+    /// <param name="operationName">Generated operation name.</param>
+    /// <param name="parameters">Criteria values, one element per criteria parameter.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="objectType"/> or <paramref name="operationName"/> is <see langword="null"/>.</exception>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public object CreateByOperation([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type objectType, string operationName, object?[]? parameters)
+    {
+      if (objectType is null)
+        throw new ArgumentNullException(nameof(objectType));
+      if (operationName is null)
```
StefanOssendorf, line 102: Why is an empty or whitespace operationName allowed? Shouldn't we guard against that, too?

Source/Csla/Server/DataPortalOperationHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// <param name="target">Business object declaring the method.</param>
+    /// <param name="isSync">True if the client made a synchronous call.</param>
+    /// <param name="serviceProvider">Current service provider.</param>
+    /// <param name="methodName">Name of the operation method.</param>
+    public static void ThrowIfAsyncMethodOnSyncClient(object target, bool isSync, IServiceProvider serviceProvider, string methodName)
+    {
+      if (!isSync)
+        return;
+
+      if (serviceProvider?.GetService(typeof(ApplicationContext)) is ApplicationContext applicationContext
```
StefanOssendorf: If `serviceProvider` can be `null` annotate the parameter.
Use the `GetService<ApplicationContext>` instead of the non-generic one.

Source/Csla/Server/DataPortalOperationHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+        && applicationContext.ExecutionLocation == ApplicationContext.ExecutionLocations.Server)
+        return;
+
+      throw new CallMethodException(
+        target.GetType().Name + "." + methodName + " " + Resources.MethodCallFailed,
```
StefanOssendorf: Shouldn`t target be guarded against null like line 58?

Source/Csla/Server/DataPortalOperationHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+    public static void ThrowIfAsyncMethodOnSyncClient(object target, bool isSync, IServiceProvider serviceProvider, string methodName)
+    {
+      if (!isSync)
+        return;
+
+      if (serviceProvider?.GetService(typeof(ApplicationContext)) is ApplicationContext applicationContext
+        && applicationContext.ExecutionLocation == ApplicationContext.ExecutionLocations.Server)
+        return;
+
+      throw new CallMethodException(
```
StefanOssendorf: Can't we use method from line 56 here? Looks like the same setup to me.

Source/Csla/Server/DataPortalOperationHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// <summary>
+    /// Resolves a service for an injected operation method parameter.
+    /// </summary>
+    /// <param name="serviceProvider">Current service provider.</param>
+    /// <param name="serviceType">Type of service.</param>
+    /// <param name="allowNull">True if the parameter allows a null value.</param>
+    /// <exception cref="InvalidOperationException">The service is required and not registered.</exception>
+    public static object? GetService(IServiceProvider serviceProvider, Type serviceType, bool allowNull)
+    {
+      if (serviceProvider is null)
```
StefanOssendorf, line 79: `serviceType` must be checked, too. I expect the serviceprovider to throw if it's null.

Source/Csla/Server/DataPortalOperationHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// Resolves a keyed service for an injected operation method parameter.
+    /// </summary>
+    /// <param name="serviceProvider">Current service provider.</param>
+    /// <param name="serviceType">Type of service.</param>
+    /// <param name="serviceKey">Service key.</param>
+    /// <param name="allowNull">True if the parameter allows a null value.</param>
+    /// <exception cref="InvalidOperationException">The service is required and not registered, or the provider does not support keyed services.</exception>
+    public static object? GetKeyedService(IServiceProvider serviceProvider, Type serviceType, object? serviceKey, bool allowNull)
+    {
+      if (serviceProvider is null)
```
StefanOssendorf, line 100: `serviceType` must be not null, too I think.

Source/Csla/Server/DataPortalTarget.cs
```diff
@@ -241,6 +263,11 @@ public Task FetchChildAsync(params object?[]? parameters)
       return InvokeChildOperationAsync<FetchChildAttribute>(parameters);
     }
 
+    public Task FetchChildAsync(string? operationName, object?[]? parameters)
+    {
+      return InvokeChildOperationAsync<FetchChildAttribute>(parameters, operationName);
+    }
+
     public Task ExecuteAsync(object criteria, bool isSync, string? operationName = null)
```
StefanOssendorf, line 271: I didn't know it exists but what about `ExecuteChild`?

Source/Csla/BusinessBindingListBase.cs
```diff
@@ -23,7 +23,7 @@ namespace Csla
   /// <typeparam name="C">Type of the child objects contained in the list.</typeparam>
   [SuppressMessage("Microsoft.Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix")]
   [Serializable]
-  public abstract class BusinessBindingListBase<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] T, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] C> :
+  public abstract partial class BusinessBindingListBase<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] T, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] C> :
```
StefanOssendorf, line 26: Why is this necessary?

StefanOssendorf (changes requested):
I'm not done yet, but a first batch.

docs/analyzers/CSLA0025-UseGeneratedDataPortalExtensionAnalyzer.md
rockfordlhotka: Yes, added `UseGeneratedDataPortalExtensionCodeFix`. The analyzer now puts the names of the matching generated extension methods in the diagnostic's properties. The fix renames the call, keeping the arguments, and adds a `using` for the business type's namespace when the extension isn't in scope. When more than one generated extension accepts the arguments (for example `GetById(int)` and `GetByNumber(long)` for `FetchAsync(1)`), it offers one fix per method. This doc section is updated, and the tests are in `UseGeneratedDataPortalExtensionCodeFixTests`. (b4f0938a0)

docs/analyzers/index.md
rockfordlhotka: Good catch. CSLA0009 is the interface variant of CSLA0002; the entry now reads "CSLA operations should not be public on an interface". (b4f0938a0)

docs/code-generators.md
rockfordlhotka: Done. The data portal operations generator and the data portal extensions generator now each have their own entry. The entry for your Csla.DataPortalExtensions package now notes that CSLA 11 and later include the generator in the `Csla` package. (b4f0938a0)

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalExtensions/DataPortalExtensionsBuilder.cs
rockfordlhotka: Agreed, the generator no longer builds diagnostics it throws away. `DataPortalExtensionsBuilder.Analyze` takes an optional reporter callback: the generator passes none, and only `DataPortalExtensionsAnalyzer` supplies one. `DiagnosticInfo` and the `GetDescriptor` id lookup are removed, and the analyzer creates `Diagnostic`s directly. The type location is now filled in only by the analyzer, so it's no longer part of the generator's extension model and doesn't invalidate the cache when code moves. (Methods still carry a location, because generated members are ordered by declaration position.) (b4f0938a0)

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalOperations/Models/LocationInfo.cs
rockfordlhotka: Done. (b4f0938a0)

Source/Csla/Server/ChildDataPortal.cs
rockfordlhotka: Agreed. Empty or white-space operation names now throw `ArgumentException`; null still throws `ArgumentNullException`. The same check is in all four `ChildDataPortal` `*ByOperation` methods. The root `IDataPortalOperationInvoker<T>` implementations in `DataPortal<T>` had no check at all, so they now validate the name the same way. Tests: `RootInvoker_InvalidOperationName_Throws` and `ChildInvoker_InvalidOperationName_Throws`. (b4f0938a0)

Source/Csla/Server/ChildDataPortal.cs
rockfordlhotka: Agreed. Empty or white-space operation names now throw `ArgumentException`; null still throws `ArgumentNullException`. The same check is in all four `ChildDataPortal` `*ByOperation` methods. The root `IDataPortalOperationInvoker<T>` implementations in `DataPortal<T>` had no check at all, so they now validate the name the same way. Tests: `RootInvoker_InvalidOperationName_Throws` and `ChildInvoker_InvalidOperationName_Throws`. (b4f0938a0)

Source/Csla/Server/ChildDataPortal.cs
rockfordlhotka: Agreed. Empty or white-space operation names now throw `ArgumentException`; null still throws `ArgumentNullException`. The same check is in all four `ChildDataPortal` `*ByOperation` methods. The root `IDataPortalOperationInvoker<T>` implementations in `DataPortal<T>` had no check at all, so they now validate the name the same way. Tests: `RootInvoker_InvalidOperationName_Throws` and `ChildInvoker_InvalidOperationName_Throws`. (b4f0938a0)

Source/Csla/Server/ChildDataPortal.cs
rockfordlhotka: Agreed. Empty or white-space operation names now throw `ArgumentException`; null still throws `ArgumentNullException`. The same check is in all four `ChildDataPortal` `*ByOperation` methods. The root `IDataPortalOperationInvoker<T>` implementations in `DataPortal<T>` had no check at all, so they now validate the name the same way. Tests: `RootInvoker_InvalidOperationName_Throws` and `ChildInvoker_InvalidOperationName_Throws`. (b4f0938a0)

Source/Csla/Server/DataPortalOperationHelper.cs
rockfordlhotka: It's never null. Generated code always passes the provider it receives in `InvokeNamedOperationAsync`/`InvokeOperationAsync`. So I removed the `?.`, added an `ArgumentNullException` guard (the parameter stays non-nullable), and switched to `GetService<ApplicationContext>()`. (b4f0938a0)

Source/Csla/Server/DataPortalOperationHelper.cs
rockfordlhotka: Yes, added an `ArgumentNullException` guard for `target`. (b4f0938a0)

Source/Csla/Server/DataPortalOperationHelper.cs
rockfordlhotka: Yes, it now throws `CreateCallMethodException(target, methodName, new NotSupportedException(...))`, so both paths build the exception the same way. (b4f0938a0)

Source/Csla/Server/DataPortalOperationHelper.cs
rockfordlhotka: Added an `ArgumentNullException` guard for `serviceType` here and in `GetKeyedService`. Covered by `OperationHelper_NullArguments_Throw`. (b4f0938a0)

Source/Csla/Server/DataPortalOperationHelper.cs
rockfordlhotka: Added an `ArgumentNullException` guard for `serviceType` here and in `GetService`. Covered by `OperationHelper_NullArguments_Throw`. (b4f0938a0)

Source/Csla/Server/DataPortalTarget.cs
rockfordlhotka: `ExecuteChild` isn't called with client criteria. It runs when a command object is a child and the parent saves it through `UpdateChild` (the `ICommandObject` branch of `DataPortalTarget.UpdateChildAsync`). So it's in the same group as `InsertChild`, `UpdateChild` and `DeleteSelfChild`: the parameters come from the parent's update call, `IChildDataPortal<T>` has no `ExecuteChild` method to generate an extension for, and there's no pre-computed operation name to pass. It still uses generated dispatch, though, through the type-based `IDataPortalOperationMapping.InvokeOperationAsync`, so no reflection is involved; only the name-based lookup doesn't apply.

Source/Csla/BusinessBindingListBase.cs
rockfordlhotka: `BusinessBindingListBase` declares `[UpdateChild] Child_UpdateAsync`, and the generator also runs when Csla itself is built. With `partial`, the generator emits the nested `IDataPortalOperations` interface, whose explicit implementation references `Child_UpdateAsync` statically. That way the trimmer, AOT and IDE analysis all see the method as used. Because the class is abstract, no mapping implementation is generated; derived classes get that. Without `partial`, CSLA0024 would also report it. The other Csla base classes with operation methods (`BusinessListBase`, `CommandBase`, `ReadOnlyBase`, ...) were made partial for the same reason, as noted under breaking changes in the release notes.

rockfordlhotka:
@StefanOssendorf both are in b4f0938a0 (and I replied inline to the first review batch).

**1. Assembly-level attribute:** Yes, it makes sense, and it's added:

```csharp
[assembly: DataPortalExtensions(Prefix = "Portal")]
```

- It generates extensions for every business class in the assembly that can have them: non-abstract, non-generic, accessible outside its containing type, and implementing `ICslaObject`. Records are skipped too. Classes that don't qualify are skipped silently. The CSLADP003/CSLADP004 "invalid target" warnings only appear for classes that carry the attribute themselves.
- A `Prefix` on a class attribute takes precedence over the assembly prefix. A class attribute without `Prefix` inherits the assembly prefix.
- `[NoDataPortalExtension]` can now also go on a class, to opt it out.
- An invalid assembly prefix is reported once, as CSLADP008 on the assembly attribute.
- The generator no longer needs `ForAttributeWithMetadataName` on the class attribute. The attribute info now comes from the operation-type pipeline it already has, plus a small options model built from the compilation and build properties.
- CSLA0025 follows the assembly attribute, including for business types from referenced assemblies.

**2. Async suffix:** Added as an MSBuild property, `Async` by default:

```xml
<CslaDataPortalExtensionsAsyncSuffix>none</CslaDataPortalExtensionsAsyncSuffix>
```

- MSBuild passes an unset `CompilerVisibleProperty` to the compiler as an empty value, so an empty value can't mean "no suffix". That's why the value `none` means no suffix. Any other value is used as the suffix, for example `Task` → `GetByIdTask`.
- With no suffix, the sync and async methods would have the same names. So sync methods aren't generated then, and CSLADP011 suggests setting `CslaGenerateSyncDataPortalExtensions` to `false`. With both set, you get your original naming: `portal.PortalGetById(42)` returning `Task<T>`.
- A value that isn't valid in an identifier is reported as CSLADP010, and `Async` is used instead.
- CSLA0025 and its new code fix use the configured suffix.

The full CI test run passes. I also checked both settings end to end with a project that imports the package's `Csla.targets`.

rockfordlhotka:
@StefanOssendorf thank you for the quality feedback. I think all your suggestions have been addressed - any more coming?

StefanOssendorf:
> @StefanOssendorf thank you for the quality feedback. I think all your suggestions have been addressed - any more coming?

Yes. I'll continue to review tomorrow. I'll build a local alpha to test in a console/lib project.

StefanOssendorf:
Btw how does the new invocation with a name handle method overloads?
A colleague showed me today that calling a Create Child where two methods only differing in injected dependencies always the one with more injections is chosen. I think with the new one the correct one is used or is it an error/warning?

rockfordlhotka:
> Btw how does the new invocation with a name handle method overloads? A colleague showed me today that calling a Create Child where two methods only differing in injected dependencies always the one with more injections is chosen. I think with the new one the correct one is used or is it an error/warning?

This sounds like a different or preexisting issue that should be filed?

docs/code-generators.md
StefanOssendorf: I meant a paragraph like `[AutoImplementProperties]` or `[AutoSerialization]`.

StefanOssendorf:
I can't pack a pre-release on my local machine. I can build the solution fine but `dotnet pack csla.build.sln -o /home/stefan/GitHub/nuget-prereleases -c Release` ends with an error and no Csla.nupkg.
Can you verify that, please?

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalOperations/DataPortalOperationsBuilder.cs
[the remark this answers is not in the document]
StefanOssendorf: Do CSLADP009 and CSLADP002 still exist?

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalOperations/DataPortalOperationsDiagnostics.cs
```diff
@@ -0,0 +1,115 @@
...
+  /// <summary>
+  /// Diagnostic descriptors reported by the data portal operations and data
+  /// portal extensions analyzers, which apply the same rules as the generators.
+  /// The generators themselves report no diagnostics.
+  /// </summary>
+  internal static class DataPortalOperationsDiagnostics
+  {
+    private const string Category = "Csla.DataPortal";
+
+    public const string DuplicateOperationNameId = "CSLADP002";
```
StefanOssendorf, line 22: Aren't all fo these missing under docs/analyzers and including the index.md there?

Source/Csla.Generators/cs/AutoImplementProperties/Csla.Generator.AutoImplementProperties.CSharp/DataPortalOperations/IncrementalDataPortalOperationsGenerator.cs
```diff
@@ -0,0 +1,73 @@
...
+          .Select(static (e, _) => e!)
+          .WithTrackingName(TrackingNames.ExtractOperationMethods)
+          .Collect();
+
+        all = all is null
+          ? entries
+          : all.Value.Combine(entries).Select(static (pair, _) => pair.Left.AddRange(pair.Right));
+      }
+
+      return all!.Value
```
StefanOssendorf, line 68: We should test this with an empty BO implementation. In that case `all` can be null and that should be handled? I don't know. Can't test this yet properly.

Source/Csla/Server/DataPortalOperationHelper.cs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// <summary>
+    /// Resolves a service for an injected operation method parameter.
+    /// </summary>
+    /// <param name="serviceProvider">Current service provider.</param>
+    /// <param name="serviceType">Type of service.</param>
+    /// <param name="allowNull">True if the parameter allows a null value.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="serviceProvider"/> or <paramref name="serviceType"/> is <see langword="null"/>.</exception>
+    /// <exception cref="InvalidOperationException">The service is required and not registered.</exception>
+    public static object? GetService(IServiceProvider serviceProvider, Type serviceType, bool allowNull)
```
StefanOssendorf, line 77: Where exactly is this method used? And wouldn't it be better to either create two methods `object? GetService(SP, Typ)` and `object GetRequiredService(SP, Type)` or just use use `serviceProvider.Get[Required]Service(type)`?

Source/Csla/Server/DataPortalOperationHelper.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// <summary>
+    /// Resolves a keyed service for an injected operation method parameter.
+    /// </summary>
+    /// <param name="serviceProvider">Current service provider.</param>
+    /// <param name="serviceType">Type of service.</param>
+    /// <param name="serviceKey">Service key.</param>
+    /// <param name="allowNull">True if the parameter allows a null value.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="serviceProvider"/> or <paramref name="serviceType"/> is <see langword="null"/>.</exception>
+    /// <exception cref="InvalidOperationException">The service is required and not registered, or the provider does not support keyed services.</exception>
+    public static object? GetKeyedService(IServiceProvider serviceProvider, Type serviceType, object? serviceKey, bool allowNull)
```
StefanOssendorf, line 98: Same question as for GetService

releasenotes.md
```diff
@@ -10,10 +10,58 @@ Primary changes in this release include:
...
+
+* Generated methods take only the criteria parameters (`[Inject]` parameters are omitted), and pass the pre-computed operation name and `[RunLocal]` setting to the data portal, so the client no longer searches for the operation method using reflection.
+* Because `IDataPortal<T>` instance methods such as `FetchAsync` take precedence over extension methods, a generated name that matches one is not generated and is reported as a warning. Use `Prefix` to give the generated methods distinct names.
+* Add `[assembly: DataPortalExtensions]` to generate extension methods for every business class in the assembly that can have them (non-abstract, non-generic classes that implement `ICslaObject` and are accessible outside their containing type). A `Prefix` on a class attribute takes precedence over the assembly `Prefix`.
+* Use `[NoDataPortalExtension]` to exclude a class or an operation method.
+* MSBuild properties control the generated names and forms:
+
+  ```xml
+  <PropertyGroup>
+    <!-- generate only the async methods (default true) -->
```
StefanOssendorf, line 46: This wording is confusing. It's not wrong but with `default true` it looks like _only async method_ generation is the default.
I think `Generate extensions for sync methods (default true)` is easier to understand?

releasenotes.md
```diff
@@ -10,10 +10,58 @@ Primary changes in this release include:
...
+* Analyzer **CSLA0025** suggests using the generated extension methods instead of calling `IDataPortal<T>` methods with untyped criteria, and provides a code fix that replaces the call.
+
+The data portal extension generator and the CSLA0025 analyzer are adapted from [Csla.DataPortalExtensions](https://github.com/StefanOssendorf/Csla.DataPortalExtensions) by [Stefan Ossendorf](https://github.com/StefanOssendorf), used under the MIT License. Thank you, Stefan!
 
 ### Breaking Changes
 
 * CSLA 11 no longer targets .NET 8 or .NET 9. Applications must target .NET 10 or later (or continue to use .NET Framework 4.6.2 through 4.8, or a `netstandard2.0`-compatible runtime).
+* The `Csla.Generator.AutoImplementProperties.CSharp` package is retired because its generators are now included in the `Csla` package. Remove any `PackageReference` (and `PackageVersion`) for it; a build error (`CSLABUILD001`) is reported while the reference remains.
+* Csla's abstract base classes that declare data portal operation methods (`BusinessListBase`, `CommandBase`, `ReadOnlyBase`, and others) are now `partial`.
```
StefanOssendorf, line 64: That's not a breaking change isn't it? The compiled class is "complete" and only within the CSLA assembly the `partial` modifier has any effect.

StefanOssendorf (changes requested):
Some thins to this review:
1. I can't pack the solution locally. Maybe it's an dotnet issue on linux or there is something wrong with the changes.
2. I skimmed through the changes. There are a lot of changes and tbh: with the usage of AI I tend to care less about the code structure because the AI is doing that job now.

